PR #1, opened 2018-10-31
fix: prevent NPE in XmlSettingsReader::getFileProperty

maxhbr:
this fixes some external issue with the number 550


bs-ondem:
Looks good. Can be merged.

PR #7, opened 2018-11-12
fix(issueProcessor): do not fail on not-applicable issue

maxhbr:
The issue validador did not handle the non-applicable issues correctly and thus failed at undesired points.


bs-ondem:
Looks good. Can be merged.

PR #8, opened 2018-11-13
fix(travis): do not build and test assembly

maxhbr:


bs-ondem:
Looks good. Can be merged.

PR #9, opened 2018-11-13
feat(travis): add JDK10 and JDK11 tests

maxhbr:
This PR also adds CLI tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,30 +6,75 @@
 # http://www.eclipse.org/legal/epl-v20.html
 #
 # SPDX-License-Identifier: EPL-2.0
-sudo: false
+sudo: true
 language: java
-cache:
-  apt: true
-  directories:
-    - $HOME/.m2
-before_cache:
-  - rm -rf $HOME/.m2/repository/org/eclipse/sw360/antenna/
-jdk:
-  - openjdk8
-  # - openjdk10
-  # - openjdk11
-  # - oraclejdk8
-  # - oraclejdk9
-  # - oraclejdk11
+addons:
+  apt:
+    packages:
+      - git
+jdk: openjdk8
+
+before_install:
+  - sudo apt-get update -qq
+  - git clone --branch sw360-3.3.0-M1 --depth 1 https://github.com/eclipse/sw360 /tmp/sw360
+  - sudo /tmp/sw360/scripts/install-thrift.sh --no-cleanup
+  - pushd /tmp/sw360
+  - mvn install -pl backend/src/src-licenseinfo,libraries/lib-datahandler -am -DskipTests
+  - popd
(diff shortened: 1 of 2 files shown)


maxhbr:
This PR currently adds the explicit dependency to the package
```xml
<dependency>
    <groupId>javax.xml.bind</groupId>
    <artifactId>jaxb-api</artifactId>
    <version>2.2.11</version>
</dependency>
```
is licensed under `CDDL-1.1 OR GPL-2.0-w/-CPE`.

.travis.yml
```diff
@@ -6,30 +6,81 @@
...
-  # - openjdk10
-  # - openjdk11
-  # - oraclejdk8
-  # - oraclejdk9
-  # - oraclejdk11
+addons:
+  apt:
+    packages:
+      - git
+# cache:
```
greimela-si: Should we enable caching again?

.travis.yml
```diff
@@ -6,30 +6,81 @@
...
-      install:
-        - mvn clean verify -P '!build-assembly' -DskipTests=true -V -B
-        - mvn install -P '!build-assembly' -DskipTests=true -B
+    # compile and execute with the same JDK
+    - &mvn-test
+      name: mvn test
+      script:
+        - mvn test -B
+        - cd example-projects/example-project
+        - mvn package
```
greimela-si: Use `mvn clean test` and `mvn clean package` to isolate the different test cases?

.travis.yml
```diff
@@ -6,30 +6,81 @@
...
+    - &differentJDK-test
+      name: mvn test, run with openjdk9
+      env:
+        - EXECUTION_JDK=9
       script:
-        - mvn test -P '!build-assembly' -B
+        - export JAVA_HOME=$HOME/custom_jdk
+        - /tmp/install-jdk.sh -f $EXECUTION_JDK -l GPL --target $JAVA_HOME
+        - cd example-projects/example-project
+        - mvn package -Dmaven.compiler.source=$EXECUTION_JDK -Dmaven.compiler.target=$EXECUTION_JDK
```
greimela-si: Also use `clean` here.

.travis.yml
maxhbr: I would enable caching in the next PR, I am also unsure whether it helps. Travis says: if you cache something which you would load from the internet, then loading it from the internet might be faster than loading it from the travis cache.

.travis.yml
maxhbr: mvn package is ran in a subdirectory, which has not yet seen any interaction

.travis.yml
maxhbr: This is a fresh checked out subdirectory (`example-projects/example-project`) in which no mvn cmd was called before

maxhbr:
@greimela-si can you please review again?

PR #16, opened 2018-12-04
refactor core datamodel of antenna

maxhbr:
replace the rigid model with an extensible and modular version. Every artifact
now consists of
- a list of facts and
- a list of flags


maxhbr:
ping @greimela-si , @bs-ondem

greimela-si:
We should release a new version after merging this PR, since it changes the core data model.

bs-jokri:
Thus the documentation needs to be adjusted and the release notes need to highlight the need to change the configuraton.

maxhbr:
@bs-jokri The documentation is now updated, but we still need then to highlight the need to change in the release notes

antenna-model/src/main/java/org/eclipse/sw360/antenna/model/artifact/rows/ArtifactCoordinates.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.sw360.antenna.model.artifact.rows;
```
greimela-si: I just noticed that the package is called `row` instead of `fact`.

greimela-si: I just noticed that the package is called `row` instead of `fact`.

PR #28, opened 2018-12-17
fix(sw360): add functionality to set CPE via rest

maxhbr:


maxhbr:
lets wait for traivis...

PR #35, opened 2018-12-21
fix: since cpeId is no longer mandatory, blindvalues are unnecessary

maxhbr:


maxhbr:
@bs-ondem can you please review?

PR #46, opened 2019-02-05
Build policy engine harness

Martin-Idel-SI:
This is the first iteration of help and documentation to develop rules for the policy engine.

The example development environment shows how it is possible to write readable and easy tests.


maxhbr:
I would prefer to have the testing harness as a separate module in the `antenna-testing` package. That would allow us to remove junit from the compile dependencies of the delivered modules.

Martin-Idel-SI:
@maxhbr Done. I also rebased on master.

PR #47, opened 2019-02-06
fix(workflowItemFactory): changed getAsMap()

neubs-bsi:
Issue: 
- When the workflow was written in the pom.xml and undeclared property values were used in the entry values: `<entryValue>${propertyName}</entryValue>`, a NullPointerException was thrown, originating from StepConfiguration.getAsMap() using a Collectors.toMap() function which can not handle null entrys when internally calling merge. 

Solution:
- Was replaced with individually putting elements into Map and then returning this Map as an unmodifiableMap

### DIFF
--- a/antenna-testing/antenna-frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testProjects/ExampleTestProject.java
+++ b/antenna-testing/antenna-frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testProjects/ExampleTestProject.java
@@ -10,6 +10,7 @@
  */
 package org.eclipse.sw360.antenna.frontend.testProjects;
 
+import com.google.common.collect.ImmutableMap;
 import org.eclipse.sw360.antenna.api.IAttachable;
 import org.eclipse.sw360.antenna.api.configuration.AntennaContext;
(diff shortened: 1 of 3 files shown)


antenna-testing/antenna-frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testProjects/ExampleTestProject.java
```diff
@@ -87,15 +88,26 @@ public String getExpectedToolConfigurationProductVersion() {
...
+                .map( s -> {
+                    if(!"Source Validator".equals(s.getName())) {
+                        return s;
+                    }
+                    final Map<String, String> map = Stream.of(s.getConfiguration().getAsMap(), Collections.singletonMap("missingSourcesSeverity", "FAIL"))
+                            .flatMap(m -> m.entrySet().stream())
+                            .collect(Collectors.toMap(
+                                    Map.Entry::getKey,
+                                    Map.Entry::getValue,
+                                    (v1, v2) -> v2));
```
Martin-Idel-SI: Part of this could be extracted into its own method and maybe even put in the Abstract base class or some Utils class (reuse for code below).

However, I don't think it's necessary.

PR #51, opened 2019-02-13
feat(antenna-workflow.analyzers): including of ORT-Result Analyzer

neubs-bsi:
Included an Ort-result-analyzer into workflow-steps/analyzers

It takes a yml file created by ORT (either an analyzer or scan result) and maps the data onto the Antenna Artifact

An additional library was introduced: faster.xml.jackson.dataformat:jackson-dataformat-yaml


antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/test/resources/scan-result.yml
```diff
@@ -0,0 +1,137 @@
+---
```
maxhbr: why is there the `---` ?

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+            CopyrightStatement copyrights = new CopyrightStatement("");
+            Iterator<JsonNode> results = scanResult.get("results").elements();
+            while (results.hasNext()) {
+                Iterator<JsonNode> licenseFindings =
+                        results.next().get("summary").get("license_findings").elements();
+
+                while (licenseFindings.hasNext()) {
+                    JsonNode licenseFind = licenseFindings.next();
+                    licenses.add(licenseFind.get("license").textValue());
+                    Iterator copyrightIterator = licenseFind.get("copyrights").elements();
```
maxhbr: throws NPE if `"copyrights"` is not set?

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private LicensePair mapObservedLicense(JsonNode scanResult) {
+        if (scanResult.get("results").isArray()) {
+            Collection<String> licenses = new ArrayList<>();
+            CopyrightStatement copyrights = new CopyrightStatement("");
+            Iterator<JsonNode> results = scanResult.get("results").elements();
+            while (results.hasNext()) {
+                Iterator<JsonNode> licenseFindings =
+                        results.next().get("summary").get("license_findings").elements();
```
maxhbr: maybe also prone to NPEs?

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+        c.setVersion(version);
+        return Optional.of(c.build());
+    }
+
+    static private Optional<ArtifactCoordinates> mapMavenCoordinates(String namespace, String name, String version) {
+        MavenCoordinates.MavenCoordinatesBuilder c = new MavenCoordinates.MavenCoordinatesBuilder();
+        c.setGroupId(namespace);
+        c.setVersion(version);
+        c.setArtifactId(name);
+        return Optional.of(c.build());
```
maxhbr: Why is optional used here? It is always present, right?

maxhbr: this is found in multiple places

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OrtResultAnalyzer extends ManualAnalyzer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(OrtResultAnalyzer.class);
+
+    static private Optional<ArtifactCoordinates> mapCoordinates(String ortIdentifier) {
```
maxhbr: why not return just `ArtifactCoordinates`

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+
+        List<Artifact> artifacts = new ArrayList<>();
+
+        JsonNode ortPackages = yamlContent.get("analyzer").get("result").get("packages");
+
+        if (ortPackages == null) {
+            return null;
+        }
+
+        boolean hasScannerResults = (yamlContent.get("scanner").elements().hasNext());
```
maxhbr: unnecessary `(` and `)`

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/test/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzerTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OrtResultAnalyzerTest {
+
+    @Test
+    public void testMapLicenses() throws URISyntaxException, IOException {
+        // create OrtResultAnalyzer
+        OrtResultAnalyzer ortResultAnalyzer = new OrtResultAnalyzer();
```
maxhbr: maybe move common code to an `@Before setupUp()` function

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/test/resources/scan-result.yml
neubs-bsi: This is part of the result file given out by ORT, I have left it in to keep it as close to a real result as possible and to be sure that it would not lead to a parsing error with the yaml-Mapper

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
neubs-bsi: if there are no copyrights then there will be an empty array
(like in test/resources/scan-result.yml line 124). Thus when in the next line the copyrightIterator.hasNext() is called it will just not find a next and go on. Since the output of the scan result is either an empty or a filled array, but never missing (otherwise ORT would have failed in some way) I thought it would not be necessary to do a null check here.

antenna-workflow-steps/analyzers/antenna-ort-result-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/OrtResultAnalyzer.java
neubs-bsi: This is the same concept as the previous comment: 
Since the output of the scan result is either an empty or a filled array, but never missing (otherwise ORT would have failed in some way) I thought it would not be necessary to do a null check here.
I can add a test case to be sure though, would that be satisfying? @maxhbr

PR #54, opened 2019-02-19
Antenna P2 cleanup

Martin-Idel-SI:
This PR cleans up P2 resolution in antenna and bundles model and api artifacts.

This is the first step to eventually enable P2 resolution in antenna.


antenna-frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AbstractAntennaCLIFrontend.java
```diff
@@ -53,18 +53,11 @@ protected Path getBuildDirFromFolder(Path folder) {
...
-     * @throws AntennaConfigurationException 
      */
     @Override
-    public AntennaFrontendHelper init() throws AntennaConfigurationException {
+    public AntennaFrontendHelper init() {
         ToolConfiguration toolConfiguration = new AntennaCLISettingsReader(pluginDescendantArtifactIdName)
                 .readSettingsToToolConfiguration(project);
-        
-        // TODO: should be removed after P2ArtifactResolverOSGi works for AbstractAntennaCLIFrontend
-        if (!toolConfiguration.isP2ResolvingDisabled()) {
```
maxhbr: it might be worth to think about how to get rid/remove the usage of the `isP2ResolvingDisabled()` flag. Can we mark it as deprecated?

Martin-Idel-SI: @maxhbr I deleted it basically everywhere. Since it throws an error in the CLI and therefore the Gradle plugin, it can't have been used there. It could be used in the maven plugin, but I found no way to actually deprecate the API and show it to the user.

For now, I just left the component and it doesn't do anything.

maxhbr: Maybe also remove the following line:
https://github.com/bsinno/osm-antenna/blob/dbfa9e902b4a7e6c28cfc5e7a929011273958d33/antenna-frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLISettingsReader.java#L180

antenna-frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLISettingsReader.java
```diff
@@ -177,7 +177,6 @@ private String getDefaultAntennaTargetDirectory(DefaultProject project) {
         // Booleans
         readAntennaBooleanSetting(reader, "attachAll", false, toolConfigBuilder::setAttachAll);
         readAntennaBooleanSetting(reader, "skip", false, toolConfigBuilder::setSkipAntennaExecution);
-        readAntennaBooleanSetting(reader, "disableP2Resolving", true, toolConfigBuilder::setDisableP2Resolving);
```
Martin-Idel-SI: @maxhbr See here for the line mentioned above.

Issue #55, opened 2019-02-20
Add P2 Resolving capabilities to Antenna


Martin-Idel-SI:
As an antenna user, I want to be able to directly resolve artifacts from P2 repositories
If a user specifies P2 repositories missing jars and/or sources should also be resolved and used for an antenna run.

PRs connected to this effort:
- #54 Initial cleanup modifying existing code
- #77 Adding the step
- #107 Adding tests (including an e2e test for travis)
- #118 Some more cleanup.
- #123 Better Jenkins.

Martin-Idel-SI:
Now live and running.

PR #77, opened 2019-02-26
Add p2 step

Martin-Idel-SI:
Connects to #55 

This PR adds P2 resolution capabilities to antenna.
With this PR, antenna can download artifacts and sources from a local P2 repository specified in the `workflow.xml`. This is done using an embedded eclipse equinox framework.

There are still a few things missing for a convenient usage
- Allow antenna to access and download from repositories over the internet
- Improve loading of dependencies: it takes a long time, there are probably too many bundles loaded and it is a bit hard to see the exact bundles.
- Allow to enter several repositories
- Revisit whether the folder where to download the dependencies cannot be set from antenna build instead of having to set it via a config option.

Once those issues are sorted out, the documentation should be linked into the main site.


Martin-Idel-SI:
If you want each PR to add user-facing functionality, splitting is not possible, because there is hardly any functionality here, anyway: We just download a source file and/or a jar from a local repository specified in the config.

This was the big problem with P2 to begin with: As it is a completely different ecosystem from maven, there are two possibilities:
- Implement our own code to parse repositories and guess URLs. This is completely non-extensible: As soon as we want to extract more information from P2 repositories, we need to implement a parser for eclipse P2 content.xml, etc. The repository format is an eclipse internal, so those could break with eclipse versions.
- Use eclipse bundles. This requires setting up OSGi within our application, which is what (rough estimate) ca. 1200 lines are all about. The rest is then using eclipse to deal with p2 repositories. This has the advantage of being arbitrarily extensible: In principle it should be possible to get every bit of information from the repository. 

If this needs splitting anyway, then here are the options:
- Add the `antenna-p2` bundle. This wouldn't add any functionality, it wouldn't be used anywhere (couldn't be used anywhere)
- Then add the workflow step, which depends on the antenna-p2 bundle and actually adds the functionality.
- To make everything a bit smaller, we could limit functionality to only extract a source jar, not a jar (this would save maybe 100 lines, if we are lucky, 200) and add that functionality later.
Other idea:
- Add the OSGi-Embeder and therefore the workflow step with a bare-bones bundle. This adds a workflow-step which does exactly nothing.
- Add and resolve eclipse dependencies. The workflow step still does nothing.
- Extend the P2Resolver in antenna-p2 to actually download jars and extend the test project.

Anyway, I'll work on fixing travis for now.

Martin-Idel-SI:
Tests should be fixed, the failure is definitely a Travis problem (`failed sudo apt update`), a travis rerun should solve the problem.

Issue #22, opened 2018-12-11
Investigate possibilities to use ORT Dependency analyzer as analyzer for SW360.antenna


blaumeiser-at-bosch:
The Open Source Review Toolkit provides some scanners for different buildsystems.
https://github.com/heremaps/oss-review-toolkit

we should check whether we can write a generic analyzer that uses their scanner and puts the data to the SW360.antenna workflow.

Here offered us to help implementing this.

Martin-Idel-SI:
ORT Analyzer will soon be ready in Antenna.

PRs connected to this issue:
#51 Add ort analyzer.

At the moment, it cannot be used in the workflow.xml.

PR #78, opened 2019-02-26
fix(ortResultAnalyzer): NPE removed and added to basic config

neubs-bsi:
Signed-off-by: Stephanie Neubauer [Stephanie.Neubauer@bosch-si.com](mailto:Stephanie.Neubauer@bosch-si.com)

PR related to issue #81 

Slight changes in OrtResultAnalyzer to handle NPE when analyzer is null.

Test were added for the addition, as well as additional once that were missing for completeness.

I also added the ort-result-analyzer to the basic configuration so it can be used in the workflow.xml.


blaumeiser-at-bosch:
@neubs-bsi: Docu is not right, link to #22 is wrong and their is no issue assigned

PR #87, opened 2019-03-07
fix(sw360): Allow to create projects without user metadata

maxhbr:
The functionality to get the metadata of a user fails, if the SW360 is configured to work with screenname instead of email. This removes the functionality and proves, that it was never necessary. The organisation is on creation filled with the organisation of the creator.


blaumeiser-at-bosch:
Fixes #88

Issue #92, opened 2019-03-12
Get rid of double csv dependencies


blaumeiser-at-bosch:
### Summary of the Improvement
Currently antenna-csv-analyzer uses org.apache.commons:commons-csv:1.6 and antenna-csv-license-knowledge-base uses com.opencsv:opencsv:4.4 for the same functionality, reading in a csv file. Clean up the redundancy and optimize for CQ handling!

### Acceptance Criteria
- [x] com.opencsv:opencsv:4.4 is removed as dependency
- [x] Both components use org.apache.commons:commons-csv:1.5, downgrade due to existing CQ for 1.5
- [x] Antenna behavior has not changed

### Definition of Done
- [x] Acceptance criteria fulfilled
- [ ] The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- [ ] All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- [ ] Documentation is updated
- [ ] The result has been reviewed: Everything works as described and can be verified reading only the documentation

blaumeiser-at-bosch:
The dependency com.opencsv:opencsv has been removed. It was indeed the case, that both components used a csv lib in basically the same way, so it was easy to replace by commons-csv. In the act, I updated the validation to be more concrete and I got rid of the index dependency by accessing columns by name not by id. No change of functionality from outside!

PR #93, opened 2019-03-13
internal(csv): Get rid of double csv dependency

blaumeiser-at-bosch:
- Remove dependency ro com.opencsv:opencsv:4.4 and replace it by org.apache.commons:commons-csv
- Downgrade org.apache.commons:commons-csv to 1.5 to match the existing CQ


blaumeiser-at-bosch:
PR for #92

blaumeiser-at-bosch:
Reviewed internally

Issue #94, opened 2019-03-13
Create the necessary CQs for antenna release


blaumeiser-at-bosch:
### Summary of the Improvement
To be able to release antenna, we have to get approval for all dependencies. The approval is done in the form of a CQ at ipzilla of Eclipse Foundation. To get through this, we also have to adapt the dependencies in order to minimize work.

### Acceptance Criteria
- [ ] Dependencies are adapted, so that existing CQs can be piggybacked whereever meaningful
- [ ] All dependencies are handled by a CQ and got at least a license approval

### Definition of Done
- [ ] Acceptance criteria fulfilled
- [ ] The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- [ ] All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- [ ] Documentation is updated
- [ ] The result has been reviewed: Everything works as described and can be verified reading only the documentation

blaumeiser-at-bosch:
CQ Documentation can be found here: https://github.com/eclipse/antenna/wiki/Dependency-:-CQ-Mapping

blaumeiser-at-bosch:
Related issue is #92

PR #95, opened 2019-03-14
Reduce mockito version to make piggyback CQ

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Related issue #94

blaumeiser-at-bosch:
No review needed, it is only ab out the version of Mockito, since the build runs this means, we do not use a feature, not available in the now used version, everything is fine, therefore

PR #96, opened 2019-03-15
Maven Downgrade due to incompatibility issues and to handle release deps

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Simple downgrade of Maven version

blaumeiser-at-bosch:
Related issue #94

PR #107, opened 2019-03-20
Add tests to P2 project

Martin-Idel-SI:
This PR is based on #77 and adds tests:

- add a P2 step to the example project, so that it will be built by each frontend
- add a P2 test project for a tycho build. This will be automatically executed by travis


Martin-Idel-SI:
Also connects to #55

Issue #97, opened 2019-03-15
Get rid of name antenna-stupid-license-knowledge-base


blaumeiser-at-bosch:
### Summary of the Improvement
The name antenna-stupid-license-knowledge-base is not adequate, please change to something better phrased

### Acceptance Criteria
- [ ] The term stupid is not used anymore in the code

### Definition of Done
- [ ] Acceptance criteria fulfilled
- [ ] The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- [ ] All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- [ ] Documentation is updated
- [ ] The result has been reviewed: Everything works as described and can be verified reading only the documentation

PR #110, opened 2019-03-22
cleanup: Get rid of the name stupid in project

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Fixes #97

PR #112, opened 2019-03-25
Version defined as commit string

Martin-Idel-SI:
- Versions are 1.0.0-SNAPSHOT by default as before
- An explicit version can be created by `mvn package -Drevision=<some revision> -Dchangelist=<some changelist>`
- To also change the version for P2, a few files need to be modified. This is indicated in the top level pom.

Note that this versioning scheme breaks the maven release plugin. Releases now have to be done "by hand" modifying the four locations (top-level-pom + three p2 files) + example projects by hand.

### DIFF
--- a/antenna-api/pom.xml
+++ b/antenna-api/pom.xml
@@ -14,7 +14,7 @@
     <parent>
         <groupId>org.eclipse.sw360.antenna</groupId>
         <artifactId>antenna-management</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <version>${revision}${qualifier}</version>
     </parent>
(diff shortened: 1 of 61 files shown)


blaumeiser-at-bosch:
@bs_ondem: Since you are making releases, you should be involved in this PR

bs-ondem:
@Martin-Idel-SI can we squash some commits?

Martin-Idel-SI:
@bs-ondem There you go. Squashed them to one commit.

blaumeiser-at-bosch:
Build ran through except for the standard p2 issue :-(, therefore I merge it

Issue #19, opened 2018-12-11
Use Case: Process project with SW360 support


blaumeiser-at-bosch:
As a project using open source, I want to use Antenna as a continuous integration mechanism to run open source management, so that I only have to get involved, if concrete actions are needed.

Use Case in detail:
Prerequisite:
A project has a list of dependencies used within the project. Manually maintained or detected by an external mechanism. The list is provided e.g. as an antenna config file.

Expected Reaction:
- The Antenna run retrieves the component (release) metadata from SW360.
- Antenna identifies missing information and gives feedback, so that the project can start the clearing process for missing artifacts imeediately
- Antenna runs a policy check over the data provided
- Antenna updates the Project information in SW360
- On demand, Antenna generates the FOSS bundle, even if not complete

blaumeiser-at-bosch:
Use case fulfilled

PR #113, opened 2019-03-28
sw360(bug): when run on windows, windows ascii was used instead of utf-8 in http-headers

neubs-bsi:
Solution done in Pair Programming with @maxhbr.

License text was represented in Windows Ascii, since tina was executed on windows.
Http headers were changed to only accept APPLICATION_JSON_UTF8

Bug found while occupied with: https://github.com/eclipse/antenna/issues/19


blaumeiser-at-bosch:
Merge this, because build issue is due to P2 issue not related to this stuff here

PR #114, opened 2019-03-28
fix(jsonreader): artifact filename is now set when only one file in pathnames

neubs-bsi:
Done:
- included mapFilename() function to set `ArtifactFilename `when pathnames only contains one file
- this then enabled the `GenericCoordinates `to be used and filed, so no null values were given to the `HttpRequest `in the `SW360Updater `anymore

Issue: bug was found when occupied with #19


blaumeiser-at-bosch:
Merge since the build problem has no connection to the change and all other build feedback is fine

PR #129, opened 2019-04-04
Replace zip dependency

Martin-Idel-SI:
The net.lingala.zip4j dependency was previously needed in P2.
With recent changes, all that we need is an "extractAll" functionality.
This was recently added in a standalone file and this PR proposes to just use that dependency.


antenna-workflow-steps/processors/enricher/antenna-p2-resolver/src/main/java/org/eclipse/sw360/antenna/p2resolver/P2RepositoryExtractor.java
```diff
@@ -28,6 +29,15 @@ private P2RepositoryExtractor() {
         // Utility class
     }
 
+    public static void installEclipseProductForP2Resolution(String extractionLocation) throws AntennaException {
+        String location = P2RepositoryExtractor.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        try (JarFile jar = new JarFile(location)) {
+            P2RepositoryExtractor.extractProductFromJar(extractionLocation, location);
+        } catch (IOException e) {
+            P2RepositoryExtractor.extractProductFromFilesystem(extractionLocation, location);
```
maxhbr: This exception handling only works for tests.

Martin-Idel-SI: Added a log comment, I think this should be enough.

PR #135, opened 2019-04-09
More code cleanup

Martin-Idel-SI:
While looking for redundant/superfluous code, my IDE found a number of additional issues. 

The first commit are some minor automatic optimisations.
The second commit simplifies a few usages of streams.
The third commit does some minor formatting (mostly imports).
The fourth commit changes a few things that could in principle be (performance) bugs.

I can squash them after a review.


antenna-core/src/main/java/org/eclipse/sw360/antenna/util/XmlSettingsReader.java
```diff
@@ -49,7 +50,7 @@
      */
     public XmlSettingsReader(String xmlString) throws ParserConfigurationException, SAXException, IOException {
         DocumentBuilder dBuilder = initDocumentBuilder();
-        this.doc = dBuilder.parse(new ByteArrayInputStream(xmlString.getBytes("UTF-8")));
+        this.doc = dBuilder.parse(new ByteArrayInputStream(xmlString.getBytes(StandardCharsets.UTF_8)));
```
blaumeiser-at-bosch: And a third way to express UTF-8 :-)

antenna-testing/antenna-frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AbstractAntennaCLIFrontendTest.java
```diff
@@ -27,9 +27,7 @@
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
```
blaumeiser-at-bosch: Ich mag * prinzipiell nicht, bei Assert kann man vielleicht ein Auge zudrücken, aber generell ist das intransparent und wird bei einer IDE eigentlich auch gar nicht gebraucht.

antenna-workflow-steps/processors/abstract-antenna-compliance-checker/src/test/java/org/eclipse/sw360/antenna/workflow/processors/checkers/AbstractComplianceCheckerTest.java
```diff
@@ -28,8 +28,7 @@
 
 import java.util.*;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
blaumeiser-at-bosch: See above, is this not a test with expected exception?

antenna-workflow-steps/processors/validators/antenna-drools-checker/src/test/java/org/eclipse/sw360/antenna/drools/DroolsEngineTest.java
```diff
@@ -25,12 +24,7 @@
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
blaumeiser-at-bosch: See above, interesting, sometimes you swap this way, sometimes the other way

Martin-Idel-SI: This is standard IntelliJ formatting and we use it everywhere: Once there are a number of imports from the same package (I guess four?), it'll use `*` importing. 

We either need a documented codestyle, or I prefer using the default of IntelliJ that we all seem to have configured (those import changes here are just in classes were no autoformatting has been done).

antenna-core/src/main/java/org/eclipse/sw360/antenna/util/XmlSettingsReader.java
Martin-Idel-SI: Good catch. This was an auto-IDE replacement. 

I guess this is the best way to express it, because there is no magic string involved and thus there is no spelling error possible. I'll change all occurrences to this way here.

antenna-workflow-steps/processors/validators/antenna-drools-checker/src/test/java/org/eclipse/sw360/antenna/drools/DroolsEngineTest.java
blaumeiser-at-bosch: :-(
Fighting the tool does not make sense. As long as we do not reconfigure that manually, it is ok

Issue #60, opened 2019-02-22
Improve documentation


Martin-Idel-SI:
The current documentation leaves a lot of holes and some things are outdated.

This issue is here to collect effort towards having great documentation in the initial release.

PRs addressing documentation:
- #36 Add documentation for some steps which previously didn't have any.

PR #138, opened 2019-04-10
Neubs bsi/doc/31

neubs-bsi:
#60 

1. checks and updates all frontend information
2. adds source zip generator to documentation
3. headers of velocity markdown files are fixed
4. usage of pom properties in workflow.xml are explained
5. sw360 update generator documentation added
6. renderer error in plugin caused by tool-config.md is fixed
7. add more velocity variables for extensions
8. remove rule engine info


Martin-Idel-SI:
Looks good to me.

PR #140, opened 2019-04-10
Allow ordering of workflowsteps

### LINKED ISSUE #139: Allow fixed order of workflow steps
### Summary of the Feature

Some workflow steps need to be executed after others (e.g. the ManifestResolver needs to be executed before the P2Resolver). Currently, the order of the Workflow steps is the order defined in the `workflow.xml` followed by the workflow steps defined in the `pom.xml`. 

Idea: Let the workflow steps define their order of execution

### Acceptance Criteria
- [ ] A workflow step has a property priority which can be freely defined as positive number
- [ ] The workflow steps configured are ordered according the numbers given in the config and executed so that low numbers are executed prior to higher numbers
One of the following two criterias is fulfilled, lets discuss that
- [ ] Alternatively, workflow steps come with a random priority from the number circle of the corresponding workflow step type (see below) and are ordered normally.
- [ ] If a priority is given twice, a warning should be given, the order in which the steps are executed is read sequence, so the one read in first is executed prior to the step read in second
- [ ] The standard workflow steps are numbered with numbers in the range 100, 200, 300, 400 ..., so that projects can configure steps in between other steps.
- [ ] The documentation of this contains a description of number circles and the correlating workflow step type as a recommendation. This way we introduce a natural order, but keep the option open to mix "groups" of steps id needed.

### Definition of Done
- [ ] Acceptance criteria fulfilled
- [ ] A PR is created, the CI infrastructure reports green
- [ ] The PR is reviewed and approved
- [ ] No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- [ ] Test cases are created to prove the functionality of the feature
- [ ] Documentation is updated

Martin-Idel-SI:
Resolves #139 

Very simple implementation:
- Every step has a number (java short) and the corresponding batches (analyzer, processor, generator, outputHandler) are executed in the order defined by these numbers.
- Warnings are written to the command line if steps have the same order number.
- A default number is given to any workflow step (obviously, this has to be overridden to avoid warnings).

Should we document and if yes, where should we document the numbers?

### DIFF
--- a/antenna-api/src/main/java/org/eclipse/sw360/antenna/api/workflow/ConfigurableWorkflowItem.java
+++ b/antenna-api/src/main/java/org/eclipse/sw360/antenna/api/workflow/ConfigurableWorkflowItem.java
@@ -26,6 +26,7 @@ public abstract class ConfigurableWorkflowItem implements IWorkflowable {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigurableWorkflowItem.class);
     protected IProcessingReporter reporter;
     protected AntennaContext context;
+    protected short workflowStepOrder = 1000;
 
     public void setAntennaContext(AntennaContext context) {
         this.context = context;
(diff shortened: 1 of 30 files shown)


Martin-Idel-SI:
Updated with the new antenna-conf-analyzer

bs-ondem:
@Martin-Idel-SI resolve conflicts please.

Martin-Idel-SI:
@bs-ondem Done.

PR #144, opened 2019-04-11
release 1.0.0-RC2

bs-ondem:
This branch was automatically created by the Jenkins job and contains the release and next iteration commits.

### DIFF
--- a/antenna-p2/bundles/antenna-p2-repository-manager/META-INF/MANIFEST.MF
+++ b/antenna-p2/bundles/antenna-p2-repository-manager/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: Antenna P2 Resolver
 Bundle-SymbolicName: antenna-p2-repository-manager;singleton:=true
 Bundle-Activator: org.eclipse.sw360.antenna.p2.Activator
 Bundle-Vendor: Eclipse Foundation
-Bundle-Version: 1.0.0.qualifier
+Bundle-Version: 1.0.0.RC2
 Eclipse-LazyStart: true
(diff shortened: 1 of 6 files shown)


Martin-Idel-SI:
The release has the drawback that the test projects don't build because the antenna-version has not been adapted in all (?) test projects. Otherwise, it seems to build alright.

bs-ondem:
@bs-ondem Do you mean the version in the release commit?

maxhbr:
I think @Martin-Idel-SI means, that you do not adopt https://github.com/eclipse/antenna/blob/552f7d161b7b91086af6083f9b19f626a23bfa95/example-projects/example-project/pom.xml#L25 to `1.0.0-RC2` for the `Prepare release` commit.

maxhbr:
Ok, how do we want to continue with that:
1. declare `release/1.0.0-RC2` as release branch and just do not merge the PR or
2. push the commits here directly to `master`

#### regarding option 1.
This will allow us to add patches to the release. (but the patched version would be `1.0.0-RC3` and should be on another brach?)
If the tag would be merged to master we still could create such a branch on demand.

I am unsure about the advantages of having the tag in the actual history, but as soon as we have to release patch releases without all new features this is already broken.

For this option the `Prepare for next development iteration 1.0.0-SNAPSHOT` commit is actually not good to have. It should be dropped.

#### regarding option 2.
that is the old / default option.

## What we maybe actually want:
maybe a dev branch model would fit our requirements. E.g. having a dev branch and only update the release/stable branch
- once we do a new release or
- for patches.

But this might introduce more complexity.

bs-ondem:
I will push the commits directly to master for this release, but we can discuss how we want to proceed in the future.

maxhbr:
should we close this PR?

bs-ondem:
Yes we can close this PR, because we have now the RC3 release.

PR #155, opened 2019-05-06
simplify achitecture: remove antenna-api package

maxhbr:
since `antenna-model` is always used, this already provided `antenna-api`. There is no need for two separate packages


maxhbr:
this pr just works due to: https://github.com/eclipse/antenna/blob/8966247f85c427c739789a08bdccc7edbf0d2022/pom.xml#L303-L308

maxhbr:
arguments against merging `antenna-model` with `antenna-core`:
- antenna-model uses the `jaxb-maven-plugin` to generate the classes from xsd
- antenna-model is packaged with `maven-bundle-plugin` to be used in the p2 context. antena-core might not be easily packaged

maxhbr:
depends on https://github.com/eclipse/antenna/pull/157

PR #158, opened 2019-05-07
add(gitignore): added gradle and .build file to gitignore

neubs-bsi:
Added all gradle build folders to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 .gitreview
 velocity.log
 antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql
+.gradle
+build


.gitignore
```diff
@@ -8,4 +8,6 @@
 *.iml
 .gitreview
 velocity.log
-antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql
\ No newline at end of file
+antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql
+**/.gradle
```
maxhbr: `**/.gradle` is equivalent to simply writing `.gradle`

neubs-bsi:
@maxhbr updated it to your review.

.gitignore
```diff
@@ -8,4 +8,6 @@
 *.iml
 .gitreview
 velocity.log
-antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql
\ No newline at end of file
+antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql
+.gradle
+**/build
```
maxhbr: sorry, same issue here, leading `**/` is always unnecessary.

neubs-bsi: changed it once more :)

PR #159, opened 2019-05-08
bug(log): move log message about m2 home variable

neubs-bsi:
Warning was deleted, since no problem could be found with and without the variable being set. 
Warning isn't useful. 

Should there be issues discovered later we will address them then directly instead of avoiding them now by throwing a warning.


neubs-bsi:
#164

PR #161, opened 2019-05-13
Do further architecture simplification

maxhbr:
based on https://github.com/eclipse/antenna/pull/155


maxhbr:
I have not changed the module names. The position for core in the file path is `antenna/core` but the module name is still `antenna-core`. I can also remove the `antenna-` prefix there.

blaumeiser-at-bosch:
Ok, please do so. There are still the two components, right, core and model

maxhbr:
right now it is still core and model, yes. But this PR is WIP and I forgot to label it, sorry.

basic-assembly/basic-cli/pom.xml
```diff
@@ -15,11 +15,11 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.eclipse.sw360.antenna</groupId>
-        <artifactId>antenna-basic-assembly</artifactId>
+        <artifactId>basic-assembly</artifactId>
         <version>${revision}${qualifier}</version>
     </parent>
 
-    <artifactId>antenna-cli</artifactId>
+    <artifactId>basic-cli</artifactId>
```
blaumeiser-at-bosch: I would prefer simply cli, basic implies that it is only rudimentary functionality, but it provides all features of Antenna, doesn't it?

basic-assembly/basic-gradle-plugin/build.gradle
```diff
@@ -37,7 +37,7 @@ gradlePlugin {
     plugins {
         OSMRunner {
             id = 'org.eclipse.sw360.antenna'
-            implementationClass = 'org.eclipse.sw360.antenna.gradle.AntennaBasicGradlePlugin'
+            implementationClass = 'org.eclipse.sw360.antenna.gradle.BasicGradlePlugin'
         }
```
blaumeiser-at-bosch: Again, BasicGradlePlugin is from my point not the right viewpoint. It should be GradlePlugin, because from viewpoint Antenna, that is what you need to use. That we customize it for the Bosch Use Case with additional workflow steps is a local phenomenon, not the view of the world.

basic-assembly/basic-cli/pom.xml
maxhbr: I thought the basic assembly is a prototype which can be adopted for internal use. And for that it is a basic version of an actual cli implementation.

basic-assembly/basic-maven-plugin/pom.xml
```diff
@@ -14,11 +14,11 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.eclipse.sw360.antenna</groupId>
-        <artifactId>antenna-basic-assembly</artifactId>
+        <artifactId>basic-assembly</artifactId>
         <version>${revision}${qualifier}</version>
     </parent>
 
-    <artifactId>antenna-maven-plugin</artifactId>
+    <artifactId>basic-maven-plugin</artifactId>
```
blaumeiser-at-bosch: And the same here, from Antenna perspective this is THE Maven Plugin, not a simple not really usable one. No one has to adapt this, if he only wants to use Antenna, right?

basic-assembly/basic-rules/pom.xml
```diff
@@ -15,19 +15,19 @@
...
     <parent>
-        <artifactId>antenna-basic-assembly</artifactId>
         <groupId>org.eclipse.sw360.antenna</groupId>
+        <artifactId>basic-assembly</artifactId>
         <version>${revision}${qualifier}</version>
     </parent>
 
-    <artifactId>antenna-basic-rules</artifactId>
+    <artifactId>basic-rules</artifactId>
     <packaging>jar</packaging>
```
blaumeiser-at-bosch: I prefer common rules not basic rules, the rules are commonly applicable and do not express a legal interpretation, but some structural requirements on a component

basic-assembly/basic-cli/pom.xml
blaumeiser-at-bosch: From my point of view, basic assembly should provide the possbility of using Antenna as is for dependency tracking and management. It should be possible to extend it, but it should not be necessary. Therefore the frontends of Antenna are fully usable frontends, at least they should be. I thought about basic-configuration, but this is "basic" since it is the fundamental configuration to get the system running. It has to be adapted e.g. with the right analysis tool for a certain use case. The question is about basic-assembly, where I think we should really target to have an assembly for Antenna. But this is only a grouping name, so it is not that important.

blaumeiser-at-bosch:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:unpack (unpack) on project antenna-documentation: Unable to find/resolve artifact. Could not find artifact org.eclipse.sw360.antenna:basic-cli:jar:1.0.0-SNAPSHOT in public (https://repo.bosch-si.com/content/groups/public) -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.

blaumeiser-at-bosch:
Build of antenna-documentation creates above problem

PR #166, opened 2019-05-14
This is a consistent prefix of the PR #161, and tries to further simplify the architecture

maxhbr:
- it creates the folder `./antenna/` and groups there the core parts.
- it startes the `./modules/` folder and starts to group modules there


maxhbr:
Sorry, I have also seen that. After a clean the code no longer build and as a fix I have pushe a more recent code state.

blaumeiser-at-bosch:
Unfortunately, still the P2 stuff is broken. What do you propose on how to proceed?

maxhbr:
I will look into this, currently I think it is a problem with the .travis.yml script and that P2 is missing.

maxhbr:
The p2-resolver was actually built on the failing test execution, so that is not the problem:

PR #175, opened 2019-05-21
fix(doc): fix md file for correct rendering

bs-ondem:
Just adds a newline between text and code block in a MD file for correct rendering.


maxhbr:
Where was it rendered incorrectly? GitHub markdown should be able to handle that.

Issue #174, opened 2019-05-21
Maven Dependency Tree Analyzer + Mvn Project update


neubs-bsi:
- Maven Dependency Tree Analyzer Documentation is not up to date
- mvn test project (which tests the analyzer) is not running

Acceptance Criteria:
- [x] make mvn test project run
- [x] update documentation with right configuration

PR #176, opened 2019-05-21
doc(mvn): correct documentation and make mvn test proj run

neubs-bsi:
Issue: #174 

Changes: 
- changed documentation
- updated pom in mvn-test-project to search dependency in system
- renamed MvnDependencyTreeAnalyzer name to reflect name put into Workflow Config for consistency

### DIFF
--- a/modules/maven/antenna-mvn-dependency-tree-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/MvnDependencyTreeAnalyzer.java
+++ b/modules/maven/antenna-mvn-dependency-tree-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/MvnDependencyTreeAnalyzer.java
@@ -80,7 +80,7 @@ public class MvnDependencyTreeAnalyzer extends AbstractAnalyzer {
 
     @Override
     public String getName() {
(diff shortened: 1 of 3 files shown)


modules/maven/antenna-mvn-dependency-tree-analyzer/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/MvnDependencyTreeAnalyzer.java
```diff
@@ -80,7 +80,7 @@ private ArtifactIdentifier getMavenCoordinates(DependencyNode node) {
 
     @Override
     public String getName() {
-        return "MVN DEPENDENCY TREE";
+        return "Maven dependency anaylzer";
```
bs-ondem: Why the renaming?

neubs-bsi: I thought it would be good to reflect the name put in the Workflow Step just for consistency. Am not married to the idea though, so if you think it would be better to leave it as it was, I would change it again

PR #181, opened 2019-05-27
remove unused fileds from toolconifg

maxhbr:
based on https://github.com/eclipse/antenna/pull/161


maxhbr:
There are soon other changes and requirements expected, wich will solve this problem in a nicer way

PR #186, opened 2019-06-05
fix(sw360updater): NPE with missing antenna config xml fixed

neubs-bsi:
The code was depending on the configuration in context to contain a `SW360CoordinateProject`, which was null without antennaconf.xml 

I introduced a null check for the project. 

Part of Issue #19

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360Updater.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360Updater.java
@@ -49,16 +49,10 @@ public class SW360Updater extends AbstractGenerator {
 
     @Override
     public void configure(Map<String, String> configMap) throws AntennaConfigurationException {
-        SW360ProjectCoordinates configuredSW360Project = context.getConfiguration().getConfiguredSW360Project();
-        projectName = configuredSW360Project.getName();
-        projectVersion = configuredSW360Project.getVersion();
+        Optional<SW360ProjectCoordinates> configuredSW360Project = Optional.ofNullable(context.getConfiguration().getConfiguredSW360Project());
 
-        if (configuredSW360Project.getName() == null || configuredSW360Project.getName().isEmpty()) {
-            projectName = context.getProject().getProjectId();
-        }
-        if (configuredSW360Project.getVersion() == null || configuredSW360Project.getVersion().isEmpty()) {
-            projectVersion = context.getProject().getVersion();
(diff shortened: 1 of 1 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SW360Updater.java
```diff
@@ -45,16 +45,24 @@ public SW360Updater() {
     @Override
     public void configure(Map<String, String> configMap) throws AntennaConfigurationException {
         SW360ProjectCoordinates configuredSW360Project = context.getConfiguration().getConfiguredSW360Project();
-        projectName = configuredSW360Project.getName();
```
blaumeiser-at-bosch: I am not a fan of this solution, it is a meaningless spagetti code solution. How about something like
Optional\<Tinaconfig\> configuration  = getTinaconfig();
projectName = retrieveName(configuration);
projectVersion = retrieveVersion(configuration);

String retrieveName(Optional\<Tinaconfig\> configuration) {
    if configuration.isPresent()
        return configuration.getName();
    else
        return context...

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360Updater.java
```diff
@@ -86,4 +81,20 @@ public void configure(Map<String, String> configMap) throws AntennaConfiguration
         }
         return Collections.emptyMap();
     }
+
+    private String retrieveName(Optional<SW360ProjectCoordinates> sw360ProjectCoordinates) {
+        if (sw360ProjectCoordinates.isPresent()) {
```
blaumeiser-at-bosch: Cooler wäre return sw360ProjectCoordinates.map(coord -> coord.getName()).orElse(context.getProject().getProjectId()) :-)

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360Updater.java
```diff
@@ -21,6 +21,7 @@
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360Component;
 import org.eclipse.sw360.antenna.sw360.rest.resource.releases.SW360Release;
 
+import javax.swing.text.html.Option;
```
blaumeiser-at-bosch: Ups wo kommt das denn her?

PR #188, opened 2019-06-06
fix(docu): adopt the missed package names in documentation

maxhbr:


neubs-bsi:
Looks good :) . 
Approved.

PR #190, opened 2019-06-11
model(facts): added facts representing different url information

neubs-bsi:
Added facts necessary for integration of new sw360 data model as part of #19 

Added: 
- ArtifactUrl as abstract class for facts that are urls
- Homepage Url
- source url (change to extend artifactUrl)
- original repository
- software heritage repository

### DIFF
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactHomepage.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactHomepage.java
@@ -1,3 +1,4 @@
+
 /*
  * Copyright (c) Bosch Software Innovations GmbH 2018-2019.
  *
@@ -12,18 +13,18 @@ package org.eclipse.sw360.antenna.model.artifact.facts;
 
 import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
 
-public class ArtifactSourceUrl extends ArtifactUrl {
-    public ArtifactSourceUrl(String sourceUrl) {
-        super(sourceUrl);
+public class ArtifactHomepage extends ArtifactUrl {
+    public ArtifactHomepage(String homepage) {
+        super(homepage);
     }
 
     @Override
     public String getFactContentName() {
-        return "Source URL";
+        return "Homepage";
     }
 
     @Override
     public Class<? extends ArtifactFact> getKey() {
-        return ArtifactSourceUrl.class;
+        return ArtifactHomepage.class;
     }
 }
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactReleaseTagURL.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactReleaseTagURL.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) Bosch Software Innovations GmbH 2018-2019.
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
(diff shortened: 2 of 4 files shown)


core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageRepository.java
```diff
@@ -0,0 +1,30 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.artifact.facts;
+
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
+
+public class ArtifactSoftwareHeritageRepository extends ArtifactUrl {
```
maxhbr: # Question:
What would be stored here? For the example of the latest `apache/thrift` release I see the following options:
- release browse link: https://archive.softwareheritage.org/browse/release/ae93ff0b4bdbd6749f75c23ad23311b512230894/
- the browse link to the revision: https://archive.softwareheritage.org/browse/revision/b2a4d4ae21c789b689dd162deb819665567f481c/
- the browse link to the directory https://archive.softwareheritage.org/browse/directory/86d041247b8ab767a3a223d9a98356d2fbb13a7e/ (note that from the directory there seems to be no way to find out which release or revision might correspond to it)
- or using [persistent identifiers](https://archive.softwareheritage.org/https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html)
  - release: https://archive.softwareheritage.org/swh:1:rel:ae93ff0b4bdbd6749f75c23ad23311b512230894/
  - revision: https://archive.softwareheritage.org/swh:1:rev:b2a4d4ae21c789b689dd162deb819665567f481c/
  - directory: https://archive.softwareheritage.org/swh:1:dir:86d041247b8ab767a3a223d9a98356d2fbb13a7e
- ...
- the link to the repository without tag: https://archive.softwareheritage.org/browse/origin/https://github.com/apache/thrift/directory/

#### More problems:
**Note:** all links usually also contain added GET arguments, which are used by the UI, e.g. the URL https://archive.softwareheritage.org/browse/release/ae93ff0b4bdbd6749f75c23ad23311b512230894/ might actually look like: https://archive.softwareheritage.org/browse/release/ae93ff0b4bdbd6749f75c23ad23311b512230894/?origin=https://github.com/apache/thrift

# Problem:
This is not (easily) parsable or can't be processed automatically.

# Options:
I think this needs to be specified and enforced by the fact implementation. I would even prefer to just save the persistent id part of the release, e.g. `swh:1:rel:ae93ff0b4bdbd6749f75c23ad23311b512230894`, instead of a URL.

maxhbr:
```suggestion
public class ArtifactSoftwareHeritageURL extends ArtifactUrl {
```
I would do that renaming, since the content of that fact is a `URL` and not a `Repository`.

neubs-bsi: The content should be decided by @blaumeiser-at-bosch

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactOriginalRepository.java
```diff
@@ -0,0 +1,29 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.artifact.facts;
+
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
+
+public class ArtifactOriginalRepository extends ArtifactUrl {
```
maxhbr:
```suggestion
public class ArtifactReleaseTagURL extends ArtifactUrl {
```

neubs-bsi:
Follow Up Issue from this PR: https://github.com/bsinno/osm-antenna/issues/754

PR #191, opened 2019-06-11
sw360(facts): added sw360 specific facts to sw360 module

neubs-bsi:
Added and moved sw360 specific facts necessary for integration of new sw360 data model as part of #19 to the sw360 module model package

Added:
- Clearing State
- Change Status

Moved:
- CPE

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/model/artifact/facts/ArtifactClearingState.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/model/artifact/facts/ArtifactClearingState.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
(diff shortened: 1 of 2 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/model/artifact/facts/ArtifactCPE.java
```diff
@@ -9,7 +9,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.eclipse.sw360.antenna.model.artifact.facts;
+package org.eclipse.sw360.antenna.sw360.model.artifact.facts;
```
maxhbr: This is not SW360 specific information, although it currently is only used by SW360. Maybe that could be merged with `Issues`/`SecurityIssues`.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/model/artifact/facts/ArtifactClearingState.java
```diff
@@ -0,0 +1,40 @@
...
+    public Class<? extends ArtifactFact> getKey() {
+        return ArtifactClearingState.class;
+    }
+
+    public enum ClearingState {
+        INITAL,
+        EXTERNAL_SOURCE,
+        AUTO_EXTRACT,
+        PROJECT_APPROVED,
+        OSM_APPROVED,
```
maxhbr:
```suggestion
        OSM_APPROVED
```

PR #195, opened 2019-06-13
fix(test): fix antenna-sw360-integration-testing

maxhbr:
this updates the `antenna-sw360-integration-testing` to the most recent changes. ~~It can be tested by running `mvn test -P it`~~, but needs a dump of the sw360 postgresql at `antenna-testing/antenna-sw360-integration-testing/src/test/resources/postgres/sw360pgdb.sql`

**Note:** currently the tests fail with `401 Unauthorized`.

**Edit:**
- This now also drops the `antenna-system-test`, which was outdated and no longer contained uniq functionality.
- This now also moves the content of the `it` profile tests into the `modules/sw360/` module and it can be executed via  `cd modules/sw360; mvn clean verify -DskipTests -P integration-test` if the corresponding dump is placed at `modules/sw360/src/test/resources/postgres/sw360pgdb.sql`


maxhbr:
This PR can now be reviewed. The integration tests with SW360 still fail with `401 unauthorized` but that might be an old and unrelated problem.

\CC @sschuberth, @neubs-bsi

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/integrationtesting/AntennaSW360IntegrationTest.java
```diff
@@ -89,7 +95,7 @@ public static void afterAll() {
     }
 
     @Before
-    public void setUp() throws IOException {
+    public void setUp() throws Exception {
```
neubs-bsi: I would prefer to catch two Exceptions: IOException and JsonException

PR #196, opened 2019-06-19
sw360(dataModel):  updated and integrated data model of sw360

neubs-bsi:
Correspding Issue: #19

Included:

```
[NAME]: Name der Komponente, also nicht im Release, sondern im Component Record, ich möchte ihn aber in einem Dialog ändern können, auch wenn das dann ein anderes Datenobjekt betrifft.
    GUI Feldname: "Component Name"
[RELEASE]: Die Version des Releases, sollte als nicht editierbares Feld im GUI dargestellt werden
    GUI Feldname: "Release"
[LICENSE]: Ein SPDX-Lizenzausdruck. Dieser Lizenzausdrucks soll so in unter dem licenses Schlüssel in den External Ids gespeichert werden. Für die Speicherung in den "Main Licenses" muss der Ausdruck auseinander genommen und die einzelnen Lizenz-Ids extrahiert werden, bevor sie in das entsprechende Feld in SW360 in Form eines Arrays gespeichert werden können
    GUI Feldname: "Final Licenses"
[DLICENSE] Ein SPDX-Lizenzausdruck
    GUI Feldname: "Declared Licenses"
[OLICENSE] Ein SPDX-Lizenzausdruck
    GUI Feldname: "Observed Licenses"
[SOURCES]: Eine URL
    GUI Feldname: "Internal Source Location"
[OREPO]: Eine URL
    GUI Feldname: "Link to Release Tag in OSS Repo"
[SWHREPO]: Eine URL
    GUI Feldname: "Link to Release ID in SW Heritage"
[COORDINATES]: Liste von Key-Value Paaren, beides Key und Value sind Strings, beliebige Anzahl von Paaren
    GUI Feldname: "Release Coordinates"
[HASHES]: Liste von Strings, beliebige Anzahl
    GUI Feldname: "Known Artifact Hashes"
[DATAQ]: Enumeration
    GUI Feldname: "Data Approval State"
[CHANGESTATUS]: Enumeration
    GUI Feldname: "OSS Usage"
```

Example Project:
- workflowsteps are not activated and commented out, since they cause authorization issues due to the client id and password


maxhbr:
@neubs-bsi is the label still valid?

neubs-bsi:
@maxhbr No. The label isn't valid anymore. This can be reviewed.

example-projects/example-project/pom.xml
```diff
@@ -88,7 +88,42 @@
                                     </entry>
                                 </configuration>
                                 <deactivated>true</deactivated>
-                            </step>
+                            </step><!--
```
maxhbr: commented code. Why not uncomment since deactivated is `true`

example-projects/example-project/pom.xml
```diff
@@ -88,7 +88,42 @@
...
+                            <step>
+                                <name>SW360 Enricher</name>
+                                <classHint>org.eclipse.sw360.antenna.sw360.workflow.processors.SW360Enricher</classHint>
+                                <configuration>
+                                    <entry>
+                                        <entryKey>rest.server.url</entryKey>
+                                        <entryValue>https://sw360.osmi.de.bosch.com/resource/api</entryValue>
```
maxhbr: proprietary URL

core/frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testing/AbstractAntennaFrontendTest.java
```diff
@@ -57,7 +57,7 @@
     @Parameterized.Parameters(name = "{index}: Test data = {1}")
     public static Collection<Object[]> data() throws URISyntaxException {
         String relativePathToP2Product = "../../../../modules/p2/antenna-p2-product/repository_manager";
-        if (! Paths.get(ClassCodeSourceLocation.getClassCodeSourceLocationAsString(AbstractAntennaFrontendTest.class),
+        if (!Paths.get(ClassCodeSourceLocation.getClassCodeSourceLocationAsString(AbstractAntennaFrontendTest.class),
```
maxhbr: unnecessary formatting

maxhbr:
Label WIP due to confusing behaviour in workflow configuration of SW360 step

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -82,13 +83,170 @@ public void configure(Map<String, String> configMap) throws AntennaConfiguration
...
+    private void addClearingStateIfAvailable(Artifact artifact, SW360Release release) {
+        if (release.getClearingState() != null && !release.getClearingState().isEmpty()) {
+            artifact.addFact(new ArtifactClearingState(
+                    ArtifactClearingState.ClearingState.valueOf(release.getClearingState())));
+        }
+    }
+
+    private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact artifact) {
+        if (sw360Release.getExternalIds() != null) {
+/*            Map<String, String> externalIds = sw360Release.getExternalIds();
```
maxhbr: commented code

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -82,13 +83,170 @@ public void configure(Map<String, String> configMap) throws AntennaConfiguration
...
+                        }
+                    }); */
+            mapCoordinates(sw360Release, artifact);
+            mapDeclaredLicense(sw360Release, artifact);
+            mapObservedLicense(sw360Release, artifact);
+            mapOriginalRepository(sw360Release, artifact);
+            mapSWHRepo(sw360Release, artifact);
+            mapHashes(sw360Release, artifact);
+            mapChangeStatus(sw360Release, artifact);
+            // TODO add mapCopyrights here, once decided how to
```
maxhbr: TODO -> Issue

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -88,4 +94,56 @@
...
+        while(it.hasNext()){
+            setMap.put(SW360Attributes.RELEASE_EXTERNAL_ID_HASHES + i, it.next());
+            i++;
+        }
+        return setMap;
+    }
+
+    private static String convertSetOfStringsToString(Set<String> stringSet) {
+        if(stringSet.isEmpty()) {
+            return null;
```
maxhbr: why not `""` as valid default?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -88,4 +94,56 @@
...
+        }
+
+        StringBuilder stringBuilder = new StringBuilder();
```
maxhbr:
```suggestion
        return stringSet.stream().collect(Collectors.joining("; "));
```
(or something like that)

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -88,4 +94,56 @@
...
+
+        stringSet.forEach(string ->
+                stringBuilder.append(string).append("; ")
```
maxhbr: Why `"; "` instead of `";"` for easier parsing?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -88,4 +94,56 @@
...
+        int i = 1;
+        Iterator<String> it = stringSet.iterator();
+        while(it.hasNext()){
+            setMap.put(SW360Attributes.RELEASE_EXTERNAL_ID_HASHES + i, it.next());
+            i++;
+        }
+        return setMap;
+    }
+
+    private static String convertSetOfStringsToString(Set<String> stringSet) {
```
maxhbr: This needs escaping of the separator `;` within the strings. Maybe use utils which already can this, e.g. a CSV library?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -82,13 +83,170 @@ public void configure(Map<String, String> configMap) throws AntennaConfiguration
...
+        Set<String> keySet = sw360Release.getExternalIds().keySet();
+
+        for (Class<? extends ArtifactCoordinates> key : SW360CoordinateKeysToArtifactCoordinates.getKeys()) {
+            String coordinateType = SW360CoordinateKeysToArtifactCoordinates.get(key);
+            if (keySet.contains(coordinateType)) {
+                String coordinateValue = sw360Release.getExternalIds().get(coordinateType);
+                String[] splitCoordiantes = coordinateValue.split(":");
+
+                ArtifactCoordinates coordinates;
+                if(splitCoordiantes.length == 3){
```
maxhbr: inconsistent formatting:
```suggestion
                if (splitCoordiantes.length == 3) {
```

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/rest/workflow/processors/SW360EnricherTest.java
```diff
@@ -46,6 +47,11 @@
     private SW360Enricher sw360Enricher;
     private SW360MetaDataReceiver connector;
 
+    private String sourceUrl = "source_url.com";
+    private String releaseTagUrl = "releaseTagUrl.com";
+    private String swhURL = "software_heritage.com";
```
maxhbr: To have tests which are nearer to reality:
```suggestion
    private String swhURL = "https://archive.softwareheritage.org/browse/release/ae93ff0b4bdbd6749f75c23ad23311b512230894/";
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -21,6 +22,17 @@
     private String cpeid;
     private String downloadurl;
     private Set<String> mainLicenseIds;
+    private Map<String, String> coordinates;
+    private Set<String> declaredLicense;
+    private Set<String> observedLicense;
+    private String releaseTagUrl;
+    private String softwareHeritageUrl;
```
maxhbr: what is the situation of the idea of saving the ID instead of the URL?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360CoordinateKeysToArtifactCoordinates.java
```diff
@@ -0,0 +1,83 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class SW360CoordinateKeysToArtifactCoordinates {
+    private static Map<Class<? extends ArtifactCoordinates>, String> sw360CoordinateKeysToArtifactCoordinates = new HashMap<>();
+
+    private static void setMap() {
```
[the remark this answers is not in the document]
neubs-bsi: I personally prefer lazy initialization. It ensures it is only instantiated when actually used at runtime while in static initialization it is used once class is loaded. 
Alternatively, I could also make it into an enum. What do you think?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
neubs-bsi: do you mean for the software Heritage URL?

maxhbr: yes, at some point there was the discussion whether to safe the ID of a release instead of a generic URL and I am not sure what the implications here are.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static boolean isValid(String swhID) {
+        List<String> objectTypes = Stream.of("snp", "rel", "rev", "dir", "cnt").collect(Collectors.toList());
+
+        String[] swhElements = swhID.split(":");
+
+        if (swhElements.length == 4
+                || swhElements[0].equals("swh")
+                || isParsable(swhElements[1])
+                || objectTypes.contains(swhElements[2])
```
maxhbr: only `rel` for release should be valid here?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,60 @@
...
+                || swhElements[3].length() == 40) {
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    private static boolean isParsable(String string) {
```
maxhbr:
```suggestion
    private static boolean isParsableInteger(String string) {
```

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,60 @@
...
+        return "Software Heritage URL";
+    }
+
+    @Override
+    public Class<? extends ArtifactFact> getKey() {
+        return ArtifactSoftwareHeritageID.class;
+    }
+
+    public static boolean isValid(String swhID) {
+        List<String> objectTypes = Stream.of("snp", "rel", "rev", "dir", "cnt").collect(Collectors.toList());
```
maxhbr: why not as static final field?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,60 @@
...
+package org.eclipse.sw360.antenna.model.artifact.facts;
+
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFactWithPayload;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ArtifactSoftwareHeritageID extends ArtifactFactWithPayload<String> {
```
maxhbr: This class requests a complicated string. It would be good if one would add here a `public String getExpectedFormat()` which returns the expected format as RegEx or Wildcard expression. So that this can be displayed in the error message.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getFactContentName() {
+        return "Software Heritage URL";
+    }
+
+    @Override
+    public Class<? extends ArtifactFact> getKey() {
+        return ArtifactSoftwareHeritageID.class;
+    }
+
+    public static boolean isValid(String swhID) {
```
maxhbr: currently this is optional to check. I would like to make that required, e.g. by the builder pattern.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private static void setOriginalRepo(SW360Release release, Artifact artifact) {
+        artifact.askForGet(ArtifactReleaseTagURL.class)
+                .ifPresent(release::setReleaseTagUrl);
+    }
+
+    private static void setSwhId(SW360Release release, Artifact artifact) {
+        artifact.askForGet(ArtifactSoftwareHeritageID.class)
+                .ifPresent(swhId -> {
+                    if(ArtifactSoftwareHeritageID.isValid(swhId)) {
```
maxhbr: There should be logging, to tell the user that he has used the wrong format

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
```diff
@@ -0,0 +1,69 @@
...
+  | Name | Name of the release, which should equal the component name. E.g. `junit:junit` |
+  | Version | Version of the release. E.g 4.8.1, which would come to the component `junit:junit:3.8.1` |
+  | CPE ID | Common Platform Enumeration ID |
+  | Download Url | The link where the source code can be found. Similar as a "Source Url" |
+  | Main License IDs | IDs of the final licenses. The IDs have been extracted from the SPDX license format. |
+  | Coordinates | These are the corresponding ${docName} coordinates that are saved under an external ID in the release object. It is a list of key-value pairs of Strings. E.g. "mvn junit:junit:3.8.1" |
+  | Declared License | The licenses declared by the release in the SPDX license format. |
+  | Observed License | The licenses observed during the ${docName} run in the SPDX license format. |
+  | Release Tag Url | This is a link to the release tag in the OSS Repo |
+  | Software Heritage ID | A link to a release ID in software heritage in the format "swh:1:rel:*" |
```
maxhbr:
```suggestion
  | Software Heritage ID | A release ID in software heritage in the format "swh:1:rel:*" |
```

maxhbr:
once the pr https://github.com/eclipse/sw360/pull/583 is merged, the `additionalData` fields should be used for some of the data

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/dependencies.csv
```diff
@@ -1,2 +1,2 @@
-Artifact Id,Group Id,Version,License Short Name,License Long Name,File Name,Coordinate Type,Observed License Short Name,Observed License Long Name,Source URL,Release Tag URL,Software Heritage URL
-commons-csv,org.apache.commons,1.4,Apache-2.0,Apache Software License 2.0,commons-csv.jar,mvn,Apache-2.0,Apache Software License 2.0,http://archive.apache.org/dist/commons/csv/source/commons-csv-1.4-src.zip,https://github.com/apache/commons-csv/tree/csv-1.4,archive.apache.org/dist/commons/csv/source/commons-csv-1.4-src.zip.sha1
+Artifact Id,Group Id,Version,License Short Name,License Long Name,File Name,Coordinate Type,Observed License Short Name,Observed License Long Name,Source URL,Release Tag URL,Software Heritage ID
```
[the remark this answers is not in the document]
neubs-bsi: Would it be okay to move those changes to the already open issue about adaptions in the CSV Analyzer? Since this is primarily a PR for sw360 functionality. 
The changes here have only been made due to changes in the model of artifact facts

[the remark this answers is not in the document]
blaumeiser-at-bosch: Sounds okish, since I already said that this PR is pretty big.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360KeysToAntennaClasses.java
```diff
@@ -0,0 +1,47 @@
+/*
```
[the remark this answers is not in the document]
maxhbr: That is not a singleton. That is just a not optimally initialized static and immutable map.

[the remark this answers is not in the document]
neubs-bsi: I realized I wasn't using this class anyways. It was a left over from earlier implementation tries. So it can be deleted.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360CoordinateKeysToArtifactCoordinates.java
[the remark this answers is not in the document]
maxhbr: That is not a singleton. That is just a poorly initialized static, final and (hopefully) immutable map.

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/dependencies.csv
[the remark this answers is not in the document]
maxhbr: > Why do we have License long names here? We need to reference the licenses, the id is therefore good enough, but we need the expression on how the licenses are related!

We need the long name if it is a license which is not known to the knowledgebase.

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
```diff
@@ -0,0 +1,69 @@
...
+ An individual release is a release of a component, meaning it always has a version and always has a component it belongs to.
+ All license and copyright information can be found here.
+
+  | Variable Name         | Meaning                       |
+  |-----------------------|-------------------------------|
+  | Component ID | ID of the component this release belongs to. E.g. The ID of the Junit component |
+  | Name | Name of the release, which should equal the component name. E.g. `junit:junit` |
+  | Version | Version of the release. E.g 4.8.1, which would come to the component `junit:junit:3.8.1` |
+  | CPE ID | Common Platform Enumeration ID |
+  | Download Url | The link where the source code can be found. Similar as a "Source Url" |
```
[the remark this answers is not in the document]
maxhbr: `Download Url` isn't the URL to the (compiled) artifact?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360CoordinateKeysToArtifactCoordinates.java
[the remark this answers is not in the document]
maxhbr: But the lazy initialization is still very ugly, consumes more lines of code and I see only disadvantages. Especially 

> It ensures it is only instantiated when actually used at runtime while in static initialization it is used once class is loaded.

is not an advantage in my opinion.

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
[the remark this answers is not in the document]
blaumeiser-at-bosch: The url to the compiled, binary artifact, really? I see this as the link to the potentially curated sources stored in a safe accessible place, e.g., internally. For compliance reason, I need access to the sources corresponding to the artifact, the real used dependency is by coordinates or hash codes uniquely defined.

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/dependencies.csv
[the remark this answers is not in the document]
blaumeiser-at-bosch: How is this saved then? The component does not provide the means to store that. Is a new license created in this case? I think we could separate this into license creation and component creation.

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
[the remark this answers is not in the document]
maxhbr: That was just purely based on the name. If I see `Download Url` of a release, I expect to find there the release instead of the sources or curated sources.

Maybe the name is not descriptive enough?

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/dependencies.csv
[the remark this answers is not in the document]
maxhbr: Our artifacts in the antenna execution contain facts with content `License extends LicenseInformation`. These contain the following fields:
- name
- longName
- text
- threatGroup
- classification.

We do not actually create licenses shared between artifacts in an antenna execution.

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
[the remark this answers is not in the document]
blaumeiser-at-bosch: Right, but being not descriptive helps using the tool in a different context :-). I see no big value in storing the url to the binary dependency, since it is typically retrieved from someplace else and the entry to the process. But a link to the "official" uses is something needed for processing.

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/dependencies.csv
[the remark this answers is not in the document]
blaumeiser-at-bosch: I had too much the SW360 view in my mind. You are right, we can handle licenses in Antenna runs when we use it for document generation. But I doubt that transporting a license text in CSV is no fun.

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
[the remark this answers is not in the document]
neubs-bsi: Would this be more adequate?: 
`  | Download Url | Link of a source archive, typically from a server, which is under the organizations control, e.g. an internal Nexus instance. |`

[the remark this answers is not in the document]
blaumeiser-at-bosch: Proposal: Link to the source archive. E.g., used to aggregate the sources of the used OSS components for the FOSS compliance bundle

antenna-documentation/src/site/markdown/sw360-data-model.md.vm
```diff
@@ -0,0 +1,69 @@
...
+  | Main License IDs | IDs of the final licenses. The IDs have been extracted from the SPDX license format. |
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Hm, I still miss the Effective License Expression here 😕

core/model/src/main/java/org/eclipse/sw360/antenna/api/exceptions/AntennaInvalidArgumentException.java
```diff
@@ -0,0 +1,25 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.api.exceptions;
+
+public class AntennaInvalidArgumentException extends AntennaException {
+    public AntennaInvalidArgumentException() {
```
blaumeiser-at-bosch: I am not a fan of new Exceptions, why is this not simply an IllegalArgumentException. What value does the Antenna prefix bring at this point?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactSoftwareHeritageID.java
```diff
@@ -0,0 +1,77 @@
...
+            return new ArtifactSoftwareHeritageID(softwareHeritageID);
+        }
+
+        private boolean isValid(String swhID) {
+            String[] swhElements = swhID.split(":");
+
+            if (swhElements.length == 4
+                    || swhElements[0].equals("swh")
+                    || isParsableInteger(swhElements[1])
+                    || "rel".equals(swhElements[2])
```
blaumeiser-at-bosch: I am not sure that "rel" is the only allowed id here, rel is definitely preferable, but I did not always find a rel, but still an id, that matched the version.

core/model/src/main/java/org/eclipse/sw360/antenna/api/exceptions/AntennaInvalidArgumentException.java
neubs-bsi: Since this an Invalid Argument when it comes to Antenna argument input, an Antenna specific exception should be thrown. 
It is not like they are giving an invalid path, but rather a format that Antenna isn't supporting.

blaumeiser-at-bosch: From my experience, defining Exceptions is a bad thing to do, it is redundant code and therefore overhead. So the AntennaException is not the best decision there is, in the first place. Especially choosing AntennaException to not be a RuntimeException is from my point of view a real mistake. I know that there is a definition from the Java world about the meaning of Exception and RuntimeException, but the handling of Exception in maintainable code is simply extremely awkward. For example, sometimes you are sure that an exception cannot occur in your code, if this exception is not a runtime exception, you dirty up your code with a catch clause or you have to throw the exception down to the root caller. Both are not really nice solutions. A runtime exception can be simply omitted, the advantage is, that if for some obscure reason your environment changes and the exception can happen, you will face it with an uncaught exception which allows you to investigate, if you silently caught the exception, this you will not face.

maxhbr: I see the checked exception as an advantage and that was the reason why I suggested to use that here.

If you prefer unchecked exceptions, we should wirte guidelines and use them.

maxhbr: If it was a `IllegalArgumentException` one would have to catch `IllegalArgumentException` for handling it, which is a to broad class of exceptions and hides many programming mistakes and other problems. Thus namespacing the exceptions helps us to only catch exceptions to which we are also responsible for throwing.

blaumeiser-at-bosch: There are better argumentations against checked exceptions:

- https://phauer.com/2015/checked-exceptions-are-evil/
- https://www.javaworld.com/article/3142626/are-checked-exceptions-good-or-bad.html

But for this PR my proposal is to get rid of the extra exception and use the standard AntennaException for now. I do not see the value of an extra exception.

One point on our last remark @maxhbr. There is no point in catching only the exception coming from us, if in a call an IllegalArgumentException is raised, this means that there is a serious issue in the call stack which has not been handled. Most likely the operation is not executed properly, independent of whether in our code or not. IllegalArgumentException and IllegalStateException have been introduced to explicitely express this. They are used e.g. in Guava.Preconditions to have a standard precondition check in a method. Reinventing the wheel at this point is imho superfluous.

neubs-bsi:
@blaumeiser-at-bosch I removed the Exception and exchanged it with an AntennaException as discussed in the daily. 
would it be possible for you to run your test sw360 project once from your machine with the changes to ensure it's working? I have certificate problems, again.

core/model/src/main/java/org/eclipse/sw360/antenna/api/exceptions/AntennaInvalidArgumentException.java
maxhbr: > There is no point in catching only the exception coming from us, ...

We know the cause of our specific exception and can handle it by logging a warning. I think that code of the kind
```java
try {
    ...
} catch (IllegalArgumentException e) {
    LOGGER.warn(e.getMessage());
}
```
is bad practice since it hides many more unexpected errors, which we do not want to handle here but not catch at all.

blaumeiser-at-bosch:
I merge this although there is a broken traveis build. The travis build is currently broken in general the observations in the build have nothing to do with the code and Jenkins reports a green build

PR #197, opened 2019-06-19
feat(fact): introduce possibility to have multiple filenames

neubs-bsi:
Corresponding Issue: #19 

##### Current Solution: 
- `ArtifactHashes` with Set of Strings with hashes is desired, but only one can be taken from `ArtifactFilename`

##### Improvement:
- Use the hash from `ArtifactFileName`
- Make it possible to have several `ArtifactFileNames` within one artifact
- redo mapping from response to artifact 
- redo mapping from artifact to sw360release
- update `SW360EnricherTest` and `SW360UpdaterTest`


neubs-bsi:
\CC @maxhbr @sschuberth

PR #198, opened 2019-06-19
Add Static code analysis via PMD and satisfy basic rules

maxhbr:


maxhbr:
\CC @sschuberth, @neubs-bsi

PR #204, opened 2019-07-03
doc: update and clarify README

maxhbr:


README.md
```diff
@@ -14,18 +14,21 @@ Learn more about Antenna in [What Antenna Does](antenna-documentation/src/site/m
 
 Please note that some dependencies of SW360antenna are only available for Java 8. So you need to use Java 8 to build the project.
 
-If you want to build antenna without the p2 parts, use
+If you want to build antenna just use maven, e.g.
```
sschuberth: Nit: While at it, should we capitalize "Antenna" and "Maven"? "mvn", however, of course needs to stay lower-case.

README.md
```diff
@@ -14,18 +14,21 @@ Learn more about Antenna in [What Antenna Does](antenna-documentation/src/site/m
...
 
 <pre>
 $ <b>mvn install</b>
 </pre>
 
-Otherwise, you can use the script `./modules/p2/prepareDependenciesForP2.sh` and afterwards call `mvn install`.
-To undo this one has the script `./modules/p2/cleanupDependenciesForP2.sh`.
-
+#### Optional Profilels
```
sschuberth: "Profiles"

README.md
```diff
@@ -14,18 +14,21 @@ Learn more about Antenna in [What Antenna Does](antenna-documentation/src/site/m
...
 You can activate the following optional profiles:
-- `-P ci`: activates also the optional profile for which includes the system tests. 
-- `-P it`: activates also the optional profile for integration testing. 
+- `-P integration-test`: activates also the optional profile for integration testing in the sw360 module
+- `-P site-tests`: which activates the site tests in `./documentation/`
+
+#### Optional Modules
+By default the p2-resolver at `./modules/p2/p2-resolver/` is not included in the build.
```
sschuberth: This still lacks a rationale *why* it is excluded by default, or more importantly, *why* would I want / need to build it?

README.md
```diff
@@ -14,18 +14,21 @@ Learn more about Antenna in [What Antenna Does](antenna-documentation/src/site/m
...
+To change that, one can for example call the corresponding prepare script `./modules/p2/prepareDependenciesForP2.sh` (without Bash support one has to follow the steps in the script by hand).
```
sschuberth: Why "for example"? What are my other options? If there are no (feasible) ones, just omit "for example".

README.md
```diff
@@ -14,18 +14,21 @@ Learn more about Antenna in [What Antenna Does](antenna-documentation/src/site/m
...
+To undo this one has the script `./modules/p2/cleanupDependenciesForP2.sh`.
```
sschuberth: Maybe better: "To remove P2 dependencies again you can use the script ..."?

README.md
sschuberth: Also, I'd appreciate a *short* explanation that "P2" is / stands for.

README.md
maxhbr: There is the alternative of calling the commands by hand (e.g. as it is required on systems where the bash script can not be executed).

README.md
maxhbr: > This still lacks a rationale _why_ it is excluded by default, or more importantly, _why_ would I want / need to build it?

It is excluded by default, since it is complicated to build (needs the prepare step), it is unnecessary in the most cases and the tests fail on some systems (e.g. on alpine or on Linux without glibc at the expected location).

sschuberth:
Thanks for the clarifications!

PR #208, opened 2019-07-08
bindings: Do not use the deprecated "lang3" package for StringEscapeUtils

sschuberth:
StringEscapeUtils was moved to the "text" package. Use a version of
"text" that was released at about the same time as the used "lang3"
version.

Signed-off-by: Sebastian Schuberth <sebastian.schuberth@bosch-si.com>

### DIFF
--- a/core/frontend-stubs/maven-frontend-stub/pom.xml
+++ b/core/frontend-stubs/maven-frontend-stub/pom.xml
@@ -98,6 +98,10 @@
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
             <artifactId>commons-csv</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


core/model/pom.xml
```diff
@@ -37,6 +37,11 @@
             <artifactId>commons-lang3</artifactId>
```
maxhbr: Why not drop the deprecated dependency?

sschuberth: Because not the whole package is deprecated; it's still valid e.g. for `StringUtils`, but `StringEscapeUtils` was indeed moved.

core/frontend-stubs/maven-frontend-stub/pom.xml
```diff
@@ -96,6 +96,10 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
```
maxhbr: Why is the dependency here necessary?

sschuberth: Because otherwise `org.eclipse.sw360.antenna.frontend.stub.mojo.DependencyTest` failed.

PR #210, opened 2019-07-12
Fix fasterjackson version

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Perhaps, but sometimes it is not worth being päpstlicher als der Papst 😃 
Both are pretty small changes and in the commit history they will still be there, except if they are squashed during merge, what we should not do in this case.

blaumeiser-at-bosch:
Have separated the two aspects

blaumeiser-at-bosch:
Merged although Travis is red, problem is flaky Travis build, no change related issue

PR #211, opened 2019-07-12
Add spotbugs to CI

maxhbr:


maxhbr:
locally all spotbugs tests are green. But I am not sure whether I like all the changes I made in this PR and I would be happy to get feedback and suggestions for rule-changes or better implementations

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
```diff
@@ -43,20 +44,18 @@
 public class JsonReader {
     private static final Logger LOGGER = LoggerFactory.getLogger(JsonReader.class);
     private static final String COMPONENTS = "components";
-    protected final Charset encoding;
-    protected final Path recordingFile;
+    private final Charset encoding;
```
bs-ondem: With setting the modifier to private, the subclasses cannot access it.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ComponentClient.java
```diff
@@ -41,7 +44,9 @@ public SW360Component getComponent(String componentId, HttpHeaders header) throw
                 new ParameterizedTypeReference<Resource<SW360Component>>() {});
 
         if (response.getStatusCode() == HttpStatus.OK) {
-            return response.getBody().getContent();
+            return Optional.ofNullable(response.getBody())
+                    .orElseThrow(() -> new AntennaException("Body was null"))
```
bs-ondem: I think the exception message is not helpful enough..

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
maxhbr: oh, the subclass in an extensions module, this means I have to add an annotation here. (I think I have set this to private more than five times now...)

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ComponentClient.java
maxhbr: One sees in which line and in which function the exception was thrown. Isn't that enough?

bs-ondem: What about the `componentId`?

maxhbr: should be fixed by: https://github.com/eclipse/antenna/compare/fc80ca9edee7309d10ea152d907a701546158c32..67b9c99cb3a76cf8a7433afaa3c811268322b680

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
maxhbr: should be fixed by: https://github.com/eclipse/antenna/compare/fc80ca9edee7309d10ea152d907a701546158c32..67b9c99cb3a76cf8a7433afaa3c811268322b680

PR #213, opened 2019-07-17
Changed api call of filename guessing in ArtifactFilename

blaumeiser-at-bosch:
Simple implementation using the first entry, perhaps we come up with a better heuristic later. Basically working when there is only one entry in the model, which is fine for many use cases

### DIFF
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactFilename.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactFilename.java
@@ -120,11 +120,10 @@ public class ArtifactFilename implements ArtifactFact<ArtifactFilename>, Artifac
                 .collect(Collectors.toList());
     }
 
-    public Optional<String> getBestFilenameGuess() {
+    public Optional<ArtifactFilenameEntry> getBestFilenameEntryGuess() {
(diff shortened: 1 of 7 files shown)


core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactFilename.java
```diff
@@ -120,11 +120,13 @@ public ArtifactFilename(String filename, String hash, String hashAlgorithm) {
...
-                .map(ArtifactFilenameEntry::getFilename)
-                .filter(Objects::nonNull)
-                .max(Comparator.comparing(String::length));
+    /*
+     * Current implementation uses a simple approach, the first entry in the set. This works in principle in use
+     * cases where there is only one entry in the model.
+     * Be aware of the limitation.
+     */
+    public Optional<ArtifactFilenameEntry> getBestFilenameEntryGuess() {
+        return artifactFilenameEntries.stream().findFirst();
```
maxhbr: Why have you dropped the heuristics to use the the longest name? this even might return an entry which only contains a hash and a filename set to `null`

blaumeiser-at-bosch:
I merge this PR although travis build is red, but travis is currently broken, the red parts show the usual broken things in travis, and Jenkins shows this build as green.

PR #215, opened 2019-07-17
Clean up exception handling in Antenna

blaumeiser-at-bosch:
Reasoning: Maintenance of an api with checked exceptions is tedious and easily leads to breaking changes, that make using components non-compilable. Also the handling of checked exceptions in general is tedious and error prone due to the necessity to declare exceptions that are not thrown in case of interfaces and inheritance and to catch them even if they are not thrown. So the big advantage of checked exceptions that a missing catch is detected shows as a two-sided sword with situations where it is good and situations in which the compiler enforces bad code. 

Also working with streams is difficult with checked exceptions.

To sum up, imho the disadvantages exceed the advantages to a large degree.


maxhbr:
please update the PR title

blaumeiser-at-bosch:
Short summary of a first evaluation of the Exception situation:

We have four remaining Exception types:
1. AntennaException
2. AntennaExecutionException (which is already a RuntimeException)
3. AntennaConfigurationException
4. AntennaComplianceException

Observations:
1. The usage of the exceptions is not unique throughout the codebase: AntennaException are used in places where a AntennaConfigurationException might be adequate.
2. AntennaException is in many places used to wrap an IOException which should not be propagated. Using here a checked exception is strange, because the checked IOException is channeled through another checked exception to get rid of the throws clauses for the one exception, but no real handling is observed. From my point of view a clear RuntimeException. Also in the other places, this indicates often major failures, which can not be recovered easily.
3. AntennaConfigurationException is a classical RuntimeException, a general interface which declares a Exception which is only thrown in a minimal number of implementations. Interestingly, sometimes also a IllegalArgumentException is used which is not caught at all. The question is, if this should not be a case for the IllegalArgumentException in general.
4. AntennaExecutionException is a RuntimeException already and should stay so.
5. AntennaComplianceException should remain a checked exception and become a independent exception not derived from AntennaException, perhaps we should get rid of the Antenna Prefix.
6. The only place I found in my check is indeed the Software Heritage id check, which could be indicated with a check exception, perhaps even something new, like a InputDataException.

@maxhbr,  @bs-ondem , @neubs-bsi , @sschuberth  fyi

blaumeiser-at-bosch:
Solves https://github.com/bsinno/osm-antenna/issues/788

blaumeiser-at-bosch:
Correction to above assessment.
The AntennaComplianceException does not make sense. It is only there to indicate that the build has to abort due to fail causing results. That is a normal exception and has nothing to do with compliance as a reason. A fail causing exception is in principle a RuntimeException because it should only be caught way up the hierarchy or not at all.

blaumeiser-at-bosch:
closes: bsinno/osm-antenna#788

core/model/src/main/java/org/eclipse/sw360/antenna/api/exceptions/ExecutionException.java
```diff
@@ -13,19 +13,19 @@
 /**
  * Used when a step of the Antenna execution workflow unexpectedly fails.
  */
-public class AntennaExecutionException extends RuntimeException {
+public class ExecutionException extends RuntimeException {
```
maxhbr: Some might confuse the name `ExecutionException` with the `ExecutionException` from `java.util.concurrent`. But I have also no suggestion for a better name.

blaumeiser-at-bosch: I have thought about this when I recognized the equal names. Since out ExecutionException is basically thrown within Antenna, in which we do not use java.util.concurrent, imho the name clash should not be that important, because it should not happen in the same code. If you are using an IDE, the IDE typically supports distinguishing the two different types.

blaumeiser-at-bosch:
> Overall it looks good. But I have a question regarding the declaring unchecked exceptions in the method signature. At some places we are declaring them in the method signature, but in some places not. Shouldn't we notify that?

Thepoint with RuntimeException is, that you can declare them, but you do have to. So the general idea is to declare them in the "official" API methods, indicating to a user of the API that he could consider them at that point or ignore them and let them fall through. In our case that means, e.g.,  the workflow step interface should declare the exception because the implementor of a workflow step should understand what is expected from him when he wants to express failure. In the single workflow step, this might not be necessary, but it could. 

The declaration is a means of documenting the functionality. And like in all documentation, the question is, where does it help understand what is happening and to react properly and where is it superfluous information that can better be detected by reading the code.

neubs-bsi:
Is this PR still worked on?

blaumeiser-at-bosch:
Yes, at some time in the future, good point for on hold 😃

blaumeiser-at-bosch:
@maxhbr @bs-ondem Ready for review now

core/runtime/src/main/java/org/eclipse/sw360/antenna/util/HttpHelper.java
```diff
@@ -35,17 +34,17 @@ public HttpHelper(ProxySettings proxySettings) {
...
 
-    public File downloadFile(String url, Path targetDirectory, String filename) throws IOException, FailedToDownloadException {
+    public File downloadFile(String url, Path targetDirectory, String filename) throws IOException {
         File targetFile = targetDirectory.resolve(filename).toFile();
 
         try (CloseableHttpResponse response = getFromUrl(url)) {
-            if (response.getStatusLine().getStatusCode() == HttpStatus.SC_NOT_FOUND) {
-                throw new FailedToDownloadException("File not found on URL=[" + url + "]");
+            if(response.getStatusLine().getStatusCode() == HttpStatus.SC_NOT_FOUND) {
+                throw new IOException("File not found on URL=["+url+"]");
```
bs-ondem: whitespaces before/after +

core/runtime/src/main/java/org/eclipse/sw360/antenna/workflow/stubs/AbstractComplianceChecker.java
```diff
@@ -36,28 +37,34 @@ public AbstractComplianceChecker() {
...
+    public Collection<Artifact> process(Collection<Artifact> artifacts) throws ExecutionException {
         LOGGER.info("Evaluate compliance rule set: {}", getRulesetDescription());
         IPolicyEvaluation evaluation = evaluate(artifacts);
         LOGGER.info("Rule evaluation done");
         LOGGER.info("Check evaluation results...");
-        execute(evaluation);
+        try {
+            execute(evaluation);
+        } catch (FailCausingException e) {
+            // ToDo: A FailCausingException should not end the build but should be reflected in the workflow result
```
bs-ondem: TODO

core/model/src/main/java/org/eclipse/sw360/antenna/api/workflow/AbstractAnalyzer.java
```diff
@@ -23,5 +23,5 @@
      */
     public abstract String getName();
 
-    public abstract WorkflowStepResult yield() throws AntennaException;
+    public abstract WorkflowStepResult yield() throws ExecutionException;
```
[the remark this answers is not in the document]
maxhbr: I think "**one is allowed** to add RuntimeExceptions to the method signature, **but not forced**" was one of the advertised advantages.

core/runtime/src/main/java/org/eclipse/sw360/antenna/workflow/stubs/AbstractComplianceChecker.java
maxhbr: to be grep safe and to support idea highlighting:
```suggestion
            // TODO: A FailCausingException should not end the build but should be reflected in the workflow result
```

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriter.java
```diff
@@ -86,7 +87,7 @@ public File createZip(List<Artifact> artifacts) {
                 zipOutput.flush();
                 output.flush();
             } catch (IOException e) {
-                throw new RuntimeException(e.getMessage() + "  The Sources-Zip File could not be created");
+                throw new ExecutionException(e.getMessage() + "  The Sources-Zip File could not be created", e);
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done

core/model/src/main/java/org/eclipse/sw360/antenna/api/workflow/AbstractAnalyzer.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done

core/model/src/main/java/org/eclipse/sw360/antenna/api/workflow/AbstractGenerator.java
```diff
@@ -11,16 +11,16 @@
...
-import org.eclipse.sw360.antenna.api.exceptions.AntennaException;
+import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
 
 import java.util.Collection;
 import java.util.Map;
 
 public abstract class AbstractGenerator extends ConfigurableWorkflowItem {
-    public Map<String,IAttachable> produce(ProcessingState previousState) throws AntennaException {
+    public Map<String,IAttachable> produce(ProcessingState previousState) throws ExecutionException {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done

core/model/src/main/java/org/eclipse/sw360/antenna/api/workflow/AbstractOutputHandler.java
```diff
@@ -11,10 +11,10 @@
...
 
 import org.eclipse.sw360.antenna.api.IAttachable;
-import org.eclipse.sw360.antenna.api.exceptions.AntennaException;
+import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 
 import java.util.Map;
 
 public abstract class AbstractOutputHandler extends ConfigurableWorkflowItem {
-    public abstract void handle(Map<String,IAttachable> generatedOutput) throws AntennaException;
+    public abstract void handle(Map<String,IAttachable> generatedOutput) throws ExecutionException;
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done

core/model/src/main/java/org/eclipse/sw360/antenna/api/workflow/AbstractProcessor.java
```diff
@@ -10,18 +10,18 @@
...
 
-import org.eclipse.sw360.antenna.api.exceptions.AntennaException;
+import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
 
 import java.util.Collection;
 
 public abstract class AbstractProcessor extends ConfigurableWorkflowItem {
-    final public WorkflowStepResult process(ProcessingState previousState) throws AntennaException {
+    final public WorkflowStepResult process(ProcessingState previousState) throws ExecutionException {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done

core/runtime/src/main/java/org/eclipse/sw360/antenna/util/HttpHelper.java
blaumeiser-at-bosch: Done

core/runtime/src/main/java/org/eclipse/sw360/antenna/workflow/stubs/AbstractComplianceChecker.java
blaumeiser-at-bosch: Done

blaumeiser-at-bosch:
@bs-ondem @maxhbr : Result of our discussion, please review again.

blaumeiser-at-bosch:
@bs-ondem , @maxhbr Just a kind reminder, I would like to see progress on this PR.

PR #216, opened 2019-07-17
fix(model): final license calculation considers duplicated licenses

bs-ondem:


core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
```diff
@@ -54,13 +57,33 @@ public static LicenseInformation getFinalLicenses(Artifact artifact) {
...
+                    Stream.of(declared.get().getLicenses(), observed.get().getLicenses())
+                            .flatMap(List::stream)
+                            .collect(Collectors.toSet()));
 
+            return createLicenseStatement(allLicenses);
+        }
         return declared.orElse(observed.orElse(new LicenseStatement()));
     }
+
+    private static LicenseInformation createLicenseStatement(List<License> allLicenses) {
```
blaumeiser-at-bosch: I thought about the computation, I think the best way is that if there is an OR or WITH in the either the declared or the observed statement, we do nothing, it is better to have the effective license not set instead of set wrong.

Damn it, the longer I think about it the more I think that the problem is not the general computation of the formula, the old algorithm was at least correct, but the usage of the formula. Actually we need two things:
1. A validation rule that triggers whenever there is an OR left in the formula during policy evaluation. This has to be selected by the user
2. When the formula is used for e.g. extracting the licenses for the Disclosure Document, we should ensure that every license is shown only once.

But I have to check with Robert about the issue, this will not solve his problem. Let us first clarify the real need, this is more difficult than expected and we should not come up with wrong results.

blaumeiser-at-bosch: And one more idea, we should check if the declared license is contained in the observed license, then taking the observed license is the way to go, only if this is not the case, we should combine the two with and, what do you think, Onur?

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactLicenseUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Artifact artifact = new Artifact("Test")
+                .addFact(new DeclaredLicenseInformation(licenseStatement))
+                .addFact(new ObservedLicenseInformation(licenseStatement2));
+
+        LicenseInformation finalLicenses = ArtifactLicenseUtils.getFinalLicenses(artifact);
+
+        assertEquals(3, finalLicenses.getLicenses().size());
+        assertEquals(1, finalLicenses.getLicenses()
+                .stream()
+                .filter(l -> l.getName() == apache.getName())
```
maxhbr:
```suggestion
                .anyMatch(l -> l.getName() == apache.getName());
```

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactLicenseUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+        licenseStatement2.setRightStatement(licenseStatement);
+
+        Artifact artifact = new Artifact("Test")
+                .addFact(new DeclaredLicenseInformation(licenseStatement))
+                .addFact(new ObservedLicenseInformation(licenseStatement2));
+
+        LicenseInformation finalLicenses = ArtifactLicenseUtils.getFinalLicenses(artifact);
+
+        assertEquals(3, finalLicenses.getLicenses().size());
+        assertEquals(1, finalLicenses.getLicenses()
```
maxhbr: We usually use fest assertions, instead of `org.junit.Assert`.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
maxhbr: > 1. A validation rule that triggers whenever there is an OR left in the formula during policy evaluation. This has to be selected by the user

This is something for an Independent validator or for the rule engine.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
```diff
@@ -54,13 +57,33 @@ public static LicenseInformation getFinalLicenses(Artifact artifact) {
...
         final Optional<LicenseInformation> observed = artifact.askForGet(ObservedLicenseInformation.class);
         if(declared.isPresent() && observed.isPresent()) {
-            final LicenseStatement effective = new LicenseStatement();
-            effective.setLeftStatement(declared.get());
-            effective.setRightStatement(observed.get());
-            effective.setOp(LicenseOperator.AND);
-            return effective;
-        }
+            List<License> allLicenses = new ArrayList<>(
+                    Stream.of(declared.get().getLicenses(), observed.get().getLicenses())
```
maxhbr: This flattens the license list and removes all information about `AND` or `OR` conjunctions between licenses.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
```diff
@@ -54,13 +57,33 @@ public static LicenseInformation getFinalLicenses(Artifact artifact) {
         final Optional<LicenseInformation> declared = artifact.askForGet(DeclaredLicenseInformation.class);
         final Optional<LicenseInformation> observed = artifact.askForGet(ObservedLicenseInformation.class);
         if(declared.isPresent() && observed.isPresent()) {
-            final LicenseStatement effective = new LicenseStatement();
-            effective.setLeftStatement(declared.get());
-            effective.setRightStatement(observed.get());
-            effective.setOp(LicenseOperator.AND);
-            return effective;
-        }
+            List<License> allLicenses = new ArrayList<>(
```
maxhbr: unnecessary of wrapping the license information into a mutable list.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
maxhbr:
```suggestion
    private static LicenseInformation createLicenseStatement(ArrayList<License> allLicenses) {
```
otherwise this will fail once an unmutable list is passed.

Essentially this is just a reduce, e.g.
```
   .reduce(
       LicenseStatement(),
       (lsmt1, lsmt2) -> {
           licenseStatement.setLeftStatement(lsmt1);
￼           licenseStatement.setOp(LicenseOperator.AND);
￼           licenseStatement.setRightStatement(lsmt2);
       });
```

blaumeiser-at-bosch: Exactly, that was what I meant with policy evaluation -> rule engine

maxhbr: Oh sorry, I have missed that essential part.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactLicenseUtils.java
```diff
@@ -54,13 +57,19 @@ public static LicenseInformation getFinalLicenses(Artifact artifact) {
...
-            final LicenseStatement effective = new LicenseStatement();
-            effective.setLeftStatement(declared.get());
-            effective.setRightStatement(observed.get());
-            effective.setOp(LicenseOperator.AND);
-            return effective;
-        }
+            final String declaredLicenseRepresentation = declared.get().evaluate();
+            final String observedLicenseRepresentation = observed.get().evaluate();
 
+            if (observedLicenseRepresentation.contains(declaredLicenseRepresentation)) {
```
maxhbr: Does `LicenseInformation::evaluate` sufficiently normalize the expression, so that the substring relation is correct? Or are there problems with non-deterministic ordering or unnecessary `(MIT AND MIT)` clauses?

maxhbr: In the case where
- observed is `(MIT OR GPL)` and
- declared is `GPL`

I think the substring relation is incorrect. The correct output would be `GPL` or `(GPL AND (MIT OR GPL))`, but not `(MIT OR GPL)`.

blaumeiser-at-bosch: @maxhbr This is a crude simplification to get rid of some MIT and MIT cases, but in general the mathematical expression (Declared) AND (Observed) is right. There are cornercases which we cannot handle with the above approach like (MIT AND BSD-3) AND (BSD-3 AND MIT) but since we do not lose the mathematical correctness, this is fine.

The simplification is in order to reduce the amount of findings by project teams who stumble accross strange expressions. But I did not want to enter a mathematical expression solver that simplifies the expressions.

Concerning your point with MIT OR GPL. I do not think that this is a relevant case. I have no idea, how the observed license expression might contain an OR operator because the observed license is a summary of what a license scanner detected in the sources. A license scanner can only detect an OR if it detects a file that contains an OR expression and even that requires some understanding and semantics which are in general undefined. And to conclude finally that it is on component level only an OR expression means that there are only files in the codebase that contain this OR expression. So the project declares that it uses one license but in all files in the codebase there is either no expression or the OR expression. That is really a weirdo case.

But @bs-ondem to get rid of even that case, I propose we change it to: if the observed license expression contains an "OR" operator, we do not simplify the effective license to the observed license, instead we make the standard (Declared) AND (Observed).

blaumeiser-at-bosch: BTW. Can we express WITH in the Antenna data model?

maxhbr: > But @bs-ondem to get rid of even that case, I propose we change it to: if the observed license expression contains an "OR" operator, we do not simplify the effective license to the observed license, instead we make the standard (Declared) AND (Observed).

Yes, that works

> BTW. Can we express WITH in the Antenna data model?

No, not yet.

bs-ondem:
@maxhbr @blaumeiser-at-bosch  Changed the final license calculation as discussed in the comments. Can be reviewed again.

core/model/src/main/resources/bindings.xjb
```diff
@@ -121,6 +121,7 @@
                  abstract public boolean isEmpty();
 
                  abstract public java.util.List<License> getLicenses();
+                 abstract public java.util.Set<License> getUniqueLicenses();
```
blaumeiser-at-bosch: My question here is, do we need the old getLicenses method at all. It does not provide any additional value except some redundant listings of licenses. I would rather prefer to change the getLIcenses method to have every license only once to having two methods here. Is there any usage which requires the List as return value or could we change that to Collection?

bs-ondem: I also believe that the return type of `getLicenses()` should be `Set<License>` but this can cause a lot of changes in the code. Probably we can create a new issue with this topic because it's not in the scope of this PR.

blaumeiser-at-bosch: That is the point, why I typically prefer Collection as return value, not List and also not Set, because with Collection you have the freedom of returning either a List or a Set. And the real advantage of a List that this is an ordered data structure which you can address by index is nowadays rarely used. In this case, there is no defined order on the licenses returned, so List has no real advantage.

But you are right, we should not change the return value type as of now. So my proposal is, we change the implementation to return in the List every license only once and get rid of getUniqueLicenses. There is no meaning in returning the list of licenses with duplicates. That you use a Set to make each entry unique is actually a implementation detail which should not be visible at the outside. For the implementation that means that you have to convert the Set back to a List. For the amount of licenses involved in an artifact, this should not be too much overhead.

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactTest.java
```diff
@@ -162,17 +161,20 @@ public void testGetFinalLicense() {
         License license2 = new License();
         license2.setName("license2");
         license2.setLongName("licenseNumber2");
+        License license3 = new License();
+        license3.setName("license3");
+        license3.setLongName("licenseNumber3");
```
maxhbr: I would prefer not to change the test setup and just change the assertions. And then create a new test for the new test-case.

blaumeiser-at-bosch:
Jenkins is red!

bs-ondem:
This PR can be reviewed again. The failing tests are fixed. I reset the test setup of `ArtifactTest.java` and corrected the assertions with the new behaviour of `ArtifactLicenseUtils::getFinalLicenses`.

core/model/src/main/resources/bindings.xjb
```diff
@@ -147,18 +146,14 @@
                  }
 
                  public java.util.List<License> getLicenses(){
-                    java.util.List<License> licenseList = new java.util.ArrayList<>();
-                    licenseList.add(this);
-                    return licenseList;
+                    return java.util.stream.Stream.of(this)
```
maxhbr:
```suggestion
                    return java.util.stream.Collections.singletonList(this);
```

the call of `.distinct()` on a stream with only one element is a no-op.

core/model/src/main/resources/bindings.xjb
```diff
@@ -217,9 +212,10 @@
                     if (this.isEmpty()) {
                         return new java.util.ArrayList<>();
                     }
-                    java.util.List<License> allLic = getLeftStatement().getLicenses();
-                    allLic.addAll(getRightStatement().getLicenses());
-                    return allLic;
+                    return java.util.stream.Stream.of(getLeftStatement().getLicenses(), getRightStatement().getLicenses())
```
maxhbr: Maybe filter out null entries? I am unsure whether they are possible.
```suggestion
                    return java.util.stream.Stream.of(getLeftStatement().getLicenses(), getRightStatement().getLicenses())
                        .filter(Objects::nonNull)
```

core/model/src/main/resources/bindings.xjb
```diff
@@ -217,9 +212,10 @@
...
+                            .flatMap(java.util.Collection::stream)
+                            .distinct()
+                            .collect(java.util.stream.Collectors.toList());
```
maxhbr: why not collect to  a set, instead of `.distinct().collect(java.util.stream.Collectors.toList());`.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/ArtifactForHTMLReport.java
```diff
@@ -55,6 +57,13 @@ public LicenseInformation getLicense() {
         return license;
     }
 
+    public String getLicenseAsHtmlListItem(License license) {
```
maxhbr: this is a util function in an unrelated class. Maybe move it to a util class.

core/model/src/main/resources/bindings.xjb
bs-ondem: We are checking before if left or right statement is null and return an empty list... https://github.com/eclipse/antenna/pull/216/files/aaeec8b839c625a37e1ce6a04e0780f058e92ce4#diff-c6e8267f8d4694964dc7847e6ad515d6R212

core/model/src/main/resources/bindings.xjb
bs-ondem: Becausee we want to keep the return type of getLicenses

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/ArtifactForHTMLReport.java
bs-ondem: A new util class or is `ArtifactUtils` suitable?

maxhbr: it should be something here in in this package, since it is only for this special html report valid

bs-ondem:
@maxhbr I push my changes based on your comments. You can have a look again

bs-ondem:
Squashed commits and rebased.

PR #217, opened 2019-07-18
Read 'downloadurl' in JSONReader and create fact based on that

kallesoranko:
We get component's source URL from Whitesource and would like to add that to the corresponding release in SW360. AFAIK https://github.com/eclipse/antenna/pull/196 will add this field to data model and with this PR we can use it in json analyzer.


blaumeiser-at-bosch:
@kallesoranko, there is one critical fail in https://travis-ci.org/eclipse/antenna/jobs/560376820, test failures, the other failures in Travis can be ignored, Travis currently is not working, but this one you should have a look at. Thanks!

kallesoranko:
I am not sure is that test failure related to this PR. Seems like there is an exception in tests for generating documentation.

blaumeiser-at-bosch:
Build is red due to flaky infrastructure, therefore merged although red builds are happening

Issue #218, opened 2019-07-18
CopyrightStatement has /n as delimiter, needs to be checked


neubs-bsi:
Follow Up issue, based on: https://github.com/eclipse/antenna/pull/209#discussion_r304752468

The `CopyrightStatement` class when merging, merges the two statements with a '/n' delimiter. 
This is contrary to the normal standard being '\n'. 

It should be checked whether this will be a problem anywhere when handling the ArtifactFact.

If so, it should be fixed.

### Acceptance Criteria
- [x] Every aspect of the usage is considered and checked
- [x] No user will encounter a String "Copyright 1/nCopyright 2" anywhere in output documents. 
- [x] If changes are necessary: Standard no todos left, PR, documentation is updated...

PR #219, opened 2019-07-18
fix(copyright): changed CopyrightStatement escape character and adapted usage of it in code

neubs-bsi:
#218 

Replaced it with \n to be conform to standards. 
CSV File delimiter is now "+".


blaumeiser-at-bosch:
I merge this because the travis failures are due to the broken travis build, Jenkins is green.

PR #221, opened 2019-07-18
feat(envVar): included environment variables in template renderer and gradle reads properties file

neubs-bsi:
- Added the possibilities to give a properties file via gradle that contains variable names and values that can be credentials. File is given via the gradle build file and added as a map to TemplateRenderer
- TemplateRenderer can now handle System environment variables in all frontends, either given when command is executed or already in system.
- Documentation was updated for each frontend
- Test were written
   - GradleTest: reads in properties file containing variable ANTENNATESTVARIABLE
   - AbstractAntennaFrontendTest: includes System variable and tests whether it is set or not and normal tests check if rendering is correct, otherwise tests would fail due to non-existing enum {ANTENNATESTVARIABLE}

### DIFF
--- a/core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
+++ b/core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
@@ -25,13 +25,13 @@ import org.xml.sax.SAXException;
 
 import javax.xml.parsers.ParserConfigurationException;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -126,17 +126,23 @@ public class AntennaCLISettingsReader {
         HashMap<String, Object> contextMap = new HashMap<>();
         contextMap.put("project", project);
         contextMap.put("basedir", parent.toString());
+
+        Optional.ofNullable(project.getPropertiesFile())
+                .map(this::mapProperties)
+                .ifPresent(contextMap::putAll);
+
         System.getProperties()
                 .entrySet()
                 .stream()
                 .filter(e -> e.getKey().toString().startsWith("proxy"))
                 .forEach(e -> contextMap.put(e.getKey().toString(), e.getValue().toString()));
+
         String renderedPom = tr.renderTemplateFile(pomFile, contextMap);
 
         try {
             return new XmlSettingsReader(renderedPom);
         } catch (IOException e) {
-            throw new IllegalArgumentException("IO exception when reading config: " + e.getMessage());
+            throw new RuntimeException("IO exception when reading config: " + e.getMessage());
         } catch (ParserConfigurationException | SAXException e) {
             throw new IllegalArgumentException("Problem parsing the config: " + e.getMessage());
         }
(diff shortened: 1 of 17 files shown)


assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIFrontend.java
```diff
@@ -41,6 +41,7 @@ public static void main(String[] args) {
             }
 
             AntennaCLIFrontend frontend = new AntennaCLIFrontend(pomFilePath.toFile());
+
```
maxhbr: unnecessary change in unrelated file

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
```diff
@@ -203,4 +209,22 @@ public ToolConfiguration readSettingsToToolConfiguration(XmlSettingsReader reade
...
+            prop.load(input);
+
+            return prop.entrySet()
+                    .stream()
+                    .collect(Collectors.toMap(
+                            p -> p.getKey().toString(),
+                            p -> p.getValue().toString()
+                    ));
+        } catch (IOException e) {
+            throw new IllegalArgumentException("IO exception when reading properties file: " + e.getMessage());
```
maxhbr: `IllegalArgumentException` is not correct here:

> Thrown to indicate that a method has been passed an illegal or inappropriate argument.

but the argument `propertiesFile` was valid and the function was not used incorrectly.

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
```diff
@@ -203,4 +209,22 @@ public ToolConfiguration readSettingsToToolConfiguration(XmlSettingsReader reade
...
+
+    private Map<String, Object> mapProperties(File propertiesFile) {
+        try (InputStream input = new FileInputStream(propertiesFile)) {
+            Properties prop = new Properties();
+
+            prop.load(input);
+
+            return prop.entrySet()
+                    .stream()
+                    .collect(Collectors.toMap(
```
maxhbr: Wasn't there at some point a filter which only allowed uppercase keys?

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
neubs-bsi: I don't know why this is shown as changed, but this is already like this:
https://github.com/eclipse/antenna/blob/fc457dd0829e0dfe5aad2bd151bafe1abadba8e1/core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java#L138

What do you propose otherwise?

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
neubs-bsi: This is reading in the properties file. We did not set any restrictions for those variables.

The uppercase keys you are referring to are in the TemplateRenderer for the system environment variables

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
maxhbr: Then this is also duplicated code from line 142ff

core/frontend-stubs/cli-frontend-stub/src/main/java/org/eclipse/sw360/antenna/frontend/stub/cli/AntennaCLISettingsReader.java
maxhbr: oh, sure, right.

blaumeiser-at-bosch:
Travis build broken, merge this, although there is a broken build

PR #222, opened 2019-07-18
fix(sw360): included update functionality for sw360 release

neubs-bsi:
This is a try to integrate a PATCH call into the SW360 get Or create Release process. 

Have not written a test yet. Just wanted feedback on approach. 

@maxhbr what do you think? Right approach?

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
@@ -44,6 +44,8 @@ public class SW360MetaDataUpdater {
     private String clientId;
     private String clientPassword;
 
+    private boolean updateReleases = false;
+
     public SW360MetaDataUpdater(String restServerUrl, String authServerUrl,
                                 String userId, String password, String clientId, String clientPassword,
                                 boolean proxyEnable, String proxyHost, int proxyPort) {
@@ -85,7 +87,11 @@ public class SW360MetaDataUpdater {
         } else {
             Optional<SW360Release> release = releaseClientAdapter.getReleaseByArtifact(component, artifact, header);
             if (release.isPresent()) {
-                return release.get();
+                if(updateReleases) {
+                    return releaseClientAdapter.addRelease(artifact, component, licenseIds, header);
+                } else {
+                    return release.get();
+                }
             } else {
                 throw new AntennaException("No release found for the artifact [" +
                         artifact + "]");
@@ -107,6 +113,10 @@ public class SW360MetaDataUpdater {
                         artifact + "]");
             }
         }
+    }
+
+    public void setUpdateReleases(boolean updateReleases) {
+        this.updateReleases = updateReleases;
     }
 
     public void createProject(String projectName, String projectVersion, Collection<SW360Release> releases) throws AntennaException, IOException {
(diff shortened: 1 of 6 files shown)


maxhbr:
I would propose another strategy:
- the SW360Enricher should pull all information from SW360 and add it to the artifact (e.g. with a new fact which just contains the raw result from SW360)
- the SW360Updater should behave differently depending on whether the Enricher has added information to the artifact
  - **the release is not present in SW360:** the release can be added to SW360
  - **if no "SW360 information" is there but the release is present in SW360:** the release in SW360 can not be updated
  - **there is "SW360 information" and the release is present in SW360:** merge our artifact information with the "SW360 information" and overwrite it in SW360. The merge strategy should be to determine where antenna has the knowledge (e.g. License and Copyrights) and take everything else from the "SW360 information".

neubs-bsi:
@blaumeiser-at-bosch

blaumeiser-at-bosch:
@maxhbr 
Concerning "no SW360 information": Is the not on purpose or by accident, because in this case, it should be easy to update. Nothig is in SW360 so any information to add is of value
Concerning the merging: Here we should be careful, because we have different use cases. In the case that an Open Source office wants to add information, we can claim that the information is valid and should overwrite the existing data. In the case that a project runs a scan with Antenna and finds data, SW360 data in general is more reliable. We can not distinguish that at the moment. The thing with the access rights in SW360 are of importance here, since the user the OS office uses is different from a normal project data.
But in general, it makes sense to get the information first from the enricher, so that the merge strategy can be determined based on previous data, especially the state of the data makes sense to look at, e.g., if it is already approved by OS office, than no data needs to be enhanced, if it is raw data, overwriting it could make sense.

Af damn, braindump above, this is definitely a more complex topic, we should discuss the different use cases.

maxhbr:
> Concerning "no SW360 information": Is the not on purpose or by accident, because in this case, it should be easy to update. Nothig is in SW360 so any information to add is of value

This can also be the case if the SW360Enricher was not part of the workflow, in which it is not easy to update.

> Concerning the merging: Here we should be careful, because we have different use cases. In the case that an Open Source office wants to add information, we can claim that the information is valid and should overwrite the existing data. In the case that a project runs a scan with Antenna and finds data, SW360 data in general is more reliable. We can not distinguish that at the moment. The thing with the access rights in SW360 are of importance here, since the user the OS office uses is different from a normal project data.

Maybe it would make sense to build a standalone jar for this "different use case", which uses SW360Antenna as a library. E.g. just consumes a CSV and sw360 credentials and synchronizes the information to SW360.

neubs-bsi:
As discussed:
There is now a patch functionality that can be called in the SW360Updater. 
However it is not yet activated for an antenna traditional user via a workflow configuration, because we have not yet decided on a role based patch requirement.

maxhbr:
This will be unnecessary due to the planned stand alone jar?

neubs-bsi:
@blaumeiser-at-bosch indicated that we can have it in antenna already, and just not call it in the Antenna Execution therefore only making it available to someone that uses antenna libraries, but I might have misunderstood

blaumeiser-at-bosch:
@maxhbr @neubs-bsi The general idea of the standalone tooling is, that it is a second wrapper around the basic Antenna/Tina functionality adding a Main function that execute a concrete set of functionality. But it is not a separate configuration of workflow steps, therefore it can use the functionality more direct than through the workflow step interface. This we want to take advantage of, so we introduce a first patch functionality that offers the use case needed by the OSS office tool and use this in the standalone tooling. It is not entered into the workflow step, so that it cannot be accidentally used by a standard Antenna/Tina user.

Important for me is that we do not introduce redundant code. From what I see in the current implementation which basically adds stuff, I get the impression that we did not hit this goal with the current implementation. If I am wrong, please ignore my comment, but we can have a look on the stuff if I hit a spot :-)

neubs-bsi:
I myself believe that I have introduced redundant code and duplications, however, I was not sure how big you wanted the separation to be.

blaumeiser-at-bosch:
Merge this althoug Travis is red for the known reason: Travis is broken

blaumeiser-at-bosch:
Correlates to #19

PR #223, opened 2019-07-22
csv(copyrights): copyrights now represented through new row, excel format stable

neubs-bsi:
As discussed this morning:

- Copyrights are read in with new line entries. The same way as Hashes. 
- An csv file was created with Excel that included several special case characters. All are working with the csv parser. So use case of creating csv file via Excel and have saved format be interpreted correctly is fullfilled.


blaumeiser-at-bosch:
Travis is broken, though the build result does not matter, Jenkins is green, therefore merge

PR #227, opened 2019-07-26
chore(commit_version.sh): always include p2 in releases

maxhbr:


blaumeiser-at-bosch:
Merged althoug red because Travis is broken

PR #244, opened 2019-08-07
fix(sw360): check for empty license information

neubs-bsi:
Resolves issue of having empty externalId value with key "license"

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
@@ -12,10 +12,13 @@ package org.eclipse.sw360.antenna.sw360.rest.resource.releases;
 
 import org.eclipse.sw360.antenna.sw360.rest.resource.SW360HalResource;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW360ReleaseEmbedded> {
     private String componentId;
@@ -175,7 +178,13 @@ public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW36
     }
 
     public Map<String, String> getExternalIds() {
-        return externalIds;
+        return Optional.ofNullable(externalIds)
+                .map(Map::entrySet)
+                .map(Collection::stream)
+                .orElse(Stream.empty())
+                .filter(entry -> !entry.getValue().equals(""))
+                .collect(Collectors.toMap(Map.Entry::getKey,
+                        Map.Entry::getValue));
     }
 
     public SW360Release setExternalIds(Map<String, String> externalIds) {
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
@@ -78,9 +78,9 @@ public class SW360ReleaseAdapterUtils {
 
     private static void setFinalLicense(SW360Release release, Artifact artifact) {
         LicenseInformation licenseInformation = ArtifactLicenseUtils.getFinalLicenses(artifact);
-
-        release.setFinalLicense(licenseInformation.evaluate());
-
+        Optional.of(licenseInformation.evaluate())
+                .filter(evaluatedLicenseInformation -> !"".equals(evaluatedLicenseInformation))
+                .ifPresent(release::setFinalLicense);
     }
 
     private static void setDeclaredLicense(SW360Release release, Artifact artifact) {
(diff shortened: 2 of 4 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -174,7 +175,14 @@ public SW360Release setChangeStatus(String changeStatus) {
     }
 
     public Map<String, String> getExternalIds() {
-        return externalIds;
+        if(externalIds != null) {
```
maxhbr: Alternatively (but possibly not better): 
```suggestion
        return Optional.of(externalIds)
            .map(Map::entrySet)
            .map(Collection::stream)
            .orElse(Stream.empty())
            .filter(stringStringEntry -> !stringStringEntry.getValue().equals("")
            .collect(Collectors.toMap(Map.Entry::getKey,
￼                             Map.Entry::getValue)
```

This solution returns an empty map, if external IDs is `null`, which might be better.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -78,9 +78,9 @@ private static void setCoordinates(SW360Release release, Artifact artifact) {
 
     private static void setFinalLicense(SW360Release release, Artifact artifact) {
         LicenseInformation licenseInformation = ArtifactLicenseUtils.getFinalLicenses(artifact);
-
-        release.setFinalLicense(licenseInformation.evaluate());
-
+        if(!licenseInformation.evaluate().equals("")) {
```
maxhbr: Multiple calls of `licenseInformation.evaluate()`, alternative with `Optional` could be:

```suggestion
         Optional.of(licenseInformation.evaluate())
             .filter(e -> !"".equals(e))
             .ifPresent(e -> release.setFinalLicense(e));
```

modules/sw360/src/test/java/org/eclipse/sw360/antenna/workflow/generator/SW360UpdaterTest.java
```diff
@@ -129,4 +127,47 @@ public String getLinkStr() {
         return artifact;
     }
 
+    @Test
+    public void testMainLicenseIsEmptyAndFinalLicenseNull() {
+        Artifact artifact  = new Artifact("JSON");
+        artifact.setProprietary(false);
+        artifact.addFact(new MavenCoordinates("artifactId(test)", "org.group.id", "1.2.3"));
+        artifact.addFact(new DeclaredLicenseInformation(new LicenseInformation() {
```
maxhbr: Why is here a custom inline `LicenseInformation` class used? That might implement behaviour which the actual implementations can not reproduce. Maybe test with actual `License` or `LicenseStatement`.

modules/sw360/src/test/java/org/eclipse/sw360/antenna/workflow/generator/SW360UpdaterTest.java
```diff
@@ -129,4 +127,47 @@ public String getLinkStr() {
...
+                return null;
+            }
+        }));
+
+        SW360Release release = new SW360Release();
+        SW360Component sw360Component = new SW360Component();
+        sw360Component.setName("artifactId(test)");
+        SW360ReleaseAdapterUtils.prepareRelease(release, sw360Component, Collections.EMPTY_SET, artifact);
+
+        assertThat(release.getFinalLicense() == null).isTrue();
```
maxhbr: If the `FinalLicense` is nullable, it might be better to use an `Optional` to make that explicit

maxhbr: I think there should be
```suggestion
        assertThat(release.getFinalLicense()).isNull();
```

modules/sw360/src/test/java/org/eclipse/sw360/antenna/workflow/generator/SW360UpdaterTest.java
neubs-bsi: Since in the Updater the `getLicenses` and `evaluate` functions of the Licenseinformation are called to get the information transferred to sw360, this seemed to me like the most reasonable approach to ensure that what we wanted to reach (empty License in Main License and hence empty string in evaluate are not mapped to the release) is accomplished. 

Using an actual License or LicenseStatement would not make that condition testable, I believe. Could be wrong though, open for suggestions.

maxhbr: > Using an actual License or LicenseStatement would not make that condition testable, I believe. Could be wrong though, open for suggestions.

but if it is not a reachable state with our implementation, the test is also useless?

neubs-bsi: No, I believe that also if we don'T have any license i think evaluate return "", so this should still be tested

maxhbr: but than it can be simulated by
```
artifact.addFact(new DeclaredLicenseInformation(null));
```
or
```
artifact.addFact(new DeclaredLicenseInformation(new LicenseStatement()));
```
?

neubs-bsi: I tested and new `LicenseStatement` when put as `ArtifactFact` results to true when:
 `assertThat(ArtifactLicenseUtils.getFinalLicenses(artifact).evaluate()).isEqualTo("");`
And tests still pass

So, I will change it as you suggested. Thanks

neubs-bsi:
@maxhbr could you merge this?

PR #245, opened 2019-08-07
fix(sw360): wrong function called when updating fixed

neubs-bsi:
Instead of calling `updateRelease`, `addRelease` was called and hence created a 409 Conflict Response when running same project twice. 

Error isn't occuring anymore. There's a new one now. Patch isn't functioning still. But normal getOrCreateRelease is (which is relevant for antenna users)

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
@@ -41,7 +41,7 @@ public class SW360ReleaseClientAdapter {
 
     public SW360Release updateRelease(SW360Release release, Artifact artifact, SW360Component sw360Component, Set<String> sw360LicenseIds, HttpHeaders header) throws AntennaException {
         SW360Release sw360ReleaseFromArtifact = new SW360Release();
-        SW360ReleaseAdapterUtils.prepareRelease(release, sw360Component, sw360LicenseIds, artifact);
+        SW360ReleaseAdapterUtils.prepareRelease(sw360ReleaseFromArtifact, sw360Component, sw360LicenseIds, artifact);
 
         if (release.equals(sw360ReleaseFromArtifact)) {
             return release;
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360Client.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360Client.java
@@ -18,6 +18,7 @@ import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.web.client.RestTemplate;
 
@@ -26,9 +27,11 @@ import java.net.Proxy;
 import java.util.Collections;
 
 public abstract class SW360Client {
+    private final boolean proxyUse;
     protected RestTemplate restTemplate;
 
     public SW360Client(boolean proxyUse, String proxyHost, int proxyPort) {
+        this.proxyUse = proxyUse && proxyHost != null;
         if (proxyUse && proxyHost != null) {
             SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
             Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
@@ -65,6 +68,14 @@ public abstract class SW360Client {
     }
 
     protected <T> ResponseEntity<T> doRestPATCH(String url, HttpEntity<?> httpEntity, ParameterizedTypeReference<T> responseType) {
-        return doRestCall(url, HttpMethod.PATCH, httpEntity, responseType);
+        if(proxyUse) {
+            throw new UnsupportedOperationException("The patch functionality used when updating releases does not support proxy use");
+        }
+        RestTemplate restTemplate = new RestTemplate();
+
+        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+        restTemplate.setRequestFactory(requestFactory);
+
+        return restTemplate.exchange(url, HttpMethod.PATCH, httpEntity, responseType);
     }
 }
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
@@ -10,17 +10,17 @@
  */
 package org.eclipse.sw360.antenna.sw360.rest.resource.releases;
 
+import org.eclipse.sw360.antenna.sw360.rest.resource.SW360Attributes;
+import org.eclipse.sw360.antenna.sw360.rest.resource.SW360CoordinateKeysToArtifactCoordinates;
 import org.eclipse.sw360.antenna.sw360.rest.resource.SW360HalResource;
 
-import java.util.Collection;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW360ReleaseEmbedded> {
+    private String releaseId;
     private String componentId;
     private String name;
     private String version;
@@ -41,6 +41,14 @@ public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW36
 
     private Map<String, String> externalIds;
 
+    public String getReleaseId() {
+        return releaseId;
+    }
+
+    public SW360Release setReleaseId(String releaseId) {
+        this.releaseId = releaseId;
+        return this;
+    }
 
     public String getComponentId() {
         return componentId;
@@ -172,6 +180,7 @@ public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW36
         return changeStatus;
     }
 
+
     public SW360Release setChangeStatus(String changeStatus) {
         this.changeStatus = changeStatus;
         return this;
@@ -189,6 +198,7 @@ public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW36
 
     public SW360Release setExternalIds(Map<String, String> externalIds) {
         this.externalIds = externalIds;
+        mapExternalIdsOnVariable();
         return this;
     }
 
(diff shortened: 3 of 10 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
```diff
@@ -40,8 +40,10 @@ public SW360Release addRelease(Artifact artifact, SW360Component sw360Component,
     }
 
     public SW360Release updateRelease(SW360Release release, Artifact artifact, SW360Component sw360Component, Set<String> sw360LicenseIds, HttpHeaders header) throws AntennaException {
+        release = release.mapExternalIdsOnVariable();
```
maxhbr:
```suggestion
        SW360Release mappedRelease = release.mapExternalIdsOnVariable();
```
to avoid mutation

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360Client.java
```diff
@@ -65,6 +66,11 @@ public SW360Client(boolean proxyUse, String proxyHost, int proxyPort) {
     }
 
     protected <T> ResponseEntity<T> doRestPATCH(String url, HttpEntity<?> httpEntity, ParameterizedTypeReference<T> responseType) {
-        return doRestCall(url, HttpMethod.PATCH, httpEntity, responseType);
+        RestTemplate restTemplate = new RestTemplate();
+
+        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
```
maxhbr: Why not instead change the `SimpleClientHttpRequestFactory requestFactory` the constructor to `HttpComponentsClientHttpRequestFactory requestFactory`?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -192,26 +201,69 @@ public void setCopyrights(String copyrights) {
...
-        declaredLicense = Optional.of(releaseWithPrecedence.getDeclaredLicense()).orElse(declaredLicense);
-        observedLicense = Optional.of(releaseWithPrecedence.getObservedLicense()).orElse(observedLicense);
-        releaseTagUrl = Optional.of(releaseWithPrecedence.getReleaseTagUrl()).orElse(releaseTagUrl);
-        softwareHeritageId = Optional.of(releaseWithPrecedence.getSoftwareHeritageId()).orElse(softwareHeritageId);
-        hashes = Optional.of(releaseWithPrecedence.getHashes()).orElse(hashes);
-        clearingState = Optional.of(releaseWithPrecedence.getClearingState()).orElse(clearingState);
-        changeStatus = Optional.of(releaseWithPrecedence.getChangeStatus()).orElse(changeStatus);
-        copyrights = Optional.of(releaseWithPrecedence.getCopyrights()).orElse(copyrights);
-        externalIds = Optional.of(releaseWithPrecedence.getExternalIds()).orElse(externalIds);
+        cpeid = Optional.ofNullable(releaseWithPrecedence.getCpeid()).orElse(Optional.ofNullable(cpeid).orElse(null));
```
maxhbr: every line has the same pattern, maybe extract into a function:
```
    private T fun(SW360Release release, Function<SW360Release,T> getter);
```

Then you would only have to write `cpeid = fun(releaseWithPrecedence, SW360Release::getCpeId)` which is cleaner and more readable

maxhbr: The call to `Optional.ofNullable(cpeid).orElse(null)` should be a no-op and always equal to `cpeid`, i.e.
- if `cpeid` is not null, it just returns `cpeid`
- if `cpeid` is null, it falls in the `orElse` case and returns `null` which is still equal to `cpeid`

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -123,6 +124,9 @@
             externalIds.put(SW360Attributes.RELEASE_EXTERNAL_ID_COPYRIGHTS, sw360Release.getCopyrights());
         }
 
+        if(externalIds.isEmpty()) {
+            return null;
```
maxhbr: why should `null` be better than a empty map? This is just prone to NPEs and requests for more null-checks later

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -192,26 +201,69 @@ public void setCopyrights(String copyrights) {
...
+                default:
+                    if (key.startsWith(SW360Attributes.RELEASE_EXTERNAL_ID_HASHES)) {
+                        if(hashes == null) {
+                            hashes = Collections.singleton(value);
+                        } else {
+                            hashes.add(value);
+                        }
+                    } else if (SW360CoordinateKeysToArtifactCoordinates.getValues().contains(key)) {
+                        if(coordinates == null) {
+                            coordinates = new HashMap<String, String>() {{
```
maxhbr: I think this way of initializing is deprecated (or at least it is ugly).
I would even prefer:
```suggestion
                            coordinates = new HashMap<String, String>();
                            coordinates.put(key, value);
```

(and even more I would prefer an immutable map, but that is a completely different story ;) )

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360Client.java
neubs-bsi: `HttpComponentsClientHttpRequestFactory` doesn't have a possibility to set a proxy

maxhbr: This means that this fix only works without a proxy?

maxhbr: Maybe throw an unsupported operation exception, if patch is requested with a proxy

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
neubs-bsi: when I put `orElse(cpeid)` in there, it caused a null pointer excpetion when it wasn't set. Hence I changed it to this mess, that worked.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -201,30 +210,77 @@ public void setCopyrights(String copyrights) {
...
+        softwareHeritageId = fun(releaseWithPrecedence, SW360Release::getSoftwareHeritageId);
+        hashes = fun(releaseWithPrecedence, SW360Release::getHashes);
+        clearingState = fun(releaseWithPrecedence, SW360Release::getClearingState);
+        changeStatus = fun(releaseWithPrecedence, SW360Release::getChangeStatus);
+        copyrights =  fun(releaseWithPrecedence, SW360Release::getCopyrights);
 
         return this;
     }
 
+    private <T> T fun(SW360Release release, Function<SW360Release, T> getter) {
```
maxhbr: maybe find a better name

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -201,30 +210,77 @@ public void setCopyrights(String copyrights) {
...
+        return Optional.ofNullable(getter.apply(release)).orElse(Optional.ofNullable(getter.apply(this)).orElse(null));
```
maxhbr: better formatting
```suggestion
        return Optional.ofNullable(getter.apply(release))
            .orElse(Optional.ofNullable(getter.apply(this))
            .orElse(null));
```

And should still be equal to:
```suggestion
        return Optional.ofNullable(getter.apply(release))
            .orElse(getter.apply(this));
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -123,6 +124,9 @@
             externalIds.put(SW360Attributes.RELEASE_EXTERNAL_ID_COPYRIGHTS, sw360Release.getCopyrights());
         }
 
+        if(externalIds.isEmpty()) {
+            return Collections.emptyMap();
```
maxhbr: this only returns an immutable map instead off the empty `externalIds` map? Why is this here?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -201,30 +210,77 @@ public void setCopyrights(String copyrights) {
...
 
     public boolean shareIdentifier(SW360Release releaseCompare) {
-        return this.componentId.equals(releaseCompare.getComponentId())
-                && this.name.equals(releaseCompare.getName())
-                && this.version.equals(releaseCompare.getVersion());
+        return this.name.equals(Optional.of(releaseCompare.getName()).orElse(""))
+                && this.version.equals(Optional.of(releaseCompare.getVersion()).orElse(""));
+    }
+
+    public SW360Release mapExternalIdsOnVariable() {
```
maxhbr: Why not always call this method after `setExternalIds()` is called?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -201,30 +210,77 @@ public void setCopyrights(String copyrights) {
...
+                            hashes.add(value);
+                        }
+                    } else if (SW360CoordinateKeysToArtifactCoordinates.getValues().contains(key)) {
+                        if(coordinates == null) {
+                            coordinates = new HashMap<String, String>();
+                            coordinates.put(key, value);
+                        } else {
+                            coordinates.put(key, value);
+                        }
+                        coordinates = Collections.unmodifiableMap(coordinates);
```
maxhbr: Here you make the coordinates immutable, so the second time one wants to put something (e.g. in line 253) it will fail

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360Client.java
```diff
@@ -26,9 +27,11 @@
 import java.util.Collections;
 
 public abstract class SW360Client {
+    private final boolean proxyUse;
     protected RestTemplate restTemplate;
 
     public SW360Client(boolean proxyUse, String proxyHost, int proxyPort) {
+        this.proxyUse = proxyUse;
```
maxhbr: To be consistent with the lines below:
```suggestion
        this.proxyUse = proxyUse && proxyHost != null;
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
neubs-bsi: Will return to null, as it was before- Since this is how it is handled when called.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
neubs-bsi: As I said: Returns Nullpointerexception when `(getter.apply(this));` is null

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
maxhbr: but if `externalIds` would be null, it would throw a NPE here?

maxhbr:
azure is green https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=364

PR #247, opened 2019-08-08
Port to use the ORT data model directly

sschuberth:
This removes the complexity to iterate over JsonNodes manually. See [1]
for the artifacts.

[1] https://jitpack.io/#heremaps/oss-review-toolkit


modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -10,22 +10,22 @@
  */
 package org.eclipse.sw360.antenna.ort.workflow.analyzers;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import com.here.ort.model.OrtResult;
```
neubs-bsi: Nit pick: imports are not properly separated

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/workflow/analyzers/OrtResultAnalyzer.java
```diff
@@ -49,60 +49,12 @@ public String getName() {
...
-        Artifact artifact = new OrtAnalyzerResultResolver().apply(analyzerResult);
-
-        scanResult.map(new OrtScannerResultResolver())
-                .ifPresent(artifact::overrideWith);
+        LOGGER.debug("Creating artifact list from ORT result file '" + ortResultFile + "'.");
+        OrtResult result = OutputFormatKt.mapper(ortResultFile).readValue(ortResultFile, OrtResult.class);
+        OrtResultArtifactResolver resolver = new OrtResultArtifactResolver(result);
 
-        return artifact;
+        return Objects.requireNonNull(result.getAnalyzer()).getResult().getPackages().stream()
```
neubs-bsi: Are we not using information from the scan results, ever? 
If I remember correctly the analyzer results didn't contain information for e.g. copyrights. 
Or do you regard this as unnecessary?

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/workflow/analyzers/OrtResultAnalyzer.java
sschuberth: Correct, but AFAIR they weren't before either, so I was unsure about the policy. Will fix.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/workflow/analyzers/OrtResultAnalyzer.java
sschuberth: We still use scan results, e.g. for license / copyright information, just not anymore as a fallback to get the list of artifacts. Getting the license / copyright information is now "hidden" behind the call to [collectLicenseFindings](https://github.com/eclipse/antenna/pull/247/files#diff-143acd6fd6ded1ddc78eb04e6186021eR34) which creates a map for exactly the package-to-license-findings lookup we need to do.

maxhbr: The implementation of `Objects.requireNonNull` is
```
public static <T> T requireNonNull(T obj) {
    if (obj == null)
        throw new NullPointerException();
    return obj;
}
```
and contains no additional benefit (except for throwing the NPE in an unexpected line and thus complicates the stack trace?)?

Maybe use `requireNonNull(T obj, String message)` or wrap into an `Optional.of(...).orElseThrow(...)...` or `Optional.of(...).orElse(...)...`.

modules/ort/pom.xml
```diff
@@ -42,6 +56,32 @@
...
+        <!-- ################################ flattened ORT dependencies ################################ -->
+        <dependency>
+            <groupId>com.github.heremaps.oss-review-toolkit</groupId>
+            <artifactId>analyzer</artifactId>
+            <version>jitpack-e972a4f3bb-1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.heremaps.oss-review-toolkit</groupId>
+            <artifactId>model</artifactId>
+            <version>jitpack-e972a4f3bb-1</version>
```
maxhbr: We try to define versions in the `dependencyManagement` in the top level pom. And if a version appears multiple times we extract it into a variable.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+        switch (pkg.getId().getType().toLowerCase()) {
+            case "nuget":
+            case "dotnet":
+                return Optional.of(mapDotNetCoordinates(name, version));
+            case "maven":
+                return Optional.of(mapMavenCoordinates(namespace, name, version));
+            case "npm":
+                return Optional.of(mapJavaScriptCoordinates(name, version));
+            default:
+                return Optional.of(mapSimpleCoordinates(name, version));
```
maxhbr: if `namespace != null` one could maybe use `return Optional.of(mapSimpleCoordinates(namespace + ":" + name, version));`. But that is potentially not a good heuristic.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    private static ArtifactCoordinates mapDotNetCoordinates(String name, String version) {
+        DotNetCoordinates.DotNetCoordinatesBuilder c = new DotNetCoordinates.DotNetCoordinatesBuilder();
```
maxhbr: I think all builder can be written as follows:
```suggestion
        return new DotNetCoordinates.DotNetCoordinatesBuilder()
            .setPackageId(name);
￼            .setVersion(version);
￼            .build();
```

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+        c.setVersion(version);
+        c.setArtifactId(name);
+        return c.build();
+    }
+
+    private static ArtifactCoordinates mapJavaScriptCoordinates(String name, String version) {
+        JavaScriptCoordinates.JavaScriptCoordinatesBuilder c = new JavaScriptCoordinates.JavaScriptCoordinatesBuilder();
+        c.setName(name);
+        c.setVersion(version);
+        c.setArtifactId(name + "-" + version);
```
maxhbr: Is that default behaviour? Maybe it would be good to implement that in the `JavaScriptCoordinates` as a fallback if `artifactId` is unset.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+                .map(LicenseSupport::mapLicenses)
+                .map(DeclaredLicenseInformation::new);
+    }
+
+    private static Optional<ArtifactFilename> mapFilename(Package pkg) {
+        final Stream<RemoteArtifact> artifacts = Stream.of(pkg.getSourceArtifact(), pkg.getBinaryArtifact());
+
+        return artifacts
+                .filter(a -> !a.getUrl().isEmpty())
+                .findFirst()
```
maxhbr: shouldn't there now be multiple ArtifactFilename entries possible in antenna?

modules/ort/pom.xml
sschuberth: Done.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/workflow/analyzers/OrtResultAnalyzer.java
sschuberth: Done. Turns out I was actually confusing `Objects` with `Optional` :laughing:

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
sschuberth: This code was just moved from how / where it was before my commit, so I'd like to keep this as-is for now.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
sschuberth: This was actually not possible, but I've fixed that in a separate PR: https://github.com/eclipse/antenna/pull/249

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
sschuberth: This code was just moved from how / where it was before my commit, so I'd like to keep this as-is for now.

bs-ondem:
@sschuberth The tests are also failing on my machine with maven@3.6.0, but not with maven@3.3.9. I suppose it's related to the `maven-core` dependency in maven-frontend-stub, which is also version 3.3.9. 

I couldn't find out what exactly is causing the `NoSuchMethod` error, but I compared the transitive dependencies of `maven-core:3.3.9` and `maven-core:3.6.0` and I noticed that 3.6.0 has no dependency to  `aether-api` anymore.

sschuberth:
Thanks a lot @bs-ondem for the valuable hint! After again talking to @blaumeiser-at-bosch about this, it occurred to me that this conflict should have never happened in the first place, as this PR only depends on the ORT model, not on the ORT analyzer. So, changing the dependency from `analyzer` to `model` makes the test failure go away :grin:

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
sschuberth: I don't know... maybe... probably :wink: Anyway, at this point I wanted to do a pure conversion of the original code. New features can be added later, now that it's easier to do that :grin:

blaumeiser-at-bosch: To my understanding, ORT distinguishes between binary and source artifact on a technology basis. I saw the result from Tuan and it showed in the Java case, that the binary artifact is the relevant thing, the source artifact is the packaged source jar, whereas in JavaScript, there is no binary artifact, i.e., the source artifact represents the real dependency used in the project. This code here looks like it takes randomly one of both, if both are present, that would not be the correct thing to do, or am I wrong?

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+                .map(ArtifactHomepage::new);
+    }
+
+    private Optional<ObservedLicenseInformation> mapObservedLicense(Package pkg) {
+        Collection<String> licenses = Optional.ofNullable(licenseFindings.get(pkg.getId()))
+                .map(Map::keySet)
+                .orElse(Collections.emptySet()).stream()
+                .map(LicenseFinding::getLicense)
+                .collect(Collectors.toList());
+
```
blaumeiser-at-bosch: Why are declared and observed licenses handled differently?

sschuberth: Both are valid questions, but I'd like to address them separately. This PR is only about porting the existing code (which already had the issues you pointed out) to use the ORT model directly. I.e. no change of functionality is done, just porting to another API.

For reference, see to original code at

* https://github.com/eclipse/antenna/pull/247/files#diff-9220c0f0beaa323e93534dc8c4cf5c56L68

* https://github.com/eclipse/antenna/pull/247/files#diff-5363d56ff7d691e5d496f9256fd61b73L68

sschuberth: Differently in what way?

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -0,0 +1,154 @@
...
+    private static Optional<ArtifactSourceUrl> mapSourceUrl(Package pkg) {
+        final Stream<String> urls = Stream.of(pkg.getVcsProcessed().getUrl(), pkg.getSourceArtifact().getUrl());
+
+        return urls.filter(u -> !u.isEmpty())
+                .findFirst()
+                .map(ArtifactSourceUrl::new);
+    }
+
+    private static Optional<DeclaredLicenseInformation> mapDeclaredLicense(Package pkg) {
```
blaumeiser-at-bosch: Differently because this is straight forward, more or less, I understand how declared licenses are handled

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
blaumeiser-at-bosch: Compared to my comment above, the handling of the observed information is much more difficult, and not easy to understand.

sschuberth: The ORT data structures for detected (aka observed) licenses are designed with associating them with copyright statements in mind. That's why a lookup by artifact is rather convoluted compared to the lookup of a declared license.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
sschuberth: I've addressed this now in https://github.com/eclipse/antenna/pull/270.

PR #243, opened 2019-08-06
Fix and update Jenkins

maxhbr:
- [x] TODO: adopt `./Jenkinsfile`
- [ ] fix tests in eclipse jenkins (probably out of scope, not necessary)
- [ ] fix configuration of build job https://ci.eclipse.org/antenna/job/build-pipeline/ to build master again

#### Maven Plugin
add the following to a pom with antenna
```xml
   <pluginRepositories>
     <pluginRepository>
       <id>eclipse-antenna-snapshot</id>
       <url>https://download.eclipse.org/antenna/snapshots/</url>
       <releases>
         <enabled>true</enabled>
       </releases>
       <snapshots>
         <enabled>true</enabled>
       </snapshots>
     </pluginRepository>
   </pluginRepositories>
```
and run `mvn package`.
#### CLI:
or download the cli jar from https://download.eclipse.org/antenna/snapshots/org/eclipse/sw360/antenna/cli/1.0.0-SNAPSHOT/?d and verify that it is working, e.g. via:
```
$java -jar /-mp/cli-1.0.0-20190813.075046-1.jar ./pom.xml
```

PR #252, opened 2019-08-16
Add the drools test dev project to build

blaumeiser-at-bosch:

### DIFF
--- a/modules/drools/pom.xml
+++ b/modules/drools/pom.xml
@@ -27,5 +27,6 @@
     <modules>
         <module>drools-checker</module>
         <module>rule-engine-testing</module>
+        <module>rules-test-development-project</module>
     </modules>
 </project>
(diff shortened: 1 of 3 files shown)


modules/drools/pom.xml
```diff
@@ -27,5 +27,6 @@
     <modules>
         <module>drools-checker</module>
         <module>rule-engine-testing</module>
+        <module>rules-test-development-project</module>
```
maxhbr: this also has to be blacklisted in https://github.com/eclipse/antenna/blob/b0a52865794f96a0af8f5d63165dcad4dea65ff6/Jenkinsfile.eclipse#L129-L133

blaumeiser-at-bosch: This line looks weird, because all the antenna-testing paths are not existing anymore?

maxhbr: oh yes, after https://github.com/eclipse/antenna/pull/243 has been merged it has to be blacklisted there.

PR #259, opened 2019-08-20
Policy Engine: Add basic engine infrastructure

blaumeiser-at-bosch:
- Basic engine
- Concept of rules
- Concept of executors of rules
- Concept of Artifact facade interface
- Configurator as place to hide all configuration aspects

### DIFF
--- a/modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
+++ b/modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * Interface for a generic rule with the standard fields relevant for all policy rules.
+ */
+public interface Rule {
+    /**
+     * @return The business id of the rule defined by the policy responsible
+     */
+    String getId();
+
+    /**
+     * @return The business name of the rule defined by the policy responsible
+     */
+    String getName();
+
+    /**
+     * @return The description given to the user of the policy in case of a policy violation
+     */
+    String getDescription();
+
+    /**
+     * @return The configured severity of a policy violation
+     */
+    RuleSeverity getSeverity();
+
+    /**
+     * @return The associated {@link RuleSet} from which the rule originates
+     */
+    RuleSet getRuleSet();
+}
--- a/modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSeverity.java
+++ b/modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSeverity.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * The severity levels of policy violations, this enables the user of the engine to react
+ * appropriately on a policy violation.
+ */
+public enum RuleSeverity {
+    INFO,
+    WARN,
+    ERROR;
+}
--- a/modules/policy/engine/pom.xml
+++ b/modules/policy/engine/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) Bosch Software Innovations GmbH 2019.
+  ~
+  ~ All rights reserved. This program and the accompanying materials
+  ~ are made available under the terms of the Eclipse Public License v2.0
+  ~ which accompanies this distribution, and is available at
+  ~ http://www.eclipse.org/legal/epl-v20.html
+  ~
+  ~ SPDX-License-Identifier: EPL-2.0
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.sw360.antenna</groupId>
+        <artifactId>policy-module</artifactId>
+        <version>${revision}${qualifier}</version>
+    </parent>
+
+    <artifactId>policy-engine</artifactId>
+    <name>policy-engine</name>
+
+    <packaging>jar</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <!-- ################################ testing dependencies ################################ -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- ################################ compliance dependency ########################### -->
+        <dependency>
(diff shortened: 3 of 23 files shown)


blaumeiser-at-bosch:
Remark: This is the basic infrastructure, it is not integrated into Antenna so far. But I think it is a good initial step for a first integration.

blaumeiser-at-bosch:
Azure build (green): https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=354

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
```diff
@@ -0,0 +1,41 @@
...
+public interface Rule {
+    /**
+     * @return The business id of the rule defined by the policy responsible
+     */
+    String getId();
+
+    /**
+     * @return The business name of the rule defined by the policy responsible
+     */
+    String getName();
```
sschuberth: Does it make sense to have an id *and* a name? What is the purpose of the one vs. the other? Why can't a name serve as the id?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return The description given to the user of the policy in case of a policy violation
+     */
+    String getDescription();
+
+    /**
+     * @return The configured severity of a policy violation
+     */
+    RuleSeverity getSeverity();
```
sschuberth: To me it feels a bit odd to have the severity attached to the rule. I believe a "severity" should be attached to an "issue", and a "rule" reports (an) issue(s). That way issues (with severities) could also be reported / generated by other means that rules, e.g. manually entered by users.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSeverity.java
```diff
@@ -0,0 +1,21 @@
...
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * The severity levels of policy violations, this enables the user of the engine to react
+ * appropriately on a policy violation
+ */
+public enum RuleSeverity {
+    INFO,
+    WARN,
+    FAIL;
```
sschuberth: Nit: IMO "FAIL" is not a severity, it's an action you may take *based on* the severity of an issue. (Or did you mean to says "FAILURE"?)

FYI, in ORT [we have](https://github.com/heremaps/oss-review-toolkit/blob/master/model/src/main/kotlin/Severity.kt) "ERROR", "WARNING" and "HINT".

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/TestUtils.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Helper for getting the element at index x in a collection
+     *
+     * @param data The collection from which the element should be retrieved
+     * @param index The index of the element to be retrieved
+     * @param <T> The data type used in the collection
+     * @return The element found
+     * @throws IndexOutOfBoundsException If the index is bigger than the collections size
+     */
+    public static <T> T getIndex(final Collection<T> data, final int index) {
```
sschuberth: Maybe use [CollectionUtils](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#get-java.lang.Object-int-) instead? (Assuming commons-collections is already in use; it probably does not pay-off to introduce a new dependency just for that helper.)

maxhbr:
since this PR is included in https://github.com/eclipse/antenna/pull/260, one also has to look at the comments there.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+
+/**
+ * The result object for a policy {@link Rule}. The {@link PolicyEngine} creates for each rule an object of this type
+ * and returns all {@link ThirdPartyArtifact} objects which have caused a policy violation for that rule
+ */
+public class EvaluationResult {
+    private Rule ruleForResult;
+    private Collection<ThirdPartyArtifact> failedArtifacts = new HashSet<>();
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Actually, this is a change I plan to do today, that was a misunderstanding from my analysis of the DroolsEngine. But indeed, when experimenting with multiple artifact rules, the result is not easy, also waiving violations is quite difficult, if the violation instance is lost.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
```diff
@@ -0,0 +1,77 @@
...
+
+    EvaluationResult(final Rule ruleForResult, final Collection<ThirdPartyArtifact> failedArtifacts) {
+        this.ruleForResult = ruleForResult;
+        this.failedArtifacts.addAll(failedArtifacts);
+    }
+
+    /**
+     * @return Business id of the associated rule
+     */
+    public String getId() {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Actually, we could discuss using Kotlin in new components. I have seen code using Lombok, perhaps this would be an option, though in this case the getter allows access as facade method, is this also supported in Kotlin or Lombok?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
blaumeiser-at-bosch: To be discussed. Marcels rules have both, an id like A9 and a name like in this case "License with no Threatgroup"

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
blaumeiser-at-bosch: Ok, this I do not understand, from the systems I know, you always configure the severity with a rule, e.g., in static code analysis, we always configured the rule to have an impact. Also compiler create a severity by the violated rule? How can the polica engine define a severity if it is not based on the rule?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSeverity.java
blaumeiser-at-bosch: I basically took them over from the Antenna concept. There the severity is something that impacts the build, so FAIL means, fail the build. But I understand your issue there, it is the decision of the workflow engine to decide what to do in case of an error.

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/TestUtils.java
blaumeiser-at-bosch: That was the trade of, I know of all those cool Apache utils, but they are all dependencies :-).

maxhbr: see also: https://github.com/eclipse/antenna/pull/260#discussion_r316152345 and https://github.com/eclipse/antenna/pull/260#discussion_r316148041

blaumeiser-at-bosch:
> see comments in #260

I think I covered them in my next changes, at least those applicable in this context.

blaumeiser-at-bosch:
@maxhbr @sschuberth I have abandoned this PR and maintain only #260. Please check the changes resulting from your remarks there. Thanks!

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+
+/**
+ * The result object for a policy {@link Rule}. The {@link PolicyEngine} creates for each rule an object of this type
+ * and returns all {@link ThirdPartyArtifact} objects which have caused a policy violation for that rule
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: done, see #260

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSeverity.java
blaumeiser-at-bosch: Switch FAIL to ERROR

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/TestUtils.java
blaumeiser-at-bosch: Added the CollectionUtils dependency with scope test

maxhbr: why not instead just make it a `List`? A collection where you want to be able to get the nth element is by definition a List, and if the collection would be a Set, this function has undeterministic behaviour?

PR #260, opened 2019-08-21
Add new Policy Engine that executes rules defined as Java implementation

blaumeiser-at-bosch:


blaumeiser-at-bosch:
Depends on #259

blaumeiser-at-bosch:
Azure build green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=359

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+        LOGGER.debug("Policy engine created");
+
+        return resultEngine;
+    }
+
+    private static RuleSet createRuleSet(final String ruleClassRef) {
+        try {
+            return (RuleSet) Class.forName(ruleClassRef).newInstance();
+        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            throw new IllegalArgumentException("Configuration Error: Instantiation of rule type failed: " + ruleClassRef, e);
```
maxhbr: Shouldn't it be some kind of `ConfigurationException` if it is a `"Configuration Error"`?

The semantics behind `IllegalArgumentException` is that `"Thrown to indicate that a method has been passed an illegal or inappropriate argument."` which is not 100% appropriate here. It could be thrown if the passed string can't possibly be a class, e.g. `""` or `null`.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param ruleSetRefs List of {@link RuleSet} implementations as qualified class name. The classes need to be
+     *                     available of the classpath
+     * @return The {@link PolicyEngine} instance to start policy evaluations
+     * @throws IllegalArgumentException If not all classes could be found on the classpath
+     */
+    public static PolicyEngine configure(final Collection<String> ruleSetRefs) throws IllegalArgumentException {
+        LOGGER.debug("Configuring the policy engine with ruleSetRefs " + ruleSetRefs.toString());
+
+        final Map<String, Collection<Rule>> ruleToExecutorMapping = new HashMap<>();
```
maxhbr: Maybe a MultiMap implementation would also be an alternative.

maxhbr: I see no reason why `Collection` couldn't be replaced with `List` or `Set` to make clear whether rules can be duplicates or are sorted.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+
+        ruleSetRefs.stream()
+                  .map(PolicyEngineConfigurator::createRuleSet)
+                  .flatMap(rs -> rs.rules().stream())
+                  .map(PolicyEngineConfigurator::mapRuleToExecutor)
+                  .forEach(data -> addRuleToExecutor(ruleToExecutorMapping, data.executorClazz, data.rule));
```
maxhbr: Mixture of functional and imperative style. A more consistent alternative might be to have a `.collect(...)` statement here.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            throw new IllegalArgumentException("Configuration Error: Instantiation of rule type failed: " + ruleClassRef, e);
+        }
+    }
+
+    private static class RuleToExecutor {
+        private final Rule rule;
+        private final String executorClazz;
+
+        private RuleToExecutor(final Rule rule, final String executorClazz) {
```
maxhbr:
```suggestion
        private RuleToExecutor(final Rule rule, final String executorClass) {
```
no need for the `zz`-hack here, and every time I see it I expect something of the type `Class<..>` instead of String.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
```diff
@@ -0,0 +1,48 @@
...
+
+        return executors.parallelStream()
+                 .map(executor -> executor.executeRules(thirdPartyArtifacts))
+                 .flatMap(results -> results.stream())
+                 .collect(Collectors.toSet());
+    }
+
+    public Collection<RuleSet> getRuleSets() {
+        return executors.stream()
+                        .map(RuleExecutor::getRuleSets)
```
maxhbr: unusual style of indentation again

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private static Collection<RuleExecutor> createExecutors(final Map<String, Collection<Rule>> ruleToExecutorMapping) {
+        final Collection<RuleExecutor> executors = new ArrayList<>();
+        executors.add(new SingleArtifactExecutor(ruleToSingleElementRule(ruleToExecutorMapping)));
+        return executors;
+    }
+
+    private static Collection<SingleArtifactRule> ruleToSingleElementRule(final Map<String, Collection<Rule>> ruleToExecutorMapping) {
+        return ruleToExecutorMapping
+                    .get(SingleArtifactExecutor.class.getName())
```
maxhbr: unusual style of indentation again

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
```diff
@@ -0,0 +1,41 @@
...
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * Interface for a generic rule with the standard fields relevant for all policy rules
+ */
+public interface Rule {
```
maxhbr: at some places in the code (e.g. in `model`) we would write `IRule` for an interface

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/TestUtils.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Helper for getting the element at index x in a collection
+     *
+     * @param data The collection from which the element should be retrieved
+     * @param index The index of the element to be retrieved
+     * @param <T> The data type used in the collection
+     * @return The element found
+     * @throws IndexOutOfBoundsException If the index is bigger than the collections size
+     */
+    public static <T> T getIndex(final Collection<T> data, final int index) {
```
maxhbr: Would not be necessary, if `Collection` would be a `List` ;)

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+
+/**
+ * The result object for a policy {@link Rule}. The {@link PolicyEngine} creates for each rule an object of this type
+ * and returns all {@link ThirdPartyArtifact} objects which have caused a policy violation for that rule
+ */
+public class EvaluationResult {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: No, because the Policy Engine should have no dependency to Antenna and its model in its core. An Antenna facade is added around the Policy Engine

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public Severity getSeverity() {
+        return Severity.valueOf(evalResult.getSeverity().name());
+    }
+
+    @Override
+    public Set<Artifact> getFailedArtifacts() {
+        return evalResult.getFailedArtifacts()
+                         .stream()
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/EvaluationResult.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return Configured severity of the associated rule
+     */
+    public RuleSeverity getSeverity() {
+        return ruleForResult.getSeverity();
+    }
+
+    /**
+     * @return Associated rule
+     */
+    public Rule getRule() { return ruleForResult; }
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done for all occurences

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
blaumeiser-at-bosch: I saw that, but this is actually a bad practice. An interface is a type and a type does not care whether it is an interface or not. Also for the user of the type, there is no real difference between an interface and a class, so this convention does not make sense for him. In the implementing code, this is also not helpful because it is easily visible that an interface is implemented not extended, so the I does not give any more information. Even the IDE shows nicely whether you have an interface, an enum, an exception or a class

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/TestUtils.java
blaumeiser-at-bosch: That is right, but sometimes you need a set and the interface changes :-).

If you need it in productive code, it might make sense to use a list, but using a list, so that your tests get around this, I prefer to have the tests a bit more complicated

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.antenna;
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+        return pResult;
+    }
+
+    @Override
+    public IPolicyEvaluation evaluate(final Collection<Artifact> artifacts) {
+        LOGGER.info("Policy Engine execution started");
+
+        final Collection<ThirdPartyArtifact> polengArtifacts = artifacts
+                .stream()
+                .map(artifact -> new AntennaArtifact(artifact))
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Implementation for the Antenna {@link IPolicyEvaluation} result based on the {@link PolicyEngine} data
+ */
+class PolicyEvaluation implements IPolicyEvaluation {
+    private final Set<IEvaluationResult> evalResults;
+
+    PolicyEvaluation(final Collection<EvaluationResult> data) {
+        evalResults = data.stream().map(res -> new AntennaEvaluationResult(res)).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+
+        final Collection<EvaluationResult> results = policyEngine.evaluate(polengArtifacts);
+
+        return new PolicyEvaluation(results);
+    }
+
+    @Override
+    public String getRulesetDescription() {
+        StringBuilder resultString = new StringBuilder("Policy Engine, with rule sets:\n");
+        policyEngine.getRuleSets().forEach(rs -> resultString.append(prepareRuleSetInfo(rs)));
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * The workspace step which integrates the {@link PolicyEngine} into the Antenna workflow engine
+ */
+public class PolicyEngineProcessor extends AbstractComplianceChecker {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PolicyEngineConfigurator.class);
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Ups, done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+                  .forEach(data -> addRuleToExecutor(ruleToExecutorMapping, data.executorClazz, data.rule));
+
+        PolicyEngine resultEngine = new PolicyEngine(createExecutors(ruleToExecutorMapping));
+
+        LOGGER.debug("Policy engine created");
+
+        return resultEngine;
+    }
+
+    private static RuleSet createRuleSet(final String ruleClassRef) {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: I am not convinced, I think the throws clause makes sense at the outer edges of the class, so for public methods or protected methods, so that a user of the class gets many informations without reading the code. Clutter internal structure with throws clauses does not improve readability imho.

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessorTest.java
```diff
@@ -0,0 +1,103 @@
...
+        testee.configure(configMap);
+
+        Artifact a1 = new Artifact("Source 1");
+        Artifact a2 = new Artifact("Source 2");
+
+        IPolicyEvaluation result = testee.evaluate(Arrays.asList(a1, a2));
+        Collection<String> ruleIds = result.getEvaluationResults()
+                .stream()
+                .map(IEvaluationResult::getId)
+                .distinct()
```
maxhbr:
```suggestion
                .distinct(Collectors.toSet());
```

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/engine/testdata/TestArtifact.java
```diff
@@ -0,0 +1,28 @@
...
+public class TestArtifact implements ThirdPartyArtifact {
+
+    @Override
+    public boolean artifactIdentified() {
+        return true;
+    }
+
+    @Override
+    public Collection<String> getConcludedLicenses() {
+        return null;
```
maxhbr: To avoid NPEs:
```suggestion
        return Collections.empty();
```

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSet.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Collection;
+
+/**
+ * This is a collection of {@link Rule} objects within a common context. The configuration of the {@link PolicyEngine}
+ * in the {@link PolicyEngineConfigurator} allows to give several rule sets to set up the policies.
+ */
+public interface RuleSet {
+    String name();
+    String version();
+    Collection<Rule> rules();
```
maxhbr: If it is called `RuleSet` it should behave like a set, e.g. does not contain duplicates. (or be actually `Set` instead of a `Collection`)

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
blaumeiser-at-bosch: I think that IllegalArgumentException matches because the method expects that a list of proper RuleSet class references is given. And if this is not the case, if the property is cluttered or misformatted, the argument given is not suitable. And the exception is described at the API level, so that the user has a clear picture on what to expect.

I am actually not a big fan of many different exception types, because from my experience in checking thrown exceptions, the thing I look at, is the stack trace. Which type the exception has, is typically not a big concern.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/CompareArtifactExecutor.java
```diff
@@ -0,0 +1,72 @@
...
+                .filter(rightHandSide -> hasArtifactPolicyViolation(rule, leftHandSide, rightHandSide))
+                .map(rightHandSide -> new ViolationIssue(rule, Arrays.asList(leftHandSide, rightHandSide)))
+                .collect(Collectors.toList());
+    }
+
+    private boolean hasArtifactPolicyViolation
+            (final CompareArtifactRule rule, final ThirdPartyArtifact leftHandSide, final ThirdPartyArtifact rightHandSide) {
+        try {
+            if (!leftHandSide.equals(rightHandSide)) {
+                rule.evaluate(leftHandSide, rightHandSide);
```
maxhbr: why is this realized with an Exception instead just a return value? This function just "casts" an exception back to `boolean`.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/CompareArtifactExecutor.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        } catch (PolicyException e) {
```
maxhbr: This only finds the first violation, instead of collecting all?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/CompareArtifactExecutor.java
```diff
@@ -0,0 +1,72 @@
...
+    private Collection<ViolationIssue> executeRuleOnArtifacts
+            (final CompareArtifactRule rule, final Collection<ThirdPartyArtifact> thirdPartyArtifacts) {
+        return thirdPartyArtifacts.stream()
+                .map(artifact -> findViolationsForLeftHandSide(rule, artifact, thirdPartyArtifacts))
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
+
+    private Collection<ViolationIssue> findViolationsForLeftHandSide
+            (final CompareArtifactRule rule, final ThirdPartyArtifact leftHandSide, final Collection<ThirdPartyArtifact> rightHandSides) {
```
maxhbr: our usual style is no newline before a `(`

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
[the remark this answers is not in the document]
maxhbr: I had a hard time finding the source for the `IllegalArgumentException`

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
blaumeiser-at-bosch: If you are working in a functional fashion, i.e., objects are typically immutable. The place where this is important is the creation and there you have all the flexibility to switch between both types. In most lists I worked with in the past, duplicates where not the main concern, but the sequence matters. But this is also an aspect of creation time, so having a List element is a slight indicator for a collection user, but actually no impact on programming.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
blaumeiser-at-bosch: I'll check, that is a valid point.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyException.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * Policy {@link Rule} execution in a {@link RuleExecutor} ends with this exception if a policy violation is found.
+ */
+public class PolicyException extends Exception {
+    public PolicyException() {
```
maxhbr: All functions throwing `PolicyException` have `void` return type, and thus could encode the result in the return value instead.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

blaumeiser-at-bosch:
@sschuberth @maxhbr I should have covered now all your remarks. Please check and review again.
Thanks!

blaumeiser-at-bosch:
Azure green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=411

blaumeiser-at-bosch:
Squashed the smaller commits and separate the work into three major commits.

blaumeiser-at-bosch:
Latest build in Azure (green): https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=424

blaumeiser-at-bosch:
Azure green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=427

blaumeiser-at-bosch:
@sschuberth @maxhbr: I would like to continue on this on Friday, could you review it until then? There are some substantial changes compared to last time you looked at it.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
```diff
@@ -0,0 +1,42 @@
...
+public class PolicyEngine {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PolicyEngine.class);
+
+    private final Collection<RuleExecutor> executors;
+
+    public PolicyEngine(final Collection<RuleExecutor> executors) {
+        this.executors = executors;
+    }
+
+    public Collection<ViolationIssue> evaluate (final Collection<ThirdPartyArtifact> thirdPartyArtifacts) {
```
sschuberth: Nit: Superfluous space before `(`.

modules/policy/engine/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-pmd-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
sschuberth: Nit: Missing final newline.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
sschuberth: This actually returns a set. Shouldn't we reflect this in the return type? In particular, to point out that `ViolationIssues` are de-duplicated?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Policy engine that executes all rules on the given data by calling the different {@link RuleExecutor} instances.
+ */
+public class PolicyEngine {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PolicyEngine.class);
+
+    private final Collection<RuleExecutor> executors;
```
sschuberth: If we just use `Collection` here, the iteration order is not necessarily the insertion order. I wonder whether we *do* want to give the caller control over the iteration / execution order, e.g. by using a `List` here. I'm aware you're anyway using `parallelStream()` below, so order of execution is non-deterministic currently anyway, so maybe `evaluate()` could get a boolean argument to toggle parallelism or not.

Basically, this boils down to a design decision, and I'd be fine with never guaranteeing execution order, but I feel this should be clearly documented. In any case I'm missing documentation in this class.

modules/policy/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+
+    <artifactId>policy-module</artifactId>
+    <name>policy-module</name>
+
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>engine</module>
+    </modules>
+</project>
```
sschuberth: Nit: Missing final newline.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/antenna/workflow/processors/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public Set<Artifact> getFailedArtifacts() {
+        return violation.getFailedArtifacts()
+                         .stream()
+                         .map(element -> ((AntennaArtifact)element).getArtifact())
+                         .collect(Collectors.toSet());
+    }
+}
```
sschuberth: Nit: Missing final newline.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSet.java
```diff
@@ -0,0 +1,22 @@
...
+package org.eclipse.sw360.antenna.policy.engine;
+
+import java.util.Collection;
+
+/**
+ * This is a collection of {@link Rule} objects within a common context. The configuration of the {@link PolicyEngine}
+ * in the {@link PolicyEngineConfigurator} allows to give several rule sets to set up the policies.
+ */
+public interface RuleSet {
+    String version();
```
sschuberth: Why is the version a free-text string that makes it hard to compare what is newer vs. older? How about making this a monotonically increasing `int`, similar to Subversion revisions, for simple version management?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
```diff
@@ -28,21 +28,21 @@ public PolicyEngine(final Collection<RuleExecutor> executors) {
...
         LOGGER.info("Start policy engine run");
         LOGGER.debug("Artifacts are" + thirdPartyArtifacts.toString());
 
         return executors
                 .parallelStream()
                 .map(executor -> executor.executeRules(thirdPartyArtifacts))
-                .flatMap(results -> results.stream())
-                .collect(Collectors.toSet());
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
```
sschuberth: Note: This is a subtle change to before, because violations are not de-duplicated here anymore.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @return A hash code that uniquely identifies the violation. It is reproducable over different rule evaluations
+     */
+    public String getViolationHash() {
+        String hashBase = failedArtifacts.stream()
+                .map(ThirdPartyArtifact::getPurl)
+                .collect(Collectors.joining(" : ", rule.getId() + " : ", ""));
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance("MD5");
```
sschuberth: Why not just use `hashBase.hashCode()`?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
```diff
@@ -25,11 +24,11 @@
  */
 public class PolicyViolation {
     private Rule rule;
-    private Collection<ThirdPartyArtifact> failedArtifacts = new HashSet<>();
+    private Collection<ThirdPartyArtifact> failingArtifacts = new HashSet<>();
```
sschuberth: Single renamings like this should be separated out to another commit.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public void configure(final Map<String, String> configMap) throws AntennaConfigurationException {
+        super.configure(configMap);
+
+        LOGGER.info("Setting up policy engine");
+
+        final String ruleSetClasses = getConfigValue(RULESETPROP, configMap).trim();
+        final Collection<String> ruleSetClassesList = Arrays.asList(ruleSetClasses.split(","))
```
maxhbr:
```suggestion
        final Collection<String> ruleSetClassesList = Arrays.stream(ruleSetClasses.split(","))
             .map(String::trim)
```

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public String getRulesetDescription() {
+        StringBuilder resultString = new StringBuilder("Policy Engine, with rule sets:\n");
+        policyEngine.getRuleSets().forEach(rs -> resultString.append(prepareRuleSetInfo(rs)));
+        return resultString.toString();
+    }
+
+    private String prepareRuleSetInfo(Ruleset ruleSet) {
```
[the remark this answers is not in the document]
maxhbr:
```suggestion
    private String prepareRulesetInfo(Ruleset ruleset) {
```

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/testdata/EPLvsGPLRule.java
```diff
@@ -0,0 +1,60 @@
...
+    private final Ruleset ruleSet;
+
+    public EPLvsGPLRule(final Ruleset ruleSet) {
+        this.ruleSet = ruleSet;
+    }
+
+    @Override
+    public Optional<PolicyViolation> evaluate(final ThirdPartyArtifact leftArtifact,
+            final ThirdPartyArtifact rightArtifact) {
+        if(leftArtifact.hasLicense("EPL.*") && rightArtifact.hasLicense("GPL.*")) {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: This is an example rule, which does not completely express the fact, but I think that the hasLicense method needs rework, the regular expression is not really nice to read and can be misinterpreted.

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/testdata/AntennaTestdata.java
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.sw360.antenna.model.artifact.Artifact;
+import org.eclipse.sw360.antenna.model.artifact.facts.DeclaredLicenseInformation;
+import org.eclipse.sw360.antenna.model.xml.generated.MatchState;
+import org.eclipse.sw360.antenna.util.LicenseSupport;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+public class AntennaTestdata {
+    public static final String UNKNOWNARTIFACTID = "UA";
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Done for all testdata constants

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+        return String.format(" - Rule set %s in version %s%n", ruleSet.getName(), ruleSet.getVersion());
+    }
+}
+
+/**
+ * Implementation for the Antenna {@link IPolicyEvaluation} result based on the {@link PolicyEngine} data
+ */
+class PolicyEvaluation implements IPolicyEvaluation {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+        this.workflowStepOrder = VALIDATOR_BASE_ORDER + 11000;
+    }
+
+    @Override
+    public void configure(final Map<String, String> configMap) throws AntennaConfigurationException {
+        super.configure(configMap);
+
+        LOGGER.info("Setting up policy engine");
+
+        final String ruleSetClasses = getConfigValue(RULESETPROP, configMap).trim();
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/PolicyEngineProcessor.java
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/AntennaArtifact.java
```diff
@@ -0,0 +1,57 @@
...
+        return ArtifactLicenseUtils.getFinalLicenses(artifact)
+                .getLicenses()
+                .stream()
+                .map(License::getName)
+                .anyMatch(license -> license.matches(licenseRegex));
+    }
+
+    @Override
+    public Optional<PackageURL> getPurl() {
+        return ArtifactUtils.getMostDominantArtifactCoordinates(Collections.EMPTY_LIST, artifact);
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: @maxhbr: How can fill the preferedCoordinatesTypes list here, how is this meant?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/SingleArtifactExecutor.java
```diff
@@ -0,0 +1,49 @@
...
+            final Collection<ThirdPartyArtifact> thirdPartyArtifacts) {
+        return thirdPartyArtifacts.stream()
+                .map(artifact -> rule.evaluate(artifact))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public Collection<Ruleset> getRuleSets() {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleExecutor.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Interface that implements an executor that goes through all {@link Rule} instances of the corresponding type and
+ * applies the rule on the available data. The design pattern is, that every type of rules has a corresponding executor.
+ * The rule is reduced to a function that checks a condition. If the condition is false, the rule returns a
+ * {@link PolicyViolation} that allows the executor to identify failing artifacts.
+ */
+interface RuleExecutor {
+    Collection<PolicyViolation> executeRules(Collection<ThirdPartyArtifact> thirdPartyArtifacts);
+
+    Collection<Ruleset> getRuleSets();
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/workflow/processors/AntennaArtifact.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Clarified via phone, done

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/Rule.java
```diff
@@ -0,0 +1,41 @@
...
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.policy.engine;
+
+/**
+ * Interface for a generic rule with the standard fields relevant for all policy rules.
+ */
+public interface Rule {
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: A Ruleset has a set of rules and allows to define a version. That is to fulfil the OSO requirement to have versioned rules. But since there might be rules of different sources I entered the Ruleset to allow different sets, e.g., the open Antenna rules, the basic OSO rules, customer specific rules. All can be defined by different rulesets.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
blaumeiser-at-bosch: hashCode provides only a single int as hash. My impression is, that the likelyhood of collisions is there. The idea of the hash code is, to have a unique id for the violation, to, e.g., use them later in a waiving mechanism. So collisions should be avoided as much as possible.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
blaumeiser-at-bosch: This is really an interesting strictness. I once learned that it is a good practice to clean up a piece of code you are working on in a change in order to improve the codebase step by step. Should be part of our discussion, my opinion here is, that if it is only improving understandability or making things more consistent is not a special commit, if the code is refactored structurally, this is of course a new commit.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngineConfigurator.java
```diff
@@ -0,0 +1,122 @@
...
+    private static Collection<RuleExecutor> createExecutors(final Map<String, Set<Rule>> executorToRuleMapping) {
+        return executorToRuleMapping.entrySet()
+                .stream()
+                .map(entry -> createExecutor(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList());
+    }
+
+    private static RuleExecutor createExecutor(String executorClass, Collection<Rule> rules) {
+        if (SINGLE_ARTIFACT_EXECUTOR.equals(executorClass)) {
+            return new SingleArtifactExecutor(ruleToSpecialArtifactRule(rules));
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Chaned it a bit, but I prefer the readbility of this instead of type magic, to be honest.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
blaumeiser-at-bosch: That is right, but I thought about it and I concluded that there are no duplications because every rule is executed on artifacts once, so two violations should always have a different context.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
sschuberth: I guess this is related to the discussion whether change == commit or change == PR. Of course, if you come across a sensible renaming while working on something else, you should be free to do that in the same PR, but I'd still vote for doing it in a separate commit, so that functional and non-functional changes are separated. But I agree to discuss that separately.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
sschuberth: Taking about collisions, how about using SHA1 instead then, also as that's more common in general nowadays? Or do you prefer MD5 as the string is shorter?

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final Collection<RuleExecutor> executors;
+
+    public PolicyEngine(final Collection<RuleExecutor> executors) {
+        this.executors = executors;
+    }
+
+    public Collection<PolicyViolation> evaluate (final Collection<ThirdPartyArtifact> thirdPartyArtifacts) {
+        LOGGER.info("Start policy engine run");
+        LOGGER.debug("Artifacts are" + thirdPartyArtifacts.toString());
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done, use purls now

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
maxhbr: I think MD5 is good enough and we will never find a collision

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyEngine.java
blaumeiser-at-bosch: There is no sequence in the execution of executors. So the collection is the right data structures. All rules are executed in parallel and are fed with all relevant combinations of artifacts. I do not see a logical sequence in Rule execution as of now as a concept.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/CompareArtifactExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+        return rightHandSides.stream()
+                .map(rightHandSide -> rule.evaluate(leftHandSide, rightHandSide))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public Collection<Ruleset> getRuleSets() {
+        return rules.stream().map(Rule::getRuleset).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: done

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/testdata/EPLvsGPLRule.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Simplified example, using EPL-1.0 and GPL-2.0-or-later

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/PolicyViolation.java
blaumeiser-at-bosch: I used MD5 because the string is shorter, indeed. I compared the results and since this should be in principle usable to refer to the violation, I thought that MD5 is the best compromise.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/ThirdPartyArtifact.java
```diff
@@ -0,0 +1,40 @@
...
+     *         and needs an identification process to clarify the component source
+     */
+    boolean isIdentified();
+
+    /**
+     * @param licenseRegex A regular expression representing a family of licenses in the form of a SPDX license reference,
+     *                     e.g., "GPL.*" for all GPL strings like GPL-2.0-only
+     * @return True, if a license that matches the pattern is found, False, if the license is not found
+     */
+    boolean hasLicense(String licenseRegex);
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: Changed the method to use a collection of license strings instead of a regex. The regex is difficult to understand and hard to express. But in general these methods are are of potential change, since the idea is to make the rules easy to understand.

blaumeiser-at-bosch:
@sschuberth @maxhbr Should have resolved all issues or at least explained the solution.

blaumeiser-at-bosch:
Azure green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=448

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/testdata/EPLvsGPLRule.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public EPLvsGPLRule(final Ruleset ruleset) {
+        this.ruleset = ruleset;
+    }
+
+    @Override
+    public Optional<PolicyViolation> evaluate(final ThirdPartyArtifact leftArtifact,
+            final ThirdPartyArtifact rightArtifact) {
+        if (leftArtifact.hasAtLeastOneLicenseOf(Arrays.asList("EPL-1.0"))
+                && rightArtifact.hasAtLeastOneLicenseOf(Arrays.asList("GPL-2.0-or-later"))) {
```
sschuberth: Just a though: Is there a risk that the deprecated id for this license, i.e. `GPL-2.0+`, is being used? If so, it should probably be listed here, too.

sschuberth:
> @sschuberth @maxhbr Should have resolved all issues or at least explained the solution.

Out of curiosity, did you provide a rationale why to prefer a string version of an integer one? At least I didn't find any. If you still don't want to use ints, maybe using [semver4j](https://github.com/vdurmont/semver4j) would be an option (which is that we are using in ORT). But that could probably be changed later, in a follow-up PR.

modules/policy/engine/src/test/java/org/eclipse/sw360/antenna/policy/workflow/processors/testdata/EPLvsGPLRule.java
blaumeiser-at-bosch: This is a test version, so I only expect what I give in the testdata :-). But in general, I thought about this when implementing the rule. My conclusion was, that this should not be the case, since the data in entered by the OSO office, so we have to teach Sheriff, which identifiers to use. In addition, I think of a rule, that checks for proper usage of SPDX identifiers, resp. gives warnings on unknown identifiers

blaumeiser-at-bosch:
> > @sschuberth @maxhbr Should have resolved all issues or at least explained the solution.
> 
> Out of curiosity, did you provide a rationale why to prefer a string version of an integer one? At least I didn't find any. If you still don't want to use ints, maybe using [semver4j](https://github.com/vdurmont/semver4j) would be an option (which is that we are using in ORT). But that could probably be changed later, in a follow-up PR.

What do you mean by string version? You mean for Rulesets?

sschuberth:
> What do you mean by string version? You mean for Rulesets?

Yes, I was referring to [this](https://github.com/eclipse/antenna/pull/260#discussion_r318426255) (unanswered) comment of mine.

modules/policy/engine/src/main/java/org/eclipse/sw360/antenna/policy/engine/RuleSet.java
blaumeiser-at-bosch: I'll discuss that with Marcel, seems that I overlooked this comment of yours, sorry. Perhaps also your proposal makes sense. But I go with your comment, that we can change it later, at least when we  implement the first "official" Ruleset

PR #264, opened 2019-08-22
arch(maven): reduce context usage to maven resolver

neubs-bsi:
refactored maven artifact resolver to have an impl class that is not dependent on the AntennaContext. 
Consequently I refactored all classes used in the `process` function of the workflow step.


neubs-bsi:
@maxhbr I changed the necessity of proxySettings in the IArtifactRequestFactory. hence a re-review please. thanks

PR #266, opened 2019-08-22
test(ci): add licenseHeader check to azure

maxhbr:


neubs-bsi:
@maxhbr is this still WIP? it's green in azure

PR #267, opened 2019-08-26
Add purl getter to ArtifactCoordinates

blaumeiser-at-bosch:
Add the support for purls : https://github.com/package-url/purl-spec to artifacts. This is done as getter in the ArtifactCoordinates fact.

Will be tested within the policy engine, as soon as this PR is through

### DIFF
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
@@ -11,9 +11,35 @@
 
 package org.eclipse.sw360.antenna.model.artifact.facts;
 
+import com.github.packageurl.MalformedPackageURLException;
+import com.github.packageurl.PackageURL;
+import com.github.packageurl.PackageURLBuilder;
+
 public abstract class ArtifactCoordinates<T extends ArtifactCoordinates> implements ArtifactIdentifier<T> {
     public abstract String getName();
+
     public abstract String getVersion();
+
+    public abstract String getType();
+
+    protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
+        return builder;
+    }
+
+    public PackageURL getPurl() {
+        PackageURLBuilder builder =  PackageURLBuilder.aPackageURL()
+                .withType(getType())
+                .withName(getName())
+                .withVersion(getVersion());
+
+        builder = addPurlFacts(builder);
+        
+        try {
+            return builder.build();
+        } catch (MalformedPackageURLException e) {
+            throw new IllegalArgumentException("Could not build Purl", e);
+        }
+    }
 
     @Override
     public String prettyPrint() {
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
@@ -37,6 +37,11 @@ public class DotNetCoordinates extends ArtifactCoordinates<DotNetCoordinates> {
     @Override
     public String getVersion() { return version; }
 
+    @Override
+    public String getType() {
+        return "nuget";
+    }
+
     public String getPackageId() { return packageId; }
 
     @Override
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
@@ -63,6 +63,11 @@ public class BundleCoordinates extends JavaCoordinates<BundleCoordinates> {
     }
 
     @Override
+    public String getType() {
+        return "p2";
+    }
+
+    @Override
     public boolean matches(ArtifactIdentifier artifactIdentifier) {
         if(artifactIdentifier instanceof BundleCoordinates) {
             final BundleCoordinates bundleCoordinates = (BundleCoordinates) artifactIdentifier;
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
@@ -11,6 +11,7 @@
 
 package org.eclipse.sw360.antenna.model.artifact.facts.java;
 
+import com.github.packageurl.PackageURLBuilder;
 import org.eclipse.sw360.antenna.model.artifact.ArtifactFactBuilder;
 import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactIdentifier;
 
@@ -59,6 +60,16 @@ public class MavenCoordinates extends JavaCoordinates<MavenCoordinates> {
     @Override
     public String getVersion() {
         return version;
+    }
+
+    @Override
+    public String getType() {
+        return "maven";
+    }
+
+    @Override
+    protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
+        return builder.withNamespace(getGroupId());
     }
 
     @Override
--- a/core/frontend-stubs/maven-frontend-stub/pom.xml
+++ b/core/frontend-stubs/maven-frontend-stub/pom.xml
@@ -187,6 +187,10 @@
             <artifactId>spring-hateoas</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.github.package-url</groupId>
+            <artifactId>packageurl-java</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.github.spotbugs</groupId>
             <artifactId>spotbugs-annotations</artifactId>
             <scope>compile</scope>
--- a/core/model/pom.xml
+++ b/core/model/pom.xml
@@ -46,6 +46,10 @@
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-plugin-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.package-url</groupId>
+            <artifactId>packageurl-java</artifactId>
+        </dependency>
         <!-- ################################ testing dependencies ################################ -->
         <dependency>
             <groupId>junit</groupId>
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/GenericArtifactCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/GenericArtifactCoordinates.java
@@ -38,6 +38,11 @@ public class GenericArtifactCoordinates extends ArtifactCoordinates {
     }
 
     @Override
+    public String getType() {
+        return "generic";
+    }
+
+    @Override
     public boolean isEmpty() {
         return (name == null || "".equals(name)) &&
                 (version == null || "".equals(version));
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
@@ -63,6 +63,11 @@ public class JavaScriptCoordinates extends ArtifactCoordinates<JavaScriptCoordin
     }
 
     @Override
+    public String getType() {
+        return "npm";
+    }
+
+    @Override
     public String toString() {
         return name + ":" + version + " (" + artifactId + ")";
     }
--- a/pom.xml
+++ b/pom.xml
@@ -262,6 +262,11 @@
                 <version>0.25.0.RELEASE</version>
             </dependency>
             <dependency>
+                <groupId>com.github.package-url</groupId>
+                <artifactId>packageurl-java</artifactId>
+                <version>1.1.0</version>
+            </dependency>
+            <dependency>
                 <groupId>com.github.spotbugs</groupId>
                 <artifactId>spotbugs-annotations</artifactId>
                 <version>3.1.12</version>


core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
```diff
@@ -37,6 +37,11 @@ public DotNetCoordinates(String packageId, String version) {
     @Override
     public String getVersion() { return version; }
 
+    @Override
+    public String getPurl() {
+        return String.format("pkg:nuget/%s@%s", getName(), getVersion());
```
sschuberth: In general, purl-components need to be percent-encoded, see

https://github.com/heremaps/oss-review-toolkit/blob/673c9b817cb71fed1bcd313d51d4b7788fbd2e61/model/src/main/kotlin/Identifier.kt#L126
https://github.com/package-url/purl-spec#rules-for-each-purl-component

modules/drools/drools-checker/src/main/java/org/eclipse/sw360/antenna/drools/Artifact.java
```diff
@@ -0,0 +1,4 @@
+package org.eclipse.sw360.antenna.drools;
```
maxhbr: missing license headers

modules/drools/drools-checker/src/main/java/org/eclipse/sw360/antenna/drools/PolicyEngine.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.sw360.antenna.drools;
```
maxhbr: missing license headers

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
```diff
@@ -62,6 +62,11 @@ public String getVersion() {
         return getBundleVersion();
     }
 
+    @Override
+    public String getPurl() {
+        return String.format("pkg:p2/%s@%s", getName(), getVersion());
```
sschuberth: Instead of having this per coordinates implementation, can't we implement his in the base class and instead have the coordinates classes only implement a `getType()` method that is called by the base class implementation? At the example of this class, `getType()` should return "p2".

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
sschuberth: Also, all implementations except the Maven one seem to miss the "namespace" part.

modules/drools/drools-checker/src/main/java/org/eclipse/sw360/antenna/drools/Artifact.java
maxhbr: are these changes related to this PR?

modules/drools/drools-checker/src/main/java/org/eclipse/sw360/antenna/drools/PolicyEngine.java
maxhbr: are these changes related to this PR?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
maxhbr: just `getType` might not be enough: e.g. Docker has multiple ways of building a purl (with tag, with hash, ...)

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -14,6 +14,7 @@
 public abstract class ArtifactCoordinates<T extends ArtifactCoordinates> implements ArtifactIdentifier<T> {
     public abstract String getName();
     public abstract String getVersion();
+    public abstract String getPurl();
```
maxhbr: How do we plan to handle coordinates with insufficient information (e.g. with missing version or missing groupId)?

maxhbr: How do we plan to handle coordinates types, which are not covered by the current purl definition? Throwing `UnsupportedOperationException`? Maybe add this to the interface?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
sschuberth: Right, still the `type` would be "docker" in all cases. `qualifiers` and `subpath` components could be added later without breaking the existing code.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
sschuberth: > How do we plan to handle coordinates types, which are not covered by the current purl definition?

Do you have any concrete example for such a type? Ideally we'd fix this by contributing to [purl-spec](https://github.com/package-url/purl-spec). But as an interim solution, yes, probably throw an exception.

maxhbr: I think for example stackage is not supported. (Unimportant: I am even unsure how something like stackage would be supported, where the version is attached to stackage itself and references consistent package sets, instead of individual versions for every single component.)

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -11,10 +11,29 @@
...
+    public abstract String getType();
+
+    public String getPurl() {
+        return String.format("pkg:%s/%s@%s", percentEncoding(getType()), percentEncoding(getName()),
+                percentEncoding(getVersion()));
+    }
+
+    protected String percentEncoding(String input) {
+        try {
+            return URLEncoder.encode(input, "UTF-8");
```
sschuberth: Just `encode` is unfortunately not enough, encoding "+" needs to be handled explicitly, see https://github.com/heremaps/oss-review-toolkit/blob/673c9b817cb71fed1bcd313d51d4b7788fbd2e61/utils/src/main/kotlin/Extensions.kt#L260-L263

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
blaumeiser-at-bosch: I do not think that missing information is an issue. In this case the component information is not sufficient, i.e., the purl is not valid and cannot be used. In this case, the artifact is not identified, so the compliance information has a major issue and therefore is only preliminary. The user has to solve that anyway.

The first usage of the purl is in the policy engine to make a unique hash of a policy violation, so that we can identify violations over different evaluations. I would prefer that the method always returns a String, even if the data is not complete. An unidentified artifact in the policy engine is a blocking violation, so the uniqueness of the violation issue is not that important, since it has to be solved.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
blaumeiser-at-bosch: I added this in my latest update. Concerning the namespace part, I think this is a minor bug in ORT, if I compare the examples from https://github.com/package-url/purl-spec:

pkg:npm/%40angular/animation@12.3.1
pkg:npm/foobar@12.3.1

pkg:nuget/EnterpriseLibrary.Common@6.0.1304

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -11,10 +11,29 @@
...
+
 public abstract class ArtifactCoordinates<T extends ArtifactCoordinates> implements ArtifactIdentifier<T> {
     public abstract String getName();
+
     public abstract String getVersion();
 
+    public abstract String getType();
+
+    public String getPurl() {
+        return String.format("pkg:%s/%s@%s", percentEncoding(getType()), percentEncoding(getName()),
```
sschuberth: The "namespace" component is still missing from here. In general a purl looks like: `pkg:<type>/<namespace>/<name>/<version>?<qualifiers>#<subpath>`, but the `?<qualifiers>#<subpath>` could be omitted for now.

If a package manager does not support a namespace, an empty string should be used, resulting in two consecutive `/` in the purl.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/BundleCoordinates.java
blaumeiser-at-bosch: I changed that as you @sschuberth proposed. All coordinates we have now, are sufficiently encoded with the current functionality. So if we face more difficult Purl cases, the idea is to overwrite the getPurl method. Nonetheless, I like the idea that you can get a repository type from the coordinates.

blaumeiser-at-bosch:
@sschuberth @maxhbr Changes are made as requested, resp. I commented on some remarks

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
```diff
@@ -61,6 +61,17 @@ public String getVersion() {
         return version;
     }
 
+    @Override
+    public String getType() {
+        return "maven";
+    }
+
+    @Override
+    public String getPurl() {
```
sschuberth: This specialization is not necessary with the generalization of namespaces, see above.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/dotnet/DotNetCoordinates.java
sschuberth: Oh, you're right, "The name is prefixed by a '/' separator when the namespace is not empty"!

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
maxhbr: but in the purl spec there are also examples like `pkg:gem/jruby-launcher@1.1.2?platform=java`

maxhbr: The purl spec even explicitly says, that this segment is not allowed to be empty:

> - **namespace**:
>
>  - The optional `namespace` contains zero or more segments, separated by slash   '/'
>  - Leading and trailing slashes '/' are not significant and should be stripped    in the canonical form. They are not part of the `namespace`
>  - Each `namespace` segment must be a percent-encoded string
>  - When percent-decoded, a segment:
>
>    - must not contain a '/'
>    - **must not be empty**
>
>  - A URL host or Authority must NOT be used as a `namespace`. Use instead a    `repository_url` qualifier. Note however that for some types, the    `namespace` may look like a host.

blaumeiser-at-bosch: See discussion above, that is correct

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
blaumeiser-at-bosch: I think, here we are in a more general discussion. If I add a getNamespace() method to ArtifactCoordinates, the whole concept of Artifact Coordinates could be simplified to one class as you did in ORT. But I did not want to open this box of Pandora for now. Making this overwrite is simplifying the world for all other coordinate types we have so far.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
sschuberth: How was this resolved? Did you forget to push?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -11,10 +11,36 @@
...
 public abstract class ArtifactCoordinates<T extends ArtifactCoordinates> implements ArtifactIdentifier<T> {
     public abstract String getName();
+
     public abstract String getVersion();
 
+    public abstract String getType();
+
+    public String getPurl() {
+        return String.format("pkg:%s/%s%s", percentEncoding(getType()), percentEncoding(getName()),
+                percentEncoding(getVersionForPurl()));
```
sschuberth: This also encodes the "@" separator for the version. I suggest to move `percentEncoding()` to inside `getVersionForPurl()` at line 31 instead.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -11,10 +11,36 @@
...
+
 public abstract class ArtifactCoordinates<T extends ArtifactCoordinates> implements ArtifactIdentifier<T> {
     public abstract String getName();
+
     public abstract String getVersion();
 
+    public abstract String getType();
+
+    public String getPurl() {
+        return String.format("pkg:%s/%s%s", percentEncoding(getType()), percentEncoding(getName()),
```
sschuberth: The `type` must not be percent-encoded, see https://github.com/package-url/purl-spec#rules-for-each-purl-component.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
```diff
@@ -61,6 +61,17 @@ public String getVersion() {
...
+        return String.format("pkg:%s/%s/%s%s", percentEncoding(getType()), percentEncoding(getGroupId()),
```
sschuberth: The `type` must not be percent-encoded, see https://github.com/package-url/purl-spec#rules-for-each-purl-component.

sschuberth:
Just FYI, there's also https://github.com/package-url/packageurl-java which could be taken into use. We're not doing so for ORT yet either, but I'm considering it.

maxhbr:
Alternatively to the comment https://github.com/eclipse/antenna/pull/267#issuecomment-524819103 by @sschuberth, I would suggest to encapsulate the purl stuff in something like:
```java
package org.eclipse.sw360.antenna.model.util;

import java.io.UnsupportedEncodingException;
import java.net.URLEncoder;
import java.util.List;
import java.util.Optional;

public class PurlUtils {
    static private String percentEncoding(String input) {
        try {
            return URLEncoder.encode(input, "UTF-8").replace("+", "%20");
        } catch (UnsupportedEncodingException e) {
            throw new IllegalStateException("Programming error: Unknown encoding", e);
        }
    }

    static public String buildPurl(String type, List<String> path, Optional<String> version) {
        final StringBuilder stringBuilder = new StringBuilder("pkg:");
        stringBuilder.append(type);
        path.stream()
                .map(PurlUtils::percentEncoding)
                .forEach(s -> {
                        stringBuilder.append("/");
                        stringBuilder.append(s);
                });
        version.map(PurlUtils::percentEncoding)
                .ifPresent(v -> {
                stringBuilder.append("@");
                stringBuilder.append(v);
        });
        return stringBuilder.toString();
    }
}
```

blaumeiser-at-bosch:
I will switch to the PURL library, that seems to be closer to the spec source.

sschuberth:
Can we please squash the commits in this PR into a single commit now? Otherwise reviewing commit by commit is really misleading now, as the last commit voids it all.

blaumeiser-at-bosch:
> Can we please squash the commits in this PR into a single commit now? Otherwise reviewing commit by commit is really misleading now, as the last commit voids it all.

@sschuberth: Done

blaumeiser-at-bosch:
Azure is green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=414

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
```diff
@@ -70,7 +81,7 @@ public boolean isEmpty() {
 
     @Override
     public boolean matches(ArtifactIdentifier artifactIdentifier) {
-        if(artifactIdentifier instanceof MavenCoordinates) {
+        if (artifactIdentifier instanceof MavenCoordinates) {
```
sschuberth: Nit: Strictly speaking, the unrelated whitespace changes here and below should be moved out to a separate commit.

blaumeiser-at-bosch: Unfortunately a result of reformatting the file after my changes above this. Really removing that from the file?

sschuberth: Well, I leave it to how strictly you handle this (hence only a "nit"). In ORT, we don't allow to mix topics within a single commit, and that includes formatting, as we try to separate functional from non-functional changes for more speedy reviews.

sschuberth: FWIW, I didn't mean to *drop* the change from this PR, but just to separate it out into another commit in *this* PR.

blaumeiser-at-bosch: These changes were not intended at all, they were result of auto formatting, so I discarded them. In general, perhaps we should define at some point a common formatting and apply this over time to the source base. But this is not part of this change :-)

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
```diff
@@ -11,10 +11,36 @@
...
+    protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
+        return builder;
+    }
+
+    public PackageURL getPurl() {
+        PackageURLBuilder builder =  PackageURLBuilder.aPackageURL()
+                .withType(getType())
+                .withName(getName());
+
+        if (getVersion() != null) {
```
sschuberth: This explicit `null` check is still there, did you intentionally keep it?

blaumeiser-at-bosch: Version is according to the purl spec an optional value. I thought that it makes sense to not set the value, if the version is unknown instead of setting a null value, so yes this is intentional.

blaumeiser-at-bosch:
Latest Azure build green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=420

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactCoordinates.java
sschuberth: My point is: Exactly that logic is already handled by the purl library (see the code I had linked in my original comment), so why also handle that case here?

blaumeiser-at-bosch: Ok accepted, actually, there was a major bug in the implementation, but now it is fixed

blaumeiser-at-bosch:
Azure is green: https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=425, therefore merging, although Travis not run.

PR #270, opened 2019-08-29
Fix mapping of ORT artifacts URLs

sschuberth:

### DIFF
--- a/modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
+++ b/modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
@@ -98,10 +98,10 @@ public class OrtResultArtifactResolver implements Function<Package, Artifact> {
     }
 
     private static Optional<ArtifactSourceUrl> mapSourceUrl(Package pkg) {
-        final Stream<String> urls = Stream.of(pkg.getVcsProcessed().getUrl(), pkg.getSourceArtifact().getUrl());
-
-        return urls.filter(u -> !u.isEmpty())
-                .findFirst()
+        // Antenna does not currently have the concept of VCS-specific clone URLs, so do not take ORT's VCS URL into
+        // account, but only the source artifact URL.
+        return Optional.of(pkg.getSourceArtifact().getUrl())
+                .filter(a -> !a.isEmpty())
                 .map(ArtifactSourceUrl::new);
     }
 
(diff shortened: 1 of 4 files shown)


modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -98,6 +98,9 @@ private static ArtifactCoordinates mapSimpleCoordinates(String name, String vers
     }
 
     private static Optional<ArtifactSourceUrl> mapSourceUrl(Package pkg) {
+        // Prefer a VCS URL for the source URL as that usually is the pristine source from which also source artifacts
+        // are created, and only if empty fall back to a source artifact URL (which might have been processed e.g. by
+        // stripping / minimizing license headers).
         final Stream<String> urls = Stream.of(pkg.getVcsProcessed().getUrl(), pkg.getSourceArtifact().getUrl());
```
maxhbr: Is the vcs url an url to a `.tar.gz` or to the github page? The second can not automatically be processed by our code and maybe even result in a html page added as source file.

sschuberth: It's neither. In ORT the VCS URL is the VCS-specific URL to clone the project using a VCS tool.

maxhbr: Our source URL should be a HTTP URL to a downloadable artifact. We have a `SourceUrlResolver` which downloads the URL if no sources are already present.

For a VCS URL we should add a new type of Fact.

sschuberth: Good to know! Does that mean in turn that is does not make sense to consider ORT's source artifact URL as part of `mapFilename()` as it's already considered here?

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -112,17 +112,11 @@ private static ArtifactCoordinates mapSimpleCoordinates(String name, String vers
...
-                .findFirst()
-                .map(a -> {
-                    String fileName = a.getUrl();
-                    String hash = a.getHash().getValue();
-                    String hashAlgorithm = a.getHash().getAlgorithm().toString();
-                    return new ArtifactFilename(fileName, hash, hashAlgorithm);
-                });
+        // Antenna's artifact filename (or filename entries) are meant to refer to binary artifacts, so only use the
+        // filename from ORT's binary artifact URL here.
+        return Optional.of(new File(pkg.getBinaryArtifact().getUrl()).getName())
```
neubs-bsi: doesn't the binary artifact also have a hash? 
In that case you could also give this to the filename

sschuberth: You're right, I dropped too much of the original code! Thanks for pointing that out, fixed.

PR #272, opened 2019-08-29
doc(sw360): update with new configuration entries

maxhbr:


neubs-bsi:
@maxhbr error parsing workflow in `AntennaCLIFrontendTest`, seems you forgot to adapt your changes from the example project there

PR #273, opened 2019-08-29
fix(sw360): internally created release could never match fetched release

neubs-bsi:


neubs-bsi:
waiting for #274

PR #277, opened 2019-08-30
Rewrite REST Serialization for SW360Releases and drop redundant information

maxhbr:


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -44,16 +45,15 @@ public static void prepareRelease(SW360Release release, SW360Component component
...
         SW360ReleaseAdapterUtils.setCopyrights(release, artifact);
+
+        return release;
     }
 
     public static boolean isValidRelease(SW360Release release) {
         if(release.getName() == null || release.getName().isEmpty()) {
             return false;
         }
-        if(release.getVersion() == null || release.getVersion().isEmpty()) {
```
neubs-bsi: Just a question: Why not name? Does a release not need a name to be valid?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -104,184 +122,144 @@ public void setDownloadurl(String downloadurl) {
...
 
-    private <T> T getDominantGetterFromVariableMergeOrNull(SW360Release release, Function<SW360Release, T> getter) {
-        return Optional.ofNullable(getter.apply(release))
-                .orElse(getter.apply(this));
+    @Override
+    public SW360ReleaseEmbedded createEmptyEmbedded() {
+        return new SW360ReleaseEmbedded();
     }
 
     @Override
```
neubs-bsi: This equal method will always be false for a release gotten from an sw360 instance vs one created via antenna, since the `super.equals(o)` will always be false, since the antenna created sw360 release `_links` and `_embedded` are always null. 

This will lead to the case, that when updating is activated the sw360 release response will never equal the internally created release.

maxhbr: I think this should already be fixed

neubs-bsi: Additionally:
ExternalIds equals also is false if the map entries are not in the same order. 
so: 
Map1: declared_license=Apache-2.0, mvn=junit:junit:4.12
Map2: mvn=junit:junit:4.12, declared_license=Apache-2.0
would not be equal.

maxhbr: the equals method on `Map` should ignore ordering. So that should not be a problem

maxhbr:
azure (https://dev.azure.com/EclipseSW360antenna/EclipseSW360antenna/_build/results?buildId=456) and travis are green

neubs-bsi:
@maxhbr What is the state of this PR?

PR #228, opened 2019-07-31
fix(sw360): move non-id fields to additionalData

maxhbr:
This modifies the sw360 adapter to work with the most recent api changes in SW360 which introduced the `additionalData` map for storing arbitrary data.

resolves https://github.com/bsinno/osm-antenna/issues/796

PR #278, opened 2019-09-03
feat(rest): use additionalData map in releases 

maxhbr:
based on: https://github.com/eclipse/antenna/pull/277

replaces: https://github.com/eclipse/antenna/pull/228


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -255,6 +257,11 @@ public SW360Release setExternalIds(Map<String,String> externalIds) {
         return this;
     }
 
+    public SW360Release setAdditionalData(Map<String,String> additionalData) {
```
bs-ondem: Rename method? This is more adding instead of setting.

PR #282, opened 2019-09-05
POC: alternative implementation for license data

maxhbr:


core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseStatement.java
```diff
@@ -0,0 +1,99 @@
...
+        if (other instanceof License) {
+            return and((License) other);
+        }
+        if (other instanceof LicenseStatement) {
+            return and((LicenseStatement) other);
+        }
+        if (other instanceof EmptyLicenseInformation) {
+            return this;
+        }
+        throw  new UnsupportedOperationException("The type of other=[" + other.getClass().getCanonicalName() + "] is unsupported");
```
sschuberth: Nit: Double whitespace.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseInformation.java
```diff
@@ -0,0 +1,17 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.license;
+
+public abstract class LicenseInformation {
```
sschuberth: I wonder whether there's a way to get rid of this `LicenseInformation` class altogether and use ORT's `SpdxExpression` (or `SpdxCompoundExpression`) class instead?

maxhbr: If `SpdxExpression` would be enriched of more information of the licenses, e.g. its full-name or its text, it would possible to use it. (We support licenses with custom text)

maxhbr: If `SpdxExpression` would be enriched with more information of the licenses, e.g. its full-name or its text, it would possible to use it. (We support licenses with custom text)

sschuberth: What would be the full name e.g. for SPDX id "Apache-2.0" in this context? The license text can be retrieved via `getText()` on any `SpdxLicense`, e.g. `SpdxLicense.forId("Apache-2.0").getText()`.

maxhbr: The fullname would be `Apache License 2.0`.

Yes, one can get **a** text of Apache-2.0 with `SpdxLicense.forId("Apache-2.0").getText()`, but that does not necessarily match the actually used formatting, spelling and copyright.

Since you have to print the actually used text, it is not enough to just use the SPDX text, which sometimes contains templating or wrong copyrights (e.g. for sleepycat I think).

We also support arbitrary shortnames, not only the SPDX subset.

sschuberth: I'd argue it's not *a* text, but *the* text as defined by the license itself. I.e. ORT's `getText()` does not (only) use [SPDX' broken plain texts](https://github.com/spdx/license-list-data/tree/master/text) (also see https://github.com/spdx/license-list-data/issues/44), but (mostly) properly formatted license text from multiple sources (similar to your https://github.com/maxhbr/LDBcollector).

But I get what you mean with "since you have to print the actually used text" anyway.

Another interesting question for me is where the full name of "Apache License 2.0" is defined. That strings sounds like a human readable name SW360 / Antenna came up with, because https://www.apache.org/licenses/LICENSE-2.0 does not list that string at all. Judging from that file, I'd more have expected something like "Apache License, Version 2.0" as the full name.

maxhbr: You are right. `Apache License, Version 2.0` is the actually correct name.

maxhbr:
closed in favor of https://github.com/eclipse/antenna/pull/314

PR #283, opened 2019-09-06
Revert changes that make Tina extensions break

blaumeiser-at-bosch:
During improvement of static code analysis, an improvement has been done, which is incompatible to tina-extensions. This reverts the changes, so that we can build Tina again.


sschuberth:
@blaumeiser-at-bosch, sorry for coming late to the party, but if this was flagged by static code analysis but does not work work for us, shouldn't this also be annotated to be ignored by static code analysis so this change is not accidentally introduced again?

maxhbr:
The better solution would be to actually add getter or setter. This would make it transparent, that these fields are used in other modules.

PR #285, opened 2019-09-10
core: Fix the casing of "setRemoveArtifact"

sschuberth:
Signed-off-by: Sebastian Schuberth <sebastian.schuberth@bosch-si.com>


maxhbr:
For the conventional changelog one would write `fix(core): Fix the casing of "setRemoveArtifact"` as commit message

sschuberth:
> For the conventional changelog one would write `fix(core): Fix the casing of "setRemoveArtifact"` as commit message

Done.

PR #289, opened 2019-09-16
doc(readme): adding logo, travis and license to readme

neubs-bsi:

Added logo to documentation resources and linked in README. 
Then added license and travis link to README.


neubs-bsi:
\cc @sschuberth

README.md
```diff
@@ -1,5 +1,10 @@
-# Antenna
+# SW360 Antenna 
 
+[![Eclipse Public License 2.0](https://img.shields.io/badge/license-EPL--2.0-green.svg "Eclipse Public License 2.0")](LICENSE)
+[![Build Status](https://travis-ci.com/eclipse/antenna.svg?branch=master)](https://travis-ci.com/eclipse/antenna)
+
+![Eclipse SW360 Antenna](antenna-documentation/src/site/resources/images/logo.png)
```
bs-ondem: What do you think about adding the logo in the same line as the heading in the first line?

neubs-bsi: I had that in the beginning, since I thought the same, then changed it, since it looks a bit weird.

sschuberth:
In the rich-diff https://github.com/eclipse/antenna/blob/neubs-bsi/doc/readme/os_links/antenna-documentation/src/site/resources/images/logo.png looks like a dead link, is that intended?

neubs-bsi:
Waiting for right logo picture

README.md
neubs-bsi: I tried it again by simply deleting the text in the header, since the logo does exactly the same. @bs-ondem do you think it is better now, or still to be improved?

bs-ondem: Imo it is better now with only the logo. :)

bs-ondem: @neubs-bsi The new logo seems to be to big. You can modify the width/height with the inline HTML tags.

PR #291, opened 2019-09-16
sw360(purl): replacing coordinates with purl in externalId

neubs-bsi:
Added purl to sw360 external Ids and removed coordinates


neubs-bsi:
@blaumeiser-at-bosch is that how you imagined it with your issue?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -72,18 +72,13 @@ private static void setCPEId(SW360Release release, Artifact artifact) {
                 .ifPresent(release::setCpeId);
     }
 
-    private static void setCoordinates(SW360Release release, Artifact artifact) {
-        release.setCoordinates(getMapOfCoordinates(artifact));
-    }
+    private static void setPackageUrl(SW360Release sw360Release, Artifact artifact) {
+        Optional<PackageURL> packageURL = artifact.askForAll(ArtifactCoordinates.class)
```
blaumeiser-at-bosch: Here, I think this findAny is not the adequate solution. In ArtifactUtils there is a default preference for coordinates which returns the coordinates according to priority, this seems to be the right thing to do.

If I remember right, we discussed that we want to store all known purls for the object. In the same way as the hash codes, i.e., with an _<number> suffix.

neubs-bsi: I had a call yesterday with @sschuberth. In our discussion we came to the conclusion that per artifact there should only be one purl. Did we get that wrong?

blaumeiser-at-bosch: This is the principle discussion whether we have one data entry for a component with different established distributions or several. E.g. a repackaged maven component, which is provided in a p2 repository has two different coordinates with the same content.

One can argue that these are two components and therefore two instances with different purls, but effectively the metadata for the component are the same, as the sources from which the component has been build are the same.

So, my preferred solution is to have all known instances of a component managed in one data object. That means multiple purls pointing to the same data object

A similar case is in nuget, where we have the original source in nuget.org, but the same  component, differently packaged from Microsoft, or with alternative package managers in other languages. In SW360, we should be able to manage multiple instances, whereas I think the way we did it in Antenna with multiple coordinates might be wrong because we look at a given set of instances, not the overall data of a component.

Adding @sschuberth into the discussion

neubs-bsi:
Waiting for #292 until continuing, since necessary for Enricher

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +130,34 @@ public void setDownloadurl(String downloadurl) {
         this.downloadurl = downloadurl;
     }
 
-
     @JsonIgnore
-    public Map<String, String> getCoordinates() {
+    public Set<String> getPurls() {
```
blaumeiser-at-bosch: I think it would be more appropriate to use the PackageURL class instead of String to handle the object and only use the String representation when transfering it to SW360.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -72,18 +74,13 @@ private static void setCPEId(SW360Release release, Artifact artifact) {
                 .ifPresent(release::setCpeId);
     }
 
-    private static void setCoordinates(SW360Release release, Artifact artifact) {
-        release.setCoordinates(getMapOfCoordinates(artifact));
-    }
+    private static void setPackageUrls(SW360Release sw360Release, Artifact artifact) {
+        Set<String> packageURLS = artifact.askForAll(ArtifactCoordinates.class).stream()
+                .map(ArtifactCoordinates::getPurl)
+                .map(PackageURL::toString)
```
blaumeiser-at-bosch: I am not sure if toString is the right method, the "official" method for getting the String representation of the PURL is canonicalize().

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +130,34 @@ public void setDownloadurl(String downloadurl) {
...
 
-    public SW360Release setCoordinates(Map<String, String> coordinates) {
-        externalIds.putAll(coordinates);
+    public SW360Release setPurls(Set<String> purls) {
+        dropAllPurls();
+
+        int i = 1;
+        for (String purl : purls) {
+            if (purl != null && !purl.isEmpty()) {
+                externalIds.put(PURL_PREFIX + i, purl);
```
maxhbr: why are the keys just the purls with enumeration, we have a better option. In my opinion it would be better to use the old prefix, something like:
```
   'mvn': 'pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1'
```
This also allows for better searches via the SW360 rest interface.

Or maybe just use the scheme as key. (yes, there can be multiple purls of the same scheme for the same artifact. But I do not think that this is a problem here)

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
```diff
@@ -242,7 +244,13 @@ private SW360Release mkSW360Release(String name) {
...
         sw360Release.setDeclaredLicense("Apache-2.0");
         sw360Release.setObservedLicense("A Test License");
-        sw360Release.setCoordinates(Collections.singletonMap(SW360CoordinateKeysToArtifactCoordinates.get(MavenCoordinates.class), "test:test1:1.2.3"));
+        sw360Release.setPurls(Collections.singleton(PackageURLBuilder.aPackageURL()
+                .withName("test")
+                .withType("maven")
+                .withNamespace("test")
+                .withVersion("1.2.3")
+                .build()
+                .toString()));
```
maxhbr: same as https://github.com/eclipse/antenna/pull/291/files#r326196146, use `canonicalize()`.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -70,8 +70,8 @@ private void addClearingStateIfAvailable(Artifact artifact, SW360Release release
     }
 
     private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact artifact) {
-        mapCoordinates(sw360Release)
-                .forEach(artifact::addFact);
+        mapCoordinatesFromPurls(sw360Release).
```
maxhbr: usually we place the dot in the next line

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
neubs-bsi: @blaumeiser-at-bosch what do you think?

blaumeiser-at-bosch: Interesting point, disadvantage here is that I have to know which kind of component that is so searching is easier on the one hand side, because I do not have to search until the index ends, but it is also harder, because I have to find out which types of purls are there.

In general, we should use the same type as key, not anything else, so in the example case, "maven" instead of "mvn"

With scheme you mean the type, right?

maxhbr: Yes, I probably mean `type`.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
sschuberth: > the metadata for the component are the same, as the sources from which the component has been build are the same.

IMO we can and should use PURL to make exactly that distinction: To differentiate between different packagings of the same component or source code (e.g. `.tar.gz` vs. `.zip`), and to differentiate between different locations from which the same package (e.g. Maven Central vs. JCenter).

Linking some (semi-)related issue discussions here:

"separate type from provider": https://github.com/package-url/purl-spec/issues/33
"Source of repackaged code": https://github.com/clearlydefined/curated-data/issues/25

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
sschuberth: While [toString() indeed just calls canonicalize()](https://github.com/package-url/packageurl-java/blob/29a68812366bd12b0541edcb1487e175b98fbaf6/src/main/java/com/github/packageurl/PackageURL.java#L366-L369) I'd also prefer to be explicit and directly call `canonicalize()` here.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
maxhbr: on the SW360 we do not distinguish between an artifact, its sources, or whether it was pulled from maven or from github. This means that in SW360 all the corresponding PURLs should be added to one release.

I think two artifacts with the same name and version are the same item in SW360.

neubs-bsi: @maxhbr  what format would you suggest for the case: two purls from the same type?
(like started in this conversation: https://github.com/eclipse/antenna/pull/291#discussion_r326483216)

blaumeiser-at-bosch: Going over the discussion between you @sschuberth and Jeff, it seems that I have a different understanding of what a purl references. From what I see, a purl references a unique id/coordinates for a component release unifying the different ways of how component coordinates are represented between technologies.

That means, that for one technology (e.g. maven) I would expect exactly one purl that references a version of a component. Having said that, a technology means basically a packaging type or a package manager.

There are several copnsequences:
1. A component release available in different package manager technologies has multiple purls, each of which uniquely identifies the component release.
2. I do not care about the provider or source of the component, if it is the content, it has the corresponding purl of the package manager technology, since both the instances are the same thing, i.e., built from the same source code.
3. If the content is different, then it is a different release, i.e., it has a different purl
4. Whether I trist that the release has the content the purl indicates is not a matter of purl but a matter of do I trust the source of this release.
5. The source of the release is something I would like to know but this is our of scope of the purl spec.
6. The purl type indentifies the technology with which I use the component in the build, aka as the package manager technology. I still can download the component from the source and build it on my own, but even then the normal way would be to build the corresponding library and use it as if I would download it directly from the internet: e.g., Build a Jar, deploy it to a maven repo manager (e.g. Nexus) and reference it with the repo manager as provider in my pom.xml.
7. If I do not use it as described in 6, e.g., if I download the source and directly build it together with my sources, this still is the same content, but from my point, I would need to use a different purl, e.g., with type github, because the way I use it is different. I do not use the component as Maven component anymore, but as bunch of sources, so different purl.

I think that covers my line of thinking.

neubs-bsi: Specifically for this PR and how to implement the solution, what is the implication of your line of thinking for this implementation?

blaumeiser-at-bosch: - We have multiple purls per release
- The purls should be requested as proposed in the ArtifactUtils default sequence, not findAny or findFirst
- Since I assume only one purl per technology, I would follow Max comment from below that the key in the external id should be the type used in the purl as redundant element
- Purls are used in the Release object with PackageURL class not string and only when trnaslated into the JSON the string representation is used.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
maxhbr: I think that the `SW360Release` should be as near as possible to the representation which will be transferred and therefore I think that it is perfectly fine to have them as string. Especially since they are stored in the `externalId` map.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -31,7 +34,10 @@
     private static final String CHANGESTATUS_KEY = "change_status";
     private static final String COPYRIGHTS_KEY = "copyrights";
     private static final String CLEARINGSTATE_KEY = "clearingState";
-    private static final String PURL_PREFIX = "purl_";
+    private static final Set<String> PURL_TYPE = Stream.of(
+            "bitbucket", "composer", "deb", "docker",
+            "gem", "generic", "github", "golang",
```
blaumeiser-at-bosch: We have type p2 as well as custom type for p2 repository links

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +136,29 @@ public void setDownloadurl(String downloadurl) {
         this.downloadurl = downloadurl;
     }
 
-
     @JsonIgnore
-    public Map<String, String> getCoordinates() {
+    public Map<String, String> getPurls() {
         return externalIds.entrySet().stream()
-                .filter(e -> SW360CoordinateKeysToArtifactCoordinates.getValues().contains(e.getKey()))
+                .filter(e ->PURL_TYPE.stream()
```
blaumeiser-at-bosch: Should be a space behind the arrow

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +136,29 @@ public void setDownloadurl(String downloadurl) {
...
+                .anyMatch(purlType -> e.getKey().equals(purlType)))
```
blaumeiser-at-bosch: Wrong indentation, that is part of the above filter, perhaps do not do it in a new line or do it as PURL_TYPE.contains(e.getKey())

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +136,29 @@ public void setDownloadurl(String downloadurl) {
...
+
+        purls.forEach(externalIds::put);
+
         return this;
     }
 
+    private void dropAllPurls() {
+        externalIds.keySet().stream()
+                .filter(e -> PURL_TYPE.stream()
+                        .anyMatch(e::startsWith))
```
blaumeiser-at-bosch: Why startsWith? Isn't that redundant code to my comment above, should then be a method, I think

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +136,29 @@ public void setDownloadurl(String downloadurl) {
...
+                .anyMatch(purlType -> e.getKey().equals(purlType)))
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
-    public SW360Release setCoordinates(Map<String, String> coordinates) {
-        externalIds.putAll(coordinates);
+    public SW360Release setPurls(Map<String, String> purls) {
+        dropAllPurls();
+
+        purls.forEach(externalIds::put);
```
blaumeiser-at-bosch: Why not putAll?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -34,7 +37,7 @@ public static SW360Release prepareRelease(SW360Component component, Set<String>
         release.setComponentId(componentId);
         release.setMainLicenseIds(sw360LicenseIds);
 
-        SW360ReleaseAdapterUtils.setCoordinates(release, artifact);
+        SW360ReleaseAdapterUtils.setPackageUrls(release,artifact);
```
blaumeiser-at-bosch: Space missing after ,

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,19 +136,27 @@ public void setDownloadurl(String downloadurl) {
...
-        externalIds.putAll(coordinates);
+    public SW360Release setPurls(Map<String, String> purls) {
+        dropAllPurls();
+
+        externalIds.putAll(purls);
+
         return this;
     }
 
+    private void dropAllPurls() {
```
maxhbr: this drops the `docker` purl or other unsupported purls, without replacing it later? that should not happen.

maxhbr: I think PURLs, which we do not handle, should be preserved in the release.

neubs-bsi: Just for clarification: you suggest to overwrite the `externalIds` entries with the entries we get when `setPurls` is called? (which would happen automatically due to it being a map when using `putAll`)

maxhbr: I am unsure, maybe one wants do drop the `p2` coordinates (and other supported types), if
- they are explicitly not provided or
- if they should be contained in the antenna artifact, if they were present previously.

maxhbr: maybe best solution is: just remove `dropAllPurls`

blaumeiser-at-bosch: Good, point @maxhbr 
Actually, the question is, what setPurls does, Is it resetting the object or simply adding the purls given. Currently it completely resets the object concerning purls. Could be the right thing to do. How often will the case happen, that there are already purls inside the object and we add more?

When I read this, I was also unsure, if the deletion should be done, but after rethinking it now, I would propose to not delete the purls and check the concrete use cases first, before we lose data.

neubs-bsi: Currently `setPurls()` is only called in Antenna when creating a new release from artifact data. 
Hence, currently there would never be any purls already safed in the release. The `mergeWith` function of the release merges all externalIds overall.
Hence, currently the `dropAllPurls` function is redundant.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -91,35 +91,19 @@ private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact ar
         artifact.addFact(new CopyrightStatement(sw360Release.getCopyrights()));
     }
 
+    private Stream<ArtifactCoordinates> mapCoordinatesFromPurls(SW360Release sw360Release) {
+        final Map<String, String> purls = sw360Release.getPurls();
+        return purls.values().stream()
+                .map(ArtifactUtils::createArtifactCoordinatesFromPurl);
```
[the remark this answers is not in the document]
maxhbr: If there are multiple non-supported types of purls, all of them get mapped to `GenericCoordinates` and thus only one of them will survive as a fact in the antenna artifact (the last one added). Here we lose information again.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -129,16 +136,16 @@ public void setDownloadurl(String downloadurl) {
         this.downloadurl = downloadurl;
     }
 
-
     @JsonIgnore
-    public Map<String, String> getCoordinates() {
+    public Map<String, String> getPurls() {
```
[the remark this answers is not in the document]
maxhbr: related to the comments below.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: You are right, that we lose information here, but this is for now an acceptable gap. What is missing is a general coordinates type for purl based coordinates. I would propose to create an issue about the gap and let the functionality be as it is. We cannot handle the information, so Antenna should not be used in a context, which is not supported. With the issue, we have something to look at, if this raises a concrete problem in the future.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: That is the problem with the key as the type of the corresponding coordinates. Perhaps the better solution is to check for the value of the Entry to be a purl, i.e., being a url with protocol "pkg"? This way we get every purl even those we do not know as of now?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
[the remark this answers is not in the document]
maxhbr: off-topic:

> What is missing is a general coordinates type for purl based coordinates.

I do not think that is that easy. If there is a generic coordinates type for purl based coordinates it collides with the already implemented coordinates. If we have multiple ways of representing the same information, all equals instances and the way we use facts in artifacts break.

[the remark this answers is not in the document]
maxhbr: on-topic:

> You are right, that we lose information here, but this is for now an acceptable gap.

I do not think that this is an acceptable gap, if we just can fix the problem by ignoring unsupported keys.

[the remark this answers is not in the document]
blaumeiser-at-bosch: You mean that we lose information, but not transform information into a different purl type like in the generic case. THis is right, I would follow you on that.

maxhbr:
> That is the problem with the key as the type of the corresponding coordinates. Perhaps the better solution is to check for the value of the Entry to be a purl, i.e., being a url with protocol "pkg"? This way we get every purl even those we do not know as of now?

If we just get every purl, we should first start supporting all features of purls and all types. Until then it is a good idea to restrict on keys which we control. This is a kind of namespacing and we should not try to understand everything every other tool does.

And there is also the problem that, if there are multiple maven purls, we end up in a nondeterministic state, where we do not know which purl is taken.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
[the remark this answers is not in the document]
maxhbr: If we just get every purl, we should first start supporting all features of purls and all types. Until then it is a good idea to restrict on keys which we control. This is a kind of namespacing and we should not try to understand everything every other tool does.

And there is also the problem that, if there are multiple maven purls, we end up in a nondeterministic state, where we do not know which purl is taken.

blaumeiser-at-bosch:
> > That is the problem with the key as the type of the corresponding coordinates. Perhaps the better solution is to check for the value of the Entry to be a purl, i.e., being a url with protocol "pkg"? This way we get every purl even those we do not know as of now?
> 
> If we just get every purl, we should first start supporting all features of purls and all types. Until then it is a good idea to restrict on keys which we control. This is a kind of namespacing and we should not try to understand everything every other tool does.
> 
> And there is also the problem that, if there are multiple maven purls, we end up in a nondeterministic state, where we do not know which purl is taken.

This open up a different problem, if we allow to have multiple purls with the same type, we need also a change in the storage model, because currently we can only hold one purl per type in SW360. So we cannot handle the case of multiple purls of the same type anyhow. This is a restriction for now. We would end up with <type>_x again to handle this, but then we run into trouble within the Coordinates as well. This becomes too big for this change.

If we can identify all purls, we should return it, the SW360 module does not have to mind the model gaps from the internal Antenna model, if we can easily find out any purl stored, we can filter at another place, e.g., when we create the coordinates objects.

blaumeiser-at-bosch:
On hold: @maxhbr implements a change in ArtifactCoordinates reducing the different types to one type which influences the way this PR needs to be implemented

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/generator/SW360UpdaterTest.java
```diff
@@ -56,9 +56,8 @@ public void artifactIsMappedToSw360ReleaseCorrectly() {
             assertThat(false).isTrue();
         }
 
-        assertThat(release.getCoordinates()).containsKeys("mvn");
-        assertThat(release.getCoordinates()).hasSize(1);
-        assertThat(release.getCoordinates()).containsValue("org.group.id:artifactId(test1):1.2.3");
```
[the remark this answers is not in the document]
neubs-bsi: I added the test in the updater test

neubs-bsi:
@blaumeiser-at-bosch && @maxhbr 

Hello,
after our planning today, I started cleaning up here. 
The commit integrates most cleared points now. 
Comments that are not addressed yet are the following:
* [Round trip of purls](https://github.com/eclipse/antenna/pull/291#discussion_r327550528)
* [name of the `getPurls()` function in `SW360Release`](https://github.com/eclipse/antenna/pull/291#discussion_r327548494)
*

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Ok, for now, will be solved after current release with the ArtifactCoordinates change

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Will be solved with the ArtifactCoordinates change that comes after the next release, for the temporary situation ok

blaumeiser-at-bosch:
@neubs-bsi Please rebase

PR #292, opened 2019-09-17
feat(purl): add a purl to ArtifactCoordinates parser

bs-ondem:
Adds a utility method to parse a PURL to a specific `ArtifactCoordinates` object.

### DIFF
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactUtils.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactUtils.java
@@ -10,6 +10,8 @@
  */
 package org.eclipse.sw360.antenna.model.util;
 
+import com.github.packageurl.MalformedPackageURLException;
+import com.github.packageurl.PackageURL;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
 import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
 import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactCoordinates;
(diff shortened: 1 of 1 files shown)


core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactUtils.java
```diff
@@ -76,4 +78,23 @@ private ArtifactUtils() {
...
     }
+
+    public static ArtifactCoordinates createArtifactCoordinates(PackageURL packageURL) {
+        String type = packageURL.getType();
+
+        switch (type) {
+            case "maven":
+                return new MavenCoordinates(packageURL.getName(), packageURL.getNamespace(), packageURL.getVersion());
+            case "npm":
+                return new JavaScriptCoordinates(packageURL.getName(), packageURL.getNamespace(), packageURL.getVersion());
```
blaumeiser-at-bosch: I have added p2 as a type for BundleCoordinates although it is not a standardized type, but since we need it.

core/model/src/main/java/org/eclipse/sw360/antenna/model/util/ArtifactUtils.java
```diff
@@ -76,4 +78,23 @@ private ArtifactUtils() {
...
+            case "nuget":
+                return new DotNetCoordinates(packageURL.getName(), packageURL.getVersion());
+            default:
+                return new GenericArtifactCoordinates(packageURL.getName(), packageURL.getVersion());
+        }
+    }
+
+    public static ArtifactCoordinates createArtifactCoordinates(String packageUrl) throws MalformedPackageURLException {
```
blaumeiser-at-bosch: Please do not throw the very specific exception. My proposal would be to use a IllegalArgumentException, since the String is not fulfilling the precondition of the packageUrl parameter.

neubs-bsi: Function name is not fully clear. Unknowing developer could give any kind of string and hope for `ArtifactCoordinates`. Maybe add something like: `fromPurl `or `fromPurlString`

bs-ondem:
@neubs-bsi @blaumeiser-at-bosch  please check my last force-push with your recommended changes.

PR #294, opened 2019-09-19
Ensure correct roundtrip of purl functionality

blaumeiser-at-bosch:
* Add a roundtrip test starting from a purl and comparing the result of the AntennaCoordinates purl creator with the original purl for all coordinates types except generic
* Fix bugs in the purl generation of AntennaCoordinates

### DIFF
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
@@ -69,7 +69,7 @@ public class MavenCoordinates extends JavaCoordinates<MavenCoordinates> {
 
     @Override
     protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
-        return builder.withNamespace(getGroupId());
+        return builder.withNamespace(getGroupId()).withName(getArtifactId());
     }
 
     @Override
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
@@ -11,6 +11,7 @@
 
 package org.eclipse.sw360.antenna.model.artifact.facts.javaScript;
 
+import com.github.packageurl.PackageURLBuilder;
 import org.eclipse.sw360.antenna.model.artifact.ArtifactFactBuilder;
 import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactCoordinates;
 import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactIdentifier;
@@ -65,6 +66,11 @@ public class JavaScriptCoordinates extends ArtifactCoordinates<JavaScriptCoordin
     @Override
     public String getType() {
         return "npm";
+    }
+
+    @Override
+    protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
+        return builder.withNamespace(getName()).withName(getArtifactId());
     }
 
     @Override
--- a/core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
+++ b/core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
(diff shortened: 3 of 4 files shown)


core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
```diff
@@ -67,6 +68,11 @@ public String getType() {
         return "npm";
     }
 
+    @Override
+    protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
+        return builder.withNamespace(getName()).withName(getArtifactId());
```
maxhbr: what about the `version`? It is handled somewhere else?

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
```diff
@@ -69,7 +69,7 @@ public String getType() {
 
     @Override
     protected PackageURLBuilder addPurlFacts(PackageURLBuilder builder) {
-        return builder.withNamespace(getGroupId());
+        return builder.withNamespace(getGroupId()).withName(getArtifactId());
```
maxhbr: what about the `version`? It is handled somewhere else?

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        ArtifactCoordinates coordinates = ArtifactUtils.createArtifactCoordinatesFromPurl(purl);
+
+        assertThat(coordinates).isInstanceOf(DotNetCoordinates.class);
+        assertThat(coordinates.getPurl().canonicalize()).isEqualTo(purl);
+    }
+
+    @Test
+    public void testPurlParserWithJavaScriptCoordinates() {
+        String purl = "pkg:npm/%40angular/animation@1.2.3";
```
maxhbr: Not a part of this PR but a problem I have with purl: i dislike how purl handles the escaping of `@` as `%40` here. I would expect an URL to be stable under URL de- and en-coding (since some utils might use it).

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.sw360.antenna.model.artifact.facts.dotnet.DotNetCoordinates;
+import org.eclipse.sw360.antenna.model.artifact.facts.java.BundleCoordinates;
+import org.eclipse.sw360.antenna.model.artifact.facts.java.MavenCoordinates;
+import org.eclipse.sw360.antenna.model.artifact.facts.javaScript.JavaScriptCoordinates;
+import org.eclipse.sw360.antenna.model.util.ArtifactUtils;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ArtifactUtilPurlTest {
```
maxhbr: This test can be written as a parametric test, to reduce code duplication.

This test should also look at coordinates without versions.

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
maxhbr: Oh sorry, that is absolutely correct and matches the spec of URL-encoding

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
blaumeiser-at-bosch: Version is handled in the base class, as is name, the addPurlFacts is only used for special coordinate parts which are not handled in the standard way, like name which is according to Onur the namespace and in Maven, where getName returns groupId:artifactId

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/java/MavenCoordinates.java
blaumeiser-at-bosch: See above

core/model/src/test/java/org/eclipse/sw360/antenna/model/test/ArtifactUtilPurlTest.java
blaumeiser-at-bosch: Done in the latest commit

PR #295, opened 2019-09-20
Add SonarCloud Analysis to Antenna

blaumeiser-at-bosch:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,11 @@ addons:
   apt:
     packages:
       - git
+  sonarcloud:
+    organization: "sw360antenna"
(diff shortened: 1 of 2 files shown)


blaumeiser-at-bosch:
Remark: Due to the fact, that Travis requires in the encrypted credentials for Sonartype the name of the repository (eclipse/antenna), the mechanism does not succeed in the case of a PR build that is raised from a fork towards the original repo. I created the mechanism as such, that the analysis is only triggered in master builds in Travis.

As a consequence, this PR does nothing, but I need it to be merged in order to check that it works in the master branch builds.

PR #297, opened 2019-09-20
Migrate main Java code of the ORT module to Kotlin

sschuberth:
To show how integration of Kotlin source code works.


modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -0,0 +1,56 @@
...
+    init {
+        workflowStepOrder = 700
+    }
+
+    @Throws(AntennaException::class)
+    override fun yield(): WorkflowStepResult =
+        try {
+            WorkflowStepResult(createArtifactList(componentInfoFile))
+        } catch (e: IOException) {
+            throw AntennaException("Error parsing the ORT result file: ${e.message}")
```
blaumeiser-at-bosch: I propose to use the ExecutionException, which is redundant from my point of view, it is also already a runtime exception, so no throwing necessary

sschuberth: Ok, I'll remove the catch / throw in a new commit. I'd like to keep it in the original porting commit to more closely resemble the original Java code in the migration step.

sschuberth:
> Just needs squashing

Why squashing?

PR #299, opened 2019-09-24
infra(fix): Fix the reference of repositories and vulnerabilities 

blaumeiser-at-bosch:
- Add the needed repositories to build from scratch
  - spring.io for gradle
  - jitpack.io for ort
- Update jackson databind


blaumeiser-at-bosch:
@maxhbr @bs-ondem 
With these changes, build is now successful, and also there is progress on Azure.

To be honest, one of those changes where it is unclear why it helps, and why it did't break in the past but only recently in some environments

PR #300, opened 2019-09-24
Use PURLs for artifact coordinates

maxhbr:
TODO:
- [ ] ~~pretty printing~~
- [ ] ~~not as fact?~~
- [x] Artifact selectors and their matching
- [x] find best purl?


core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/CsvAnalyzerImpl.java
```diff
@@ -139,15 +136,15 @@ private ArtifactCoordinates createCoordinates(CSVRecord record) {
 
         switch (type) {
             case "mvn":
-                return new MavenCoordinates(record.get(NAME), record.get(GROUP), record.get(VERSION));
```
blaumeiser-at-bosch: To be honest I hoped that we get rid of these switch stuff with the change. A constructor that takes type, namespace, name and version and perhaps one without namespace. This looks very complex and it actually always the same.

maxhbr: the `ArtifactCoordinatesUtils` are just a tool to make the code migration easy. This is an intermediate state of the migration, of a WIP PR.

maxhbr: With the current dev state, one would write:
```java
    private Coordinate createCoordinate(CSVRecord record) {
        final Coordinate.Builder builder = Coordinate.builder()
                .withName(record.get(NAME))
                .withVersion(record.get(VERSION));

        String type = record.isMapped(COORDINATE_TYPE) ? record.get(COORDINATE_TYPE) : "mvn";
        switch (type) {
            case "mvn":
                builder.withType(Coordinate.Types.MAVEN);
                builder.withNamespace(record.get(GROUP));
            case "dotnet":
                builder.withType(Coordinate.Types.NUGET);
            case "javascript":
                builder.withType(Coordinate.Types.NPM);
                builder.withNamespace(record.get(GROUP));
            case "bundle":
                builder.withType(Coordinate.Types.BUNDLE);
        }
        
        return builder.build();
    }
```

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/CsvAnalyzerImpl.java
```diff
@@ -134,21 +132,26 @@ private Artifact createNewArtifact(CSVRecord record) {
...
 
+        String type = record.isMapped(COORDINATE_TYPE) ? record.get(COORDINATE_TYPE) : "mvn";
         switch (type) {
             case "mvn":
-                return new MavenCoordinates(record.get(NAME), record.get(GROUP), record.get(VERSION));
+                builder.withType(Coordinate.Types.MAVEN);
+                builder.withNamespace(record.get(GROUP));
             case "dotnet":
-                return new DotNetCoordinates(record.get(NAME), record.get(VERSION));
+                builder.withType(Coordinate.Types.NUGET);
```
blaumeiser-at-bosch: break; ???

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/ArtifactCoordinates.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public boolean containsPurl(String coordinateString)  {
+        return containsPurl(Coordinate.of(coordinateString));
+    }
+
+    public boolean containsPurl(Coordinate coordinate) {
+        return coordinates.values().contains(coordinate);
+    }
+
+    public Set<Coordinate> getCoordinates() {
```
blaumeiser-at-bosch: There we are again in the discussion whether Collection is not the better result. Actually in such a case an unmodifiable Collection seems to represent the information given back the best, IMO

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/ArtifactCore.java
```diff
@@ -63,11 +64,19 @@ public ArtifactCore addFact(ArtifactFact artifactFact) {
...
-    public ArtifactCore addFact(ArtifactFactBuilder artifactFactBuilder) {
-        if(artifactFactBuilder != null) {
-            return addFact(artifactFactBuilder.build());
-        }
-        return this;
+    public ArtifactCore addCoordinate(Coordinate coordinate) {
+        return addFact(new ArtifactCoordinates(coordinate));
+    }
+
+    public Set<Coordinate> getCoordinates() {
```
blaumeiser-at-bosch: See above about Set vs. Collection

core/model/src/main/java/org/eclipse/sw360/antenna/model/coordinates/Coordinate.java
```diff
@@ -0,0 +1,224 @@
+package org.eclipse.sw360.antenna.model.coordinates;
```
blaumeiser-at-bosch: Puh, this whole wrapping is awful lot of code and since it is a pure delegate with everything PackageURL offers, it feels a lot like explicit complexity for not much gain. Does not feel right.

modules/ort/src/main/java/org/eclipse/sw360/antenna/ort/resolver/OrtResultArtifactResolver.java
```diff
@@ -52,49 +50,25 @@ public Artifact apply(Package pkg) {
...
         String version = pkg.getId().getVersion();
 
+        final Coordinate.Builder builder = Coordinate.builder()
+                .withName(name)
+                .withVersion(version);
         switch (pkg.getId().getType().toLowerCase()) {
             case "nuget":
             case "dotnet":
-                return Optional.of(mapDotNetCoordinates(name, version));
+                builder.withType(Coordinate.Types.NUGET);
```
blaumeiser-at-bosch: break;???

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/ArtifactCoordinates.java
maxhbr: `Collection` is the better input. But specific types are the better result.

If a function understands `Collection` as **input, one can understand everything**. If one outputs a `Collection`, **the caller must understand all types of collections**. If I return a Set the user still can treat it as a Collection, so nothing is lost here.

core/model/src/main/java/org/eclipse/sw360/antenna/model/coordinates/Coordinate.java
maxhbr: What features would you be willing to sacrifice for a less complex wrapping?

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/analyzers/CsvAnalyzerImpl.java
blaumeiser-at-bosch: RIght, but wouldn't be the default generic which would be the default case of the switch?

core/model/src/main/java/org/eclipse/sw360/antenna/model/coordinates/Coordinate.java
blaumeiser-at-bosch: Actually my point is that it is so close to the PackageURL concept, that I do not see an easy way to replace PackageURL with something else. Since this only wraps without any logic, and transfers the PackageURL model directly I think we should use PackageURL directly knowing that this dependency then spready all over the code, but it will be an easy find and replace if we implement the class on our own in Antenna and replace the import with the local implementation to get rid of the dependency.

blaumeiser-at-bosch: After discussion ok

maxhbr:
note for self: commit before squash: e62edecd3b52aeacc6d83aa7cf3d3099650708e2

PR #304, opened 2019-09-30
Remove the bug warnings from SonarCloud

blaumeiser-at-bosch:

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,1 +1,0 @@
-theme: jekyll-theme-cayman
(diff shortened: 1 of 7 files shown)


.ci-scripts/test-for-licenseHeaders.sh
```diff
@@ -59,7 +59,7 @@ getFiles() {
     fi |
         grep -Ev '^./.git' |
         grep -Ev '\.(csv|rdf|ent|dtd|png|gitignore|gitattributes|md|bat|jar|json|couch|view|MF|xz|index)' |
-        grep -Ev '(gradlew|build.gradle|settings.gradle|gradle.properties|gradle/wrapper)' |
+        grep -Ev '(gradlew|build.gradle|settings.gradle|gradle.properties|gradle/wrapper|_config.yml)' |
```
neubs-bsi: I believe this can be deleted, it should be a residue from activating gh pages with some difficulties. I would rather not have this file in the master branch. On our example project no such file was necessary and gh pages was still working.

PR #309, opened 2019-10-02
Fix the null license issue

blaumeiser-at-bosch:
Make the creation of license objects aware of null information

Add a more cautious creation of swh ids.

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
@@ -17,6 +17,7 @@ import org.eclipse.sw360.antenna.model.artifact.facts.*;
 import org.eclipse.sw360.antenna.model.reporting.MessageType;
 import org.eclipse.sw360.antenna.model.util.ArtifactLicenseUtils;
 import org.eclipse.sw360.antenna.model.xml.generated.License;
+import org.eclipse.sw360.antenna.model.xml.generated.LicenseInformation;
 import org.eclipse.sw360.antenna.model.xml.generated.LicenseOperator;
 import org.eclipse.sw360.antenna.model.xml.generated.LicenseStatement;
 import org.eclipse.sw360.antenna.sw360.SW360MetaDataReceiver;
@@ -29,6 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -73,13 +75,18 @@ public class SW360EnricherImpl {
         mapCoordinates(sw360Release)
                 .forEach(artifact::addFact);
 
-        artifact.addFact(new DeclaredLicenseInformation(makeLicenseStatementFromString(sw360Release.getDeclaredLicense())));
-        artifact.addFact(new ObservedLicenseInformation(makeLicenseStatementFromString(sw360Release.getObservedLicense())));
-        artifact.addFact(new ArtifactReleaseTagURL(sw360Release.getReleaseTagUrl()));
+        addLicenseFact(Optional.ofNullable(sw360Release.getDeclaredLicense()), artifact, DeclaredLicenseInformation::new);
+        addLicenseFact(Optional.ofNullable(sw360Release.getObservedLicense()), artifact, ObservedLicenseInformation::new);
+        Optional.ofNullable(sw360Release.getReleaseTagUrl())
+                .map(ArtifactReleaseTagURL::new)
+                .ifPresent(artifact::addFact);
         try {
-            artifact.addFact(new ArtifactSoftwareHeritageID.Builder(sw360Release.getSoftwareHeritageId()).build());
+            Optional.ofNullable(sw360Release.getSoftwareHeritageId())
+                    .map(ArtifactSoftwareHeritageID.Builder::new)
+                    .map(ArtifactSoftwareHeritageID.Builder::build)
+                    .ifPresent(artifact::addFact);
         } catch (IllegalArgumentException e) {
-            LOGGER.warn(e.getMessage());
+            LOGGER.info(e.getMessage());
         }
         sw360Release.getHashes().stream()
                 .map(hash -> new ArtifactFilename(null, hash))
@@ -88,7 +95,15 @@ public class SW360EnricherImpl {
                 .map(ArtifactChangeStatus.ChangeStatus::valueOf)
                 .map(ArtifactChangeStatus::new)
                 .ifPresent(artifact::addFact);
-        artifact.addFact(new CopyrightStatement(sw360Release.getCopyrights()));
+        Optional.ofNullable(sw360Release.getCopyrights())
+                .map(CopyrightStatement::new)
+                .ifPresent(artifact::addFact);
+    }
+
+    private void addLicenseFact(Optional<String> licenseList, Artifact artifact, Function<LicenseInformation, ArtifactLicenseInformation> licenseCreator) {
+        licenseList.map(this::makeLicenseStatementFromString)
+                .map(licenseCreator::apply)
+                .ifPresent(artifact::addFact);
     }
 
     private License makeLicenseStatementFromString(String license) {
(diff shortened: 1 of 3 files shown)


blaumeiser-at-bosch:
WIP: I committed too much

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -73,12 +76,15 @@ private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact ar
         mapCoordinates(sw360Release)
                 .forEach(artifact::addFact);
 
-        artifact.addFact(new DeclaredLicenseInformation(makeLicenseStatementFromString(sw360Release.getDeclaredLicense())));
-        artifact.addFact(new ObservedLicenseInformation(makeLicenseStatementFromString(sw360Release.getObservedLicense())));
+        addLicenseFact(Optional.ofNullable(sw360Release.getDeclaredLicense()), artifact, DeclaredLicenseInformation.class);
```
maxhbr: The function `makeLicenseStatementFromString` is now unused

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -73,12 +76,15 @@ private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact ar
         mapCoordinates(sw360Release)
                 .forEach(artifact::addFact);
 
-        artifact.addFact(new DeclaredLicenseInformation(makeLicenseStatementFromString(sw360Release.getDeclaredLicense())));
```
maxhbr: This can be solved in a simpler way:
```suggestion
        Optional.ofNullable(sw360Release.getDeclaredLicense())
            .map(this::makeLicenseStatementFromString)
            .map(DeclaredLicenseInformation::new)
            .ifPresent(artifact::add);
        Optional.ofNullable(sw360Release.getObservedLicense())
            .map(this::makeLicenseStatementFromString)
            .map(ObservedLicenseInformation::new)
            .ifPresent(artifact::add);
```

This results in less code and does not use reflection.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -91,6 +97,21 @@ private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact ar
         artifact.addFact(new CopyrightStatement(sw360Release.getCopyrights()));
     }
 
+    private void addLicenseFact(Optional<String> licenseList, Artifact artifact, Class<? extends ArtifactLicenseInformation> licenseClass) {
```
maxhbr: Why not simply pass the constructor as function (e.g. `DeclaredLicenseInformation::new`) instead of unnecessary use of reflection?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
blaumeiser-at-bosch: That is not correct, it is used in the addLicenseFact method in a map statement.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
blaumeiser-at-bosch: Lastest change reflects your proposal

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
blaumeiser-at-bosch: This is so ugly :-)

blaumeiser-at-bosch:
Fixes: https://github.com/bsinno/osm-antenna/issues/844

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -73,11 +75,14 @@ private void mapExternalIdsOnSW360Release(SW360Release sw360Release, Artifact ar
         mapCoordinates(sw360Release)
                 .forEach(artifact::addFact);
 
-        artifact.addFact(new DeclaredLicenseInformation(makeLicenseStatementFromString(sw360Release.getDeclaredLicense())));
-        artifact.addFact(new ObservedLicenseInformation(makeLicenseStatementFromString(sw360Release.getObservedLicense())));
+        addLicenseFact(Optional.ofNullable(sw360Release.getDeclaredLicense()), artifact, DeclaredLicenseInformation::new);
+        addLicenseFact(Optional.ofNullable(sw360Release.getObservedLicense()), artifact, ObservedLicenseInformation::new);
         artifact.addFact(new ArtifactReleaseTagURL(sw360Release.getReleaseTagUrl()));
```
neubs-bsi: should the other data variables not also be put into an optional? (e.g. CopyrightStatement and ArtifactReleaseTagURL)

blaumeiser-at-bosch: Good point

maxhbr: facts are build in a way, that every fact implements an `.isEmpty()` method. And if it is empty it is treated similar to a `Optional.empty()` and adding an empty fact is a no-op.

neubs-bsi: Does this apply to all of them or is there another one with special creation conditions? Since there have been issues with empty softwareheritage id and license? Even though they are special cases because they have special creation conditions.

maxhbr: All facts implement that method:

https://github.com/eclipse/antenna/blob/a89d2bddace67078617296b2084ae8bfcb70bc81/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/ArtifactFact.java#L16

blaumeiser-at-bosch: From what I saw, adding an empty fact is not a noop, because if I do not add it, it is not there and I do not have a problem, with empty facts I have to handle explicitely that the fact is empty and this does not make sense, because if I ask for a fact, I would expect that the fact is either there and has some content or is not returned at all.

blaumeiser-at-bosch: SO if I look at the code, it would make sense to not add the fact in org.eclipse.sw360.antenna.model.artifact.ArtifactCore#addFact(org.eclipse.sw360.antenna.model.artifact.ArtifactFact), i.e., if fact.isEmpty is true. Then I would agree that it is a noop and I would think that the askFor methods would work as expected.

maxhbr: Yes. It is the other way round. One can add an empty fact and thus explicitly say that this statement should be emtpy. So adding is not a no-op. Only from the view of an empty artifact. Adding an empty fact is equal to "there is no fact at all". But if there would be a previous fact it would get overwritten.

blaumeiser-at-bosch: Ok, but then, it makes sense to not add facts in the sw360 enricher because the semantics is not that if sw360 does not know a fact, it is not there, only that sw360 does not know whether there is a fact.

PR #311, opened 2019-10-03
fix(sw360): Fix license handling in SW360Updater

blaumeiser-at-bosch:
Wrong behavior: Updater creates a license for every piece of license string found and always fill in Main Licenses
Right behavior: Updater only checks for availability of license and sets only those licenses in Main Licenses, that are already known.

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
@@ -13,7 +13,6 @@ package org.eclipse.sw360.antenna.sw360;
 
 import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
-import org.eclipse.sw360.antenna.model.util.ArtifactLicenseUtils;
 import org.eclipse.sw360.antenna.model.xml.generated.License;
 import org.eclipse.sw360.antenna.sw360.adapter.SW360ComponentClientAdapter;
 import org.eclipse.sw360.antenna.sw360.adapter.SW360LicenseClientAdapter;
@@ -28,9 +27,11 @@ import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpHeaders;
 
 import java.io.IOException;
-import java.util.*;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class SW360MetaDataUpdater {
     private static final Logger LOGGER = LoggerFactory.getLogger(SW360MetaDataUpdater.class);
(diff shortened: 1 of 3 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -52,20 +54,16 @@ public SW360MetaDataUpdater(SW360ConnectionConfiguration sw360ConnectionConfigur
...
         Set<String> licenseIds = new HashSet<>();
 
         List<License> licenses = flattenedLicenses(artifact);
         for (License license : licenses) {
-            SW360License sw360License;
-            if (!licenseClientAdapter.isLicenseOfArtifactAvailable(license, header)) {
-                sw360License = licenseClientAdapter.addLicense(license, header);
-            } else {
+            if (licenseClientAdapter.isLicenseOfArtifactAvailable(license, header)) {
                 LOGGER.debug("License [" + license.getName() + "] already exists in SW360.");
```
maxhbr:
```suggestion
                LOGGER.debug("License [" + license.getName() + "] found in SW360.");
```

maxhbr:
This just ignores unknown licenses. As a result `MIT AND some-non-commercial-license` gets listed as just `MIT`. I would suggest to not set anything in the release, if parts could not be understood.

blaumeiser-at-bosch:
That is not completely correct, this only affects the Main Licenses field in SW360, not the data in the additional data section. There the correct license expression is defined with all parts. But this way, the Main License field at least stores the links to the known licenses. So, this behavior gives the Compliance Data Expert the full information on the data situation for the curation task.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
blaumeiser-at-bosch: fixed

maxhbr:
But for SW360 and in the overviews this is hidden data. Also for the compliance exports this results in partial data.

blaumeiser-at-bosch:
This might be the case, but only for releases that have not been approved by the compliance office, from my point of view such releases do not contain reliable information.

maxhbr:
then this has to be represented in the export and overview fields.

But I still see no reason for partial data in the main licenses field.

blaumeiser-at-bosch:
> Then this has to be represented as partial / icomplete data in the export and overviews. I also think that this does not match the way, SW360 is usually used by others.
> 
> But I still see no reason for partial data in the main licenses field.

@maxhbr: Changed according to your proposal

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -252,12 +254,12 @@ public SW360Release setCopyrights(String copyrights) {
         return this;
     }
 
-    public SW360Release setExternalIds(Map<String,String> externalIds) {
+    public SW360Release setExternalIds(Map<String, String> externalIds) {
```
maxhbr: I think the old code looks better? Is this the auto-formatter?

blaumeiser-at-bosch: I think so, although a space after comma is standard, isn't it?

blaumeiser-at-bosch:
@maxhbr @neubs-bsi New change reflects the fact that the getLicenses method had an unexpected behavior, because it returned an empty list, if not all licenses were found. Now it returns a mapping that allows the UpdaterImpl code to handle explicitely with the case that not all licenses are found.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -55,22 +55,21 @@ public SW360MetaDataUpdater(SW360ConnectionConfiguration sw360ConnectionConfigur
         this.uploadSources = uploadSources;
     }
 
-    public Set<String> getOrCreateLicenses(Artifact artifact) {
+    public Map<License, Optional<SW360License>> getLicenses(Artifact artifact) {
```
maxhbr: Keys of the map are never accessed, why not instead use `List<Optional<SW360License>>`. And since, if at least one entry is empty, it is treated as empty. It might be even better to use `Optional<List<SW360License>>`.

The function `licenseClientAdapter.getSW360LicenseByAntennaLicense(license, header)` has no side effect? Why is the early exit solution gone?

blaumeiser-at-bosch: I do not agree, you are right, that the keys are not used, but I do not like lists with occacional empty stuff that lost its meaning, by using the Map, an empty value has meaning. In addition, I think this method might be usabe in other context, too. In the current change, this might already be the function I need.

I do not get your remark on the early exit solution? You mean that the method returns an empty list if at least one license is not available. I didn't like this because the method in this place should not be aware that if one license is missing, no information should be given back. This is not intuitive. So that was the reason, why I reimplemented this to return all known license and to let the updater decide what to do in case of missing licenses.

maxhbr: I thought you prefer simpler and shorter code, which does not try to solve issues which might appear in the future.

Then I would say that `Map<License, Optional<SW360License>>` is a very complicated datastructure, and it might be good to introduce a custom class to explain the implicit semantics. Or maybe add a comment.

blaumeiser-at-bosch: Have a look at my new change, I reversed the map thing, but changed the parameter slightly. I even think the functionality is now better distributed. And it allows to compare the numbers at the right place.

blaumeiser-at-bosch: Ups sorry, I answered in the wrong place, please have a look at my latest change, it should be more appropriate

PR #316, opened 2019-10-08
infra(update): Update the used ORT revision to include needed changes

sschuberth:


blaumeiser-at-bosch:
@sschuberth There seems to be a bug in ORT. At least there is a very nice exception here!

blaumeiser-at-bosch:
Merge this, small merge and I need the change urgently

PR #317, opened 2019-10-08
infra(update): Update jaxb2 to avoid a bogus warning

sschuberth:
This should make the warning

> The POM for org.glassfish.jaxb:jaxb-runtime:jar:2.2.11 is invalid

from Maven go away.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -422,12 +422,12 @@
                 <plugin>
                     <groupId>org.codehaus.mojo</groupId>
                     <artifactId>jaxb2-maven-plugin</artifactId>
-                    <version>2.2</version>
+                    <version>2.5.0</version>
                     <dependencies>
                         <dependency>
                             <groupId>org.jvnet.jaxb2_commons</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -422,12 +422,12 @@
...
                     <groupId>org.codehaus.mojo</groupId>
                     <artifactId>jaxb2-maven-plugin</artifactId>
-                    <version>2.2</version>
+                    <version>2.5.0</version>
                     <dependencies>
                         <dependency>
                             <groupId>org.jvnet.jaxb2_commons</groupId>
                             <artifactId>jaxb2-basics</artifactId>
-                            <version>0.11.0</version>
+                            <version>1.11.1</version>
```
maxhbr: maybe extract this into a property? Since the version appears twice.

sschuberth: The version "1.11.1" indeed appears twice, but for another artifact named "jaxb2-basics*-runtime*". Or are you referring to something else?

sschuberth: If so, I can fix that in a follow-up PR.

PR #322, opened 2019-10-10
Improve JavaScriptCoordinates and fix SW360ComponentAdapterUtils::createComponentName 

bs-ondem:
#### Description
This PR improves the `JavaScriptCoordinates` class with a new instance variable `namespace`. This change is made to fix the `SW360ComponentAdapterUtils::createComponentName` and to be able to represent the JS coordinates of components like `@angular/core`.

#### Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Improvements
- [ ] Documentation update
- [ ] Other

#### How Has This Been Tested?
- [x] Specific Unit Test
- [x] Executing the Example Projects
- [x] `mvn test` of whole project
- [ ] `mvn test` of submodule
- [ ] `gradle test` (for the gradle plugin)

#### Checklist
- [x] My code follows the style and guidelines of this project
- [x] I have self-reviewed my code 
- [x] I have provided tests that prove my change is working 
- [x] Existing tests still pass with my changes 

### DIFF
--- a/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
+++ b/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
@@ -268,9 +268,9 @@ public class JsonReader {
     private Optional<ArtifactCoordinates> mapJavaScriptCoordinates(JsonObject objCoordinates) {
         if (objCoordinates != null) {
             JavaScriptCoordinates.JavaScriptCoordinatesBuilder c = new JavaScriptCoordinates.JavaScriptCoordinatesBuilder();
-            c.setName((String) objCoordinates.get("name"));
+            setJavaScriptCoordinatesPackageName(c, (String) objCoordinates.get("name"));
+            setJavaScriptCoordinatesNamespace(c, (String) objCoordinates.get("name"));
             c.setVersion((String) objCoordinates.get("version"));
-            c.setArtifactId(objCoordinates.get("name") + "-" + objCoordinates.get("version"));
             return Optional.of(c.build());
         }
         return Optional.empty();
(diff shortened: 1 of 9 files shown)


core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/jsonreader/JsonReader.java
```diff
@@ -302,6 +302,32 @@ private MatchState mapMatchState(JsonObject obj) {
         return Optional.empty();
     }
 
+    private String splitJavaScriptCoordinatesName(String key, String name) {
```
blaumeiser-at-bosch: I do not like this kind of solution where you process something multiple times, because there is no way to return two values.
Perhaps better is:
setNamespace and setName.
In this case, setNamespace could even do not set the namespace at all instead of setting it with an empty string.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
```diff
@@ -70,65 +75,72 @@ public String getType() {
...
+        return builder.withNamespace(namespace).withName(packageName);
     }
 
     @Override
     public String toString() {
-        return name + ":" + version + " (" + artifactId + ")";
+        return Stream.of(namespace, packageName, version)
+                .filter(Objects::nonNull)
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.joining(":"));
```
blaumeiser-at-bosch: In Javascript, would not '/' be the right separator?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
```diff
@@ -45,7 +46,7 @@ public static String createComponentName(Artifact artifact) {
         return getMostDominantArtifactCoordinates(artifact)
                 .map(ArtifactCoordinates::getName)
                 .filter(n -> ! n.isEmpty())
-                .orElse(artifact.toString()); // TODO: ugly hack
+                .orElseThrow(() -> new ExecutionException("Component name of " + artifact.toString() + " could not be created"));
```
blaumeiser-at-bosch: I do not think, that this is a case for an Exception, no coordinates is not really very likely, since we try to handle this in most cases with generic coordinates. But if this is the case for whatever reason, the result can be something ugly, but processing should continue.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/javaScript/JavaScriptCoordinates.java
blaumeiser-at-bosch: Ok

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
blaumeiser-at-bosch: I referred to the TODO statement :-)

bs-ondem:
@blaumeiser-at-bosch Check my last fixup commit

blaumeiser-at-bosch:
Merging this since author cannot merge today

neubs-bsi:
This PR left the TODO in the code and does not actually introduce a solution for the issue:
https://github.com/eclipse/antenna/blob/5f471bce3c840759f4bf31f431d3814809bfbbee/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java#L48
There still needs to be a different solution than just printing `artifact.toString()`. 
@blaumeiser-at-bosch @bs-ondem

blaumeiser-at-bosch:
Right, but this was in the code already, something for cleaning up

PR #339, opened 2019-10-29
change condition for checking for git diff in writeIndexRootFile

neubs-bsi:
Since the Jenkins job failed when trying to checkout to current branch
because changes in branch might be lost, condition must not have worked.

Hence this changes the if condition to decide if a commit should happen.
`--quiet` was added to stop the search as soon as a difference has been
found. The command implies the git parameter `exit-code` and will exit
with 1 when the tree is modified
`--` was added for the command to work as intended, since it separates
the path from the rest of the argument and helps Git know that `HEAD`
is not the name of a file.
The condition is specifically designed to exit with an error if the
script is using `set -e`, as described here:
https://stackoverflow.com/a/28296286

#### Request Reviewer
@sschuberth 

#### Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Improvements
- [x] Documentation update
- [ ] Other

#### How Has This Been Tested?
- [x] Testing with a shell script on a mock git project

#### Checklist
- [x] My code follows the style and guidelines of this project
- [x] I have self-reviewed my code 
- [ ] I have provided tests that prove my change is working 
- [ ] Existing tests still pass with my changes 

### DIFF
--- a/antenna-documentation/scripts/writeIndexRootFile.sh
+++ b/antenna-documentation/scripts/writeIndexRootFile.sh
@@ -51,7 +51,7 @@ git clean -fd
 
 generateIndexFile | tee index.html
 # if there is a difference after writing the index html, commit
-if ! git diff-index HEAD; then
+if ! git diff-index --quiet HEAD --; then
   git add index.html
   git commit -m "update index file to be according to new version ${CURRENT_COMMIT}" -s
 fi


sschuberth:
I believe the commit message is a bit misleading, because what you want is not really `--quiet`, but the `--exit-code` option it implies. I agree it *also* makes sense to suppress the output, but that's just cosmetics and not the real fix.

Also, could you please elaborate in more detail *why* `--` is need to make the command work as intended? Usually, `--` is just used to indicate that any following argument starting with `-` (like a file name) are not options anymore to the original program, but in this case nothing follows `--` so I wonder why it would be needed.

neubs-bsi:
@sschuberth 
This is a direct copy from `git help diff-index`:
```
     --quiet
           Disable all output of the program. Implies --exit-code.
```


As well as this link explains the set up of the if condition: https://stackoverflow.com/a/28296286 

I will change the commit message to be more precise.

PR #345, opened 2019-11-11
Add an analyzer step that uses the ORT Analyzer programmatically

sschuberth:

Yes, the ORT Analyzer module is added as a dependency.

### Type of Change

new feature

A test yet needs to be added.

### Checklist
Must:


sschuberth:
The integration of ORT into Antenna causes technical problems due to a mismatch between the Maven libraries that the ORT Analyzer uses and the libraries used by the Maven version running the Antenna plugin. Due to this, but also from a workflow perspective, we decided to rather go the opposite way and integrate relevant parts of Antenna into ORT instead of further integrating ORT into Antenna. Thus, I'm closing this unmerged,

PR #355, opened 2019-11-20
Clean up logging output in Antenna

blaumeiser-at-bosch:
* Moved most info logging to debug logging
* Cleaned up the information in a normal antenna run

### Request Reviewer
@neubs-bsi 
@bs-ondem 

### Type of Change

*improvement*

### How Has This Been Tested?
Manual execution of the workflow and observe result

### Checklist
Must:

### DIFF
--- a/modules/maven/src/main/java/org/eclipse/sw360/antenna/maven/HttpRequester.java
+++ b/modules/maven/src/main/java/org/eclipse/sw360/antenna/maven/HttpRequester.java
@@ -53,7 +53,7 @@ public class HttpRequester extends IArtifactRequester {
         File localJarFile = targetDirectory.resolve(jarBaseName).toFile();
 
         if (localJarFile.exists()) {
-            LOGGER.info("The file " + localJarFile + " already exists and won't be downloaded again");
+            LOGGER.debug("The file " + localJarFile + " already exists and won't be downloaded again");
             return Optional.of(localJarFile);
         }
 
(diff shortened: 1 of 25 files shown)


modules/maven/src/main/java/org/eclipse/sw360/antenna/maven/HttpRequester.java
```diff
@@ -53,7 +53,7 @@ public HttpRequester(ProxySettings proxySettings) {
         File localJarFile = targetDirectory.resolve(jarBaseName).toFile();
 
         if (localJarFile.exists()) {
-            LOGGER.info("The file " + localJarFile + " already exists and won't be downloaded again");
+            LOGGER.debug("The file " + localJarFile + " already exists and won't be downloaded again");
```
neubs-bsi: I would make this a warning

modules/maven/src/main/java/org/eclipse/sw360/antenna/maven/MavenInvokerRequester.java
```diff
@@ -72,7 +72,7 @@ public MavenInvokerRequester(File basedir, DefaultInvoker defaultInvoker, Option
         File expectedJarFile = getExpectedJarFile(mavenCoordinate, targetDirectory, classifierInformation);
 
         if (expectedJarFile.exists()) {
-            LOGGER.info("The file " + expectedJarFile + " already exists and won't be downloaded again");
+            LOGGER.debug("The file " + expectedJarFile + " already exists and won't be downloaded again");
```
neubs-bsi: I would make this a warning

modules/maven/src/main/java/org/eclipse/sw360/antenna/maven/HttpRequester.java
bs-ondem: Why should this warn the user? For me it's a warning, if the jar could't not be downloaded or something like that. But if the jar already locally exists, everything is fine.

PR #360, opened 2019-11-22
Rework exception handling in Sw360 module

### LINKED ISSUE #358: SW360 Clean up exception and error handling

Since SW360 sometimes throws RunTimeExceptions causing Antenna to fail with an exception before being able to checking the response status. 
This causes error messages without the necessary information and sometimes causes Antenna to break even though the desired outcome in SW360 is reached (#357)

### Acceptance Criteria
- [ ] Implement a clean Exception and Response Handling
    - [ ] Exceptions should be caught
    - [ ] Object responsible for error should be mentioned in output message
- [ ] Stop Antenna from failing when individual releases cause exceptions when created or patched, instead print an error message to the logger
- [ ] No error messages should be printed to logger if they are not in fact an error (e.g. #357)

### Definition of Done
- Documentation is updated

neubs-bsi:
Fixes #358
Exception handling was reworked to catch places where due to a
runtime excpetion thrown within SW360 the Antenna build failed
without throwing an ExecutionException and without a log/excpetion
explaining the error and source properly.

Furthermore most thrown ExecutionExceptions where removed and
replaced with LOGGER inputs to avoid the Antenna build to fail.
Now an error or debug message will be logged.
Errors in the communication with SW360 should not make the Antenna
run fail, since they are not crucial for an attribution document.

The function getProject in the SW360ProjectClient and the test
testing that function are removed because they are never used.

@blaumeiser-at-bosch 

*Type of change*:  improvements | bug fixes

### How Has This Been Tested?
Example project was run with SW360 Enricher and Updater activated. 
Some errors where provoked and resulted in the wished behaviour. 

### Checklist
Must:

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
@@ -32,12 +32,12 @@ public class SW360ComponentClientAdapter {
         this.componentClient = new SW360ComponentClient(restUrl, proxySettings);
     }
 
-    public SW360Component getOrCreateComponent(SW360Component componentFromRelease, HttpHeaders header) {
+    public Optional<SW360Component> getOrCreateComponent(SW360Component componentFromRelease, HttpHeaders header) {
         if(componentFromRelease.getComponentId() != null) {
             return getComponentById(componentFromRelease.getComponentId(), header);
         }
-        return getComponentByName(componentFromRelease.getName(), header)
-                .orElseGet(() -> createComponent(componentFromRelease, header));
+        return Optional.of(getComponentByName(componentFromRelease.getName(), header)
+                .orElseGet(() -> createComponent(componentFromRelease, header)));
     }
 
     public SW360Component createComponent(SW360Component component, HttpHeaders header) {
@@ -47,7 +47,7 @@ public class SW360ComponentClientAdapter {
         return componentClient.createComponent(component, header);
     }
 
-    public SW360Component getComponentById(String componentId, HttpHeaders header) {
+    public Optional<SW360Component> getComponentById(String componentId, HttpHeaders header) {
         return componentClient.getComponent(componentId, header);
     }
 
@@ -67,7 +67,8 @@ public class SW360ComponentClientAdapter {
                 .collect(Collectors.toList());
 
         for (String componentId : componentIds) {
-            completeComponents.add(getComponentById(componentId, header));
+            getComponentById(componentId, header)
+                    .map(completeComponents::add);
         }
 
         return completeComponents.stream()
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
@@ -68,14 +68,15 @@ public class SW360ReleaseClientAdapter {
 
         if (releaseFromArtifact.getComponentId() == null) {
             final SW360Component componentFromRelease = SW360ComponentAdapterUtils.createFromRelease(releaseFromArtifact);
-            final SW360Component componentFromSW360 = sw360ComponentClientAdapter.getOrCreateComponent(componentFromRelease, header);
-            if (componentFromSW360.get_Embedded().getReleases().stream()
-                    .map(SW360SparseRelease::getVersion)
-                    .anyMatch(releaseFromArtifact.getVersion()::equals)) {
-                throw new ExecutionException("The release already exists in the found component");
-            }
-
-            releaseFromArtifact.setComponentId(componentFromSW360.getComponentId());
+            final Optional<SW360Component> componentFromSW360 = sw360ComponentClientAdapter.getOrCreateComponent(componentFromRelease, header);
+            componentFromSW360.ifPresent(cfs -> {
+                if (cfs.get_Embedded().getReleases().stream()
+                        .map(SW360SparseRelease::getVersion)
+                        .anyMatch(releaseFromArtifact.getVersion()::equals)) {
+                    throw new ExecutionException("The release already exists in the found component");
+                }
+                releaseFromArtifact.setComponentId(cfs.getComponentId());
+            });
         }
 
         final SW360Release releaseFromSW360 = releaseClient.createRelease(releaseFromArtifact, header);
@@ -88,11 +89,11 @@ public class SW360ReleaseClientAdapter {
         }
     }
 
-    public SW360Release getReleaseById(String releaseId, HttpHeaders header) {
+    public Optional<SW360Release> getReleaseById(String releaseId, HttpHeaders header) {
         return releaseClient.getRelease(releaseId, header);
     }
 
-    public SW360Release enrichSparseRelease(SW360SparseRelease sparseRelease, HttpHeaders header) {
+    public Optional<SW360Release> enrichSparseRelease(SW360SparseRelease sparseRelease, HttpHeaders header) {
         return getReleaseById(sparseRelease.getReleaseId(), header);
     }
 
@@ -106,7 +107,9 @@ public class SW360ReleaseClientAdapter {
 
     public Optional<SW360Release> getRelease(SW360Release sw360ReleaseFromArtifact, HttpHeaders header) {
         return getSparseRelease(sw360ReleaseFromArtifact, header)
-                .map(sr -> enrichSparseRelease(sr, header));
+                .map(sr -> enrichSparseRelease(sr, header))
+                .filter(Optional::isPresent)
+                .map(Optional::get);
     }
 
     public Optional<SW360SparseRelease> getReleaseByExternalIds(Map<String,String> externalIds, HttpHeaders headers) {
@@ -142,7 +145,7 @@ public class SW360ReleaseClientAdapter {
                     .findFirst()
                     .flatMap(release -> SW360HalResourceUtility.getLastIndexOfSelfLink(release.get_Links()));
             if (releaseId.isPresent()) {
-                return Optional.of(getReleaseById(releaseId.get(), header));
+                return getReleaseById(releaseId.get(), header);
             }
         }
         return Optional.empty();
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
@@ -10,11 +10,14 @@
  */
 package org.eclipse.sw360.antenna.sw360.rest;
 
+import org.apache.commons.lang3.Validate;
 import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.sw360.rest.resource.SW360HalResource;
 import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360Attachment;
 import org.eclipse.sw360.antenna.sw360.utils.RestUtils;
 import org.eclipse.sw360.antenna.util.ProxySettings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
@@ -22,12 +25,16 @@ import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Optional;
+
+import static org.eclipse.sw360.antenna.sw360.rest.SW360ClientUtils.checkRestStatus;
 
 public abstract class SW360AttachmentAwareClient<T extends SW360HalResource<?,?>> extends SW360Client {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SW360AttachmentAwareClient.class);
     private static final String ATTACHMENTS_ENDPOINT = "/attachments";
 
     public SW360AttachmentAwareClient(ProxySettings proxySettings) {
@@ -44,7 +51,8 @@ public abstract class SW360AttachmentAwareClient<T extends SW360HalResource<?,?>
 
     public T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, String kindToAttach, HttpHeaders header) {
         if (!Files.exists(fileToAttach)) {
-            throw new ExecutionException("The file=[" + fileToAttach + "], which should be attached to release, does not exist");
+            LOGGER.warn("The file=[" + fileToAttach + "], which should be attached to release, does not exist");
+            return itemToModify;
         }
 
         MultiValueMap<String, Object> multipartRequest = new LinkedMultiValueMap<>();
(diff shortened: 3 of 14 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ReleaseClient.java
```diff
@@ -80,46 +87,65 @@ public SW360Release getRelease(String releaseId, HttpHeaders header) {
...
         HttpEntity<String> httpEntity = RestUtils.convertSW360ResourceToHttpEntity(sw360Release, header);
 
-        ResponseEntity<Resource<SW360Release>> response = doRestPATCH(getEndpoint() + "/" + sw360Release.getReleaseId(), httpEntity,
-                new ParameterizedTypeReference<Resource<SW360Release>>() {});
+        ResponseEntity<Resource<SW360Release>> response;
+        try {
+            response = doRestPATCH(getEndpoint() + "/" + sw360Release.getReleaseId(), httpEntity,
+                    new ParameterizedTypeReference<Resource<SW360Release>>() {});
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            LOGGER.error("Request to update release {} failed with {}", sw360Release.getName(), e.getStatusCode());
```
blaumeiser-at-bosch: Just came to my mind, but this is valid for all other occurences as well. It is always bad to simply swallow the exception completely. Proposal, please add a second logger call with DEBUG level that logs the exception not only the status code.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360UserClient.java
```diff
@@ -37,8 +39,15 @@ public String getEndpoint() {
     }
 
     public SW360User getUserByEmail(String email, HttpHeaders header) {
-        ResponseEntity<Resource<SW360User>> response = doRestGET(getEndpoint() + "/" + email, header,
-                new ParameterizedTypeReference<Resource<SW360User>>() {});
+        ResponseEntity<Resource<SW360User>> response;
+        try {
+            response = doRestGET(getEndpoint() + "/" + email, header,
+                    new ParameterizedTypeReference<Resource<SW360User>>() {
```
blaumeiser-at-bosch: Ugly new line

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClientTest.java
```diff
@@ -227,43 +226,12 @@ public void testCreateProjectWithSuccess() {
         assertEquals(PROJECT_EMAIL_VALUE_1, project.get_Embedded().getCreatedBy().getEmail());
     }
 
-    @Test( expected = HttpClientErrorException.class )
+    @Test
```
blaumeiser-at-bosch: I wouldn't get rid of these tests. The thing is, that in the success case, the returned object is a different instance right? You could test that the returned object is really the same instance (== instead of equals) and in the success case the object is only semantically equal.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
```diff
@@ -32,12 +32,12 @@ public SW360ComponentClientAdapter(String restUrl, ProxySettings proxySettings)
...
     }
 
-    public SW360Component getOrCreateComponent(SW360Component componentFromRelease, HttpHeaders header) {
+    public Optional<SW360Component> getOrCreateComponent(SW360Component componentFromRelease, HttpHeaders header) {
         if(componentFromRelease.getComponentId() != null) {
             return getComponentById(componentFromRelease.getComponentId(), header);
         }
-        return getComponentByName(componentFromRelease.getName(), header)
-                .orElseGet(() -> createComponent(componentFromRelease, header));
+        return Optional.of(getComponentByName(componentFromRelease.getName(), header)
```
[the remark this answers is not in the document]
neubs-bsi: This function is called in the `createRelease` method. Hence, when it is called, the release the component has been created from is very unlikely to have a component id or release id mapped to it. 
Should it have an id mapped to it, we get the component by the id from sw360, if not we try to find it by name in SW360. We can't derive the Id from a release that is just being created, hence the component id would be empty and needed to be derived from seraching by name. 
If the component does not exist by name in SW360, we create the component and give that back to the release. 

That's how I understand the function. 

Though i do agree, it is a little tangled and should probably be reworked in a separate pull request.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -62,14 +67,20 @@ public T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, String kin
...
+        try {
+            response = restTemplate.postForEntity(self + ATTACHMENTS_ENDPOINT, requestEntity, getHandledClassType());
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            LOGGER.warn("Request to get attach {} to {} failed with {}", fileToAttach, self, e.getStatusCode());
+            return itemToModify;
+        }
 
         if (response.getStatusCode().is2xxSuccessful()) {
             return Optional.ofNullable(response.getBody())
                     .orElseThrow(() -> new ExecutionException("Body was null"));
```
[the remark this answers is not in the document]
neubs-bsi: `itemToModify` is the original item. Or do you mean something else?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClient.java
```diff
@@ -65,37 +71,32 @@ public String getEndpoint() {
...
-            throw new ExecutionException("Request to create project " + sw360Project.getName() + " failed with "
-                    + response.getStatusCode());
+        ResponseEntity<Resource<SW360Project>> response;
+        try {
+            response = doRestPOST(getEndpoint(), httpEntity,
+                    new ParameterizedTypeReference<Resource<SW360Project>>() {});
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            LOGGER.debug("Request to create project {} failed with {}",
+                    sw360Project.getName(), e.getStatusCode());
+            return sw360Project;
```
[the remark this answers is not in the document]
neubs-bsi: No, after the try-catch-statement, the response is checked (successful or not) and then the response body is returned

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClientTest.java
neubs-bsi: I agree with the `testCreateProjectWithBadStatusCode` test. 
the other ones I deleted, because the function they test is not existing anymore

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Yes, itemToModify

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClient.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Sorry, misread the red part

modules/sw360/src/test/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClientTest.java
blaumeiser-at-bosch: testProjectWithSuccess to test a successful run through the functionality?

blaumeiser-at-bosch: As discussed, I understand the issue now, and my comment is obsolete

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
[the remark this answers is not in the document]
neubs-bsi: Would it be okay for you, if I include the null check in the first if statement here? since the outcome is pretty much the same?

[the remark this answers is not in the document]
blaumeiser-at-bosch: Why not .orElse(itemToModify) instead of the orElseThrow? Cool would be one functional sequence of using only the Optional with a closing orElse? But not necessarily better readable.




Definitely cooler 👿

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
```diff
@@ -106,7 +107,9 @@ public SW360Release enrichSparseRelease(SW360SparseRelease sparseRelease, HttpHe
 
     public Optional<SW360Release> getRelease(SW360Release sw360ReleaseFromArtifact, HttpHeaders header) {
         return getSparseRelease(sw360ReleaseFromArtifact, header)
-                .map(sr -> enrichSparseRelease(sr, header));
+                .map(sr -> enrichSparseRelease(sr, header))
+                .filter(Optional::isPresent)
```
blaumeiser-at-bosch: I think you need flatMap here, which decreases cascaded Optionals

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -62,14 +67,24 @@ public T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, String kin
 
     private T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, HttpEntity<MultiValueMap<String, Object>> requestEntity) {
         final String self = itemToModify.get_Links().getSelf().getHref();
-        ResponseEntity<T> response = restTemplate.postForEntity(self + ATTACHMENTS_ENDPOINT, requestEntity, getHandledClassType());
+        ResponseEntity<T> response;
```
blaumeiser-at-bosch: Idea was to have something like this, to prevent this mixed imperative and functional stuff:
```suggestion
        return Optional.ofNullable({
            try {
                restTemplate.postForEntity(self + ATTACHMENTS_ENDPOINT, requestEntity, getHandledClassType());
            } catch (...) {
                return null;
            })
            .filter(res -> res.getStatusCode().is2xxSuccessful())
            .map(ResponseEntity::getBody)
            .orElseGet(...);
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ProjectClient.java
```diff
@@ -55,47 +60,48 @@ public String getEndpoint() {
...
-                new ParameterizedTypeReference<Resource<SW360Project>>() {});
-
-        if (response.getStatusCode() == HttpStatus.CREATED) {
-            return Optional.ofNullable(response.getBody())
-                    .orElseThrow(() -> new ExecutionException("Body was null"))
-                    .getContent();
-        } else {
-            throw new ExecutionException("Request to create project " + sw360Project.getName() + " failed with "
-                    + response.getStatusCode());
+        ResponseEntity<Resource<SW360Project>> response;
```
blaumeiser-at-bosch: There are many items of this pattern doSomeRequest -> checkResult, where the active code and the error handled are very much intermingled. Lets talk about a more clearer solution tomorrow, I would like to separate the error handling more from the business code.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
neubs-bsi: With this solution Logging wouldn't be able if the response is not successful.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AuthenticationClient.java
```diff
@@ -56,23 +56,22 @@ public String getOAuth2AccessToken(String username, String password, String clie
...
         HttpEntity<String> httpEntity = new HttpEntity<>(body, headers);
-        ResponseEntity<String> response;
         try {
-            response = doRestCall(getEndpoint(), HttpMethod.POST, httpEntity, String.class);
+            ResponseEntity<String> response = doRestCall(getEndpoint(), HttpMethod.POST, httpEntity, String.class);
+
+            if (response.getStatusCode() == HttpStatus.OK) {
+                try {
+                    return (String) new ObjectMapper().readValue(response.getBody(), HashMap.class).get(JSON_TOKEN_KEY);
+                } catch (IOException e) {
```
blaumeiser-at-bosch: The idea is that this catch is also moved to the big catch below?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AuthenticationClient.java
```diff
@@ -56,23 +56,22 @@ public String getOAuth2AccessToken(String username, String password, String clie
         headers.add(HttpHeaders.CONTENT_TYPE, TOKEN_CONTENT_TYPE);
 
         HttpEntity<String> httpEntity = new HttpEntity<>(body, headers);
-        ResponseEntity<String> response;
         try {
-            response = doRestCall(getEndpoint(), HttpMethod.POST, httpEntity, String.class);
+            ResponseEntity<String> response = doRestCall(getEndpoint(), HttpMethod.POST, httpEntity, String.class);
+
+            if (response.getStatusCode() == HttpStatus.OK) {
```
blaumeiser-at-bosch: HEre I propose a method checkStatus(condition) throwing an IllegalStateException which is caught in the big catch below again.

Apply this pattern to the other instances of this sequence as well.

Or even better, do not do an else branch but return the exception at the end of the method or the itemToModify

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360ClientUtils.java
```diff
@@ -46,17 +41,25 @@ private SW360ClientUtils() {
...
 
     protected static List<SW360SparseComponent> getSw360SparseComponents(ResponseEntity<Resource<SW360ComponentList>> response) {
-        SW360ComponentList resource = Optional.ofNullable(response.getBody())
-                .map(Resource::getContent)
-                .orElseGet(() -> {
-                    LOGGER.debug("Request to get sparse components returned empty body");
-                    return new SW360ComponentList();
-                });
+        checkResponse(response);
+        SW360ComponentList resource =response.getBody().getContent();
```
blaumeiser-at-bosch: a space missing

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -62,14 +69,18 @@ public T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, String kin
...
+            ResponseEntity<T> response = restTemplate.postForEntity(self + ATTACHMENTS_ENDPOINT, requestEntity, getHandledClassType());
 
-        if (response.getStatusCode().is2xxSuccessful()) {
-            return Optional.ofNullable(response.getBody())
-                    .orElseThrow(() -> new ExecutionException("Body was null"));
-        } else {
-            throw new ExecutionException("Request to get attach " + fileToAttach + " to " + self + " failed with "
-                    + response.getStatusCode());
+            checkRestStatus(response);
+            return response.getBody();
```
blaumeiser-at-bosch: We should not return null here, use the getSaveOrThrow with Identical method as second parameter

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AuthenticationClient.java
```diff
@@ -46,24 +50,26 @@ public String getEndpoint() {
...
+            HttpHeaders headers = new HttpHeaders();
+            addBasicAuthentication(headers, clientId, clientPassword);
+            headers.add(HttpHeaders.CONTENT_TYPE, TOKEN_CONTENT_TYPE);
+
+            HttpEntity<String> httpEntity = new HttpEntity<>(body, headers);
+
+            ResponseEntity<String> response = doRestCall(getEndpoint(), HttpMethod.POST, httpEntity, String.class);
+
+            checkRestStatus(response);
+            return (String) new ObjectMapper().readValue(response.getBody(), HashMap.class).get(JSON_TOKEN_KEY);
```
blaumeiser-at-bosch: What is returned here if the body is null?

neubs-bsi: readValue() has an internal null check for its content.

PR #361, opened 2019-11-22
SW360: roundtrip attachments

neubs-bsi:
Prepares to make attachments work
1. in updater with no exceptions being thrown for duplicates
2. in enricher to download source and binary files

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
@@ -14,6 +14,8 @@ import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
 import org.eclipse.sw360.antenna.sw360.rest.SW360ReleaseClient;
 import org.eclipse.sw360.antenna.sw360.rest.resource.SW360HalResourceUtility;
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360AttachmentType;
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360SparseAttachment;
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360Component;
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360ComponentEmbedded;
 import org.eclipse.sw360.antenna.sw360.rest.resource.releases.SW360Release;
@@ -82,8 +84,8 @@ public class SW360ReleaseClientAdapter {
         return releaseClient.createRelease(releaseFromArtifact, header);
     }
 
-    public SW360Release uploadAttachments(SW360Release sw360item, Map<Path, String> attachments, HttpHeaders header) {
-        for(Map.Entry<Path, String> attachment : attachments.entrySet()) {
+    public SW360Release uploadAttachments(SW360Release sw360item, Map<Path, SW360AttachmentType> attachments, HttpHeaders header) {
+        for(Map.Entry<Path, SW360AttachmentType> attachment : attachments.entrySet()) {
             sw360item = releaseClient.uploadAndAttachAttachment(sw360item, attachment.getKey(), attachment.getValue(), header);
         }
         return sw360item;
@@ -149,5 +151,14 @@ public class SW360ReleaseClientAdapter {
             }
         }
         return Optional.empty();
+    }
+
+    public List<SW360SparseAttachment> getAttachmentsOfRelease(SW360Release release, HttpHeaders header) {
+        return releaseClient.getItemAttachments(release, header);
+    }
+
+    public Optional<Path> downloadAttachment(SW360Release release, SW360SparseAttachment attachment, Path downloadPath, HttpHeaders header) {
+        return Optional.ofNullable(release.get_Links().getSelf())
+                .flatMap(self -> releaseClient.downloadAttachment(self.getHref(), attachment, downloadPath, header));
     }
 }
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
@@ -86,9 +86,16 @@ public class SW360ReleaseClientAdapter {
 
     public SW360Release uploadAttachments(SW360Release sw360item, Map<Path, SW360AttachmentType> attachments, HttpHeaders header) {
         for(Map.Entry<Path, SW360AttachmentType> attachment : attachments.entrySet()) {
-            sw360item = releaseClient.uploadAndAttachAttachment(sw360item, attachment.getKey(), attachment.getValue(), header);
+            if (!attachmentIsPotentialDuplicate(attachment.getKey(), sw360item.get_Embedded().getAttachments())) {
+                sw360item = releaseClient.uploadAndAttachAttachment(sw360item, attachment.getKey(), attachment.getValue(), header);
+            }
         }
         return sw360item;
+    }
+
+    private boolean attachmentIsPotentialDuplicate(Path attachment, List<SW360SparseAttachment> attachments) {
+        return attachments.stream()
+                .anyMatch(attachment1 -> attachment1.getFilename().equals(attachment.getFileName().toString()));
     }
 
     public Optional<SW360Release> getReleaseById(String releaseId, HttpHeaders header) {
@@ -151,10 +158,6 @@ public class SW360ReleaseClientAdapter {
             }
         }
         return Optional.empty();
-    }
-
-    public List<SW360SparseAttachment> getAttachmentsOfRelease(SW360Release release, HttpHeaders header) {
-        return releaseClient.getItemAttachments(release, header);
     }
 
     public Optional<Path> downloadAttachment(SW360Release release, SW360SparseAttachment attachment, Path downloadPath, HttpHeaders header) {
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
@@ -79,14 +79,14 @@ public class SW360ReleaseClientAdapter {
             });
         }
 
-        final SW360Release releaseFromSW360 = releaseClient.createRelease(releaseFromArtifact, header);
+        return releaseClient.createRelease(releaseFromArtifact, header);
+    }
 
-        final Optional<Path> sourceFile = releaseFromArtifact.getSourceFile();
-        if (uploadSource && sourceFile.isPresent()) {
-            return releaseClient.uploadAndAttachAttachment(releaseFromSW360, sourceFile.get(), "SOURCE", header);
-        } else {
-            return releaseFromSW360;
+    public SW360Release uploadAttachments(SW360Release sw360item, Map<Path, String> attachments, HttpHeaders header) {
+        for(Map.Entry<Path, String> attachment : attachments.entrySet()) {
+            sw360item = releaseClient.uploadAndAttachAttachment(sw360item, attachment.getKey(), attachment.getValue(), header);
         }
+        return sw360item;
     }
 
     public Optional<SW360Release> getReleaseById(String releaseId, HttpHeaders header) {
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
@@ -77,7 +77,7 @@ public abstract class SW360AttachmentAwareClient<T extends SW360HalResource<?,?>
             Validate.validState(response.getBody() != null);
             return response.getBody();
         } catch (HttpClientErrorException | HttpServerErrorException e) {
-            LOGGER.warn("Request to get attach {} to {} failed with {}", fileToAttach, self, e.getStatusCode());
+            LOGGER.warn("Request to attach {} to {} failed with {}", fileToAttach, self, e.getStatusCode());
             LOGGER.debug("Error: ", e);
             return itemToModify;
         } catch (ExecutionException e) {
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
@@ -14,24 +14,30 @@ import org.apache.commons.lang3.Validate;
 import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.sw360.rest.resource.SW360HalResource;
 import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360Attachment;
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360AttachmentList;
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360AttachmentType;
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360SparseAttachment;
 import org.eclipse.sw360.antenna.sw360.utils.RestUtils;
 import org.eclipse.sw360.antenna.util.ProxySettings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.core.io.FileSystemResource;
-import org.springframework.http.HttpEntity;
-import org.springframework.http.HttpHeaders;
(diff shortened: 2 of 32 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
```diff
@@ -80,7 +85,11 @@ public SW360Release createRelease(SW360Release releaseFromArtifact, boolean uplo
 
         final SW360Release releaseFromSW360 = releaseClient.createRelease(releaseFromArtifact, header);
 
-        final Optional<Path> sourceFile = releaseFromArtifact.getSourceFile();
+        return checkAndMakeAttachment(releaseFromArtifact, releaseFromSW360, uploadSource, header);
+    }
+
+    private SW360Release checkAndMakeAttachment(SW360Release release, SW360Release releaseFromSW360, boolean uploadSource, HttpHeaders header) {
```
blaumeiser-at-bosch: Naming is not nice, it is only the source attachment, not any attachment.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -60,6 +71,28 @@ public T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, String kin
         return uploadAndAttachAttachment(itemToModify, fileToAttach, requestEntity);
     }
 
+    protected boolean isAttachmentDuplicate(T itemToModify, SW360Attachment attachment, HttpHeaders header) {
```
blaumeiser-at-bosch: If we only compare by filename, this is not enough, default is that we overwrite the existing attachment instead of not uploading it.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ReleaseAdapterUtils.java
```diff
@@ -134,4 +134,9 @@ private static void setCopyrights(SW360Release release, Artifact artifact) {
         Optional<CopyrightStatement> cs = artifact.askFor(CopyrightStatement.class);
         cs.ifPresent(copyrightStatement -> release.setCopyrights(copyrightStatement.toString()));
     }
+
+    public static void setSourceFile(SW360Release release, Artifact artifact) {
+        Optional<ArtifactSourceFile> asf = artifact.askFor(ArtifactSourceFile.class);
```
blaumeiser-at-bosch: asf???

neubs-bsi:
> I miss the get part of the roundtrip for the enricher? Is this not part of this commit? Is this already existing?

This is still a draft.
The attachment functionality is not in the enricher yet. 
I wanted to get the exception PR done first, since this PR depends on it.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -17,10 +17,17 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.Map;
 
 public class SW360Enricher extends AbstractProcessor {
+    private static final String DOWNLOAD_ATTACHMENTS = "download.attachments";
+    private static final String DOWNLOAD_ATTACHMENTS_DIR="download.directory";
+    private boolean downloadAttachments;
+    private Path downloadPath = Paths.get("");
```
bs-ondem: Probably we could make the default downloadPath to antenna target directory?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -39,6 +46,11 @@ public void configure(Map<String, String> configMap) {
         String sw360ProxyHost = context.getToolConfiguration().getProxyHost();
         int sw360ProxyPort = context.getToolConfiguration().getProxyPort();
 
+        downloadAttachments = getBooleanConfigValue(DOWNLOAD_ATTACHMENTS, configMap);
+        if (downloadAttachments) {
+            downloadPath = Paths.get(getConfigValue(DOWNLOAD_ATTACHMENTS_DIR, configMap));
```
bs-ondem: The user has to specify always the absolute path? Is this documented somewhere?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -77,12 +82,54 @@ private T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, HttpEntit
...
+            ResponseEntity<byte[]> response = restTemplate.exchange(url, HttpMethod.GET, httpEntity, byte[].class);
+            checkRestStatus(response);
+            Validate.validState(response.getBody() != null);
+
+            Optional<byte[]> body = Optional.of(response.getBody());
+
+            return Optional.of(Files.write(downloadPath.resolve(attachment.getFilename()), body.get()));
+        } catch (HttpClientErrorException e) {
+
+            LOGGER.debug("Request to get attachment {} from {} failed with {}", attachmentId, itemHref, e.getStatusCode());
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: I would like to add, that the general handling here and in the method above is a bit weird. Not being able to provide the attachment is in general a problem because we will not be able to generate things, e.g., the source bundle properly. All of these exceptions should result in a warning at least. We should also always log the exception stack trace as debug information, that will help us if we encounter the issue in a support call (see below).

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
```diff
@@ -77,12 +82,54 @@ private T uploadAndAttachAttachment(T itemToModify, Path fileToAttach, HttpEntit
...
+            LOGGER.warn("Request to attach {} to {} failed with {}", fileToAttach, self, e.getStatusCode());
             LOGGER.debug("Error: ", e);
             return itemToModify;
         } catch (ExecutionException e) {
             LOGGER.warn("Request to attach {} to {} failed with {}", fileToAttach, self, e.getMessage());
             return itemToModify;
         }
     }
+
+    public List<SW360SparseAttachment> getReleaseAttachments(T itemToModify, HttpHeaders header) {
```
blaumeiser-at-bosch: I am just thinking whether this method makes sense at all. There might be a multitude of attachments and from an Antenna point of view, we should be interested in some of them. But typically we only need to download the sources nothing else. Or is it that the method only returns the references to the attachments as a list and I have to use the download method below for the real download?

By the way, does it make sense to make a _getReleaseAttachments_ method and then use a parameter T that allows to download any kind of attachment?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
```diff
@@ -150,4 +151,13 @@ public SW360Release createRelease(SW360Release releaseFromArtifact, boolean uplo
         }
         return Optional.empty();
     }
+
+    public List<SW360SparseAttachment> getAttachmentsOfRelease(SW360Release release, HttpHeaders header) {
```
blaumeiser-at-bosch: For my understanding, this only returns the list of available attachments, right? Or does it also download the attachments, if so, my comment from below is valid here as well.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/attachments/SW360SparseAttachment.java
```diff
@@ -0,0 +1,76 @@
...
+    public String getFilename() {
+        return filename;
+    }
+
+    public SW360SparseAttachment setFilename(String filename) {
+        this.filename = filename;
+        return this;
+    }
+
+    public String getAttachmentType() {
```
blaumeiser-at-bosch: Is attachment type potentially an enum?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360AttachmentAdapterUtils.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class SW360AttachmentAdapterUtils {
+
+    public static Map<Path, String> getAttachmentsFromArtifact(Artifact artifact) {
+        Map<Path, String> attachments = new HashMap<>();
+
+        getSourceFile(artifact).map(sourceFile -> attachments.put(sourceFile, "SOURCE"));
```
blaumeiser-at-bosch: See above comment: An enum to be used here?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
```diff
@@ -54,13 +60,29 @@ public SW360EnricherImpl(IProcessingReporter reporter, SW360MetaDataReceiver con
...
+                    downloadAttachments(sw360Release, artifact);
+                }
             } else {
                 warnAndReport(artifact, "No SW360 release found for artifact.", MessageType.PROCESSING_FAILURE);
             }
         }
         return intermediates;
     }
 
+    private void downloadAttachments(SW360Release sw360Release, Artifact artifact) {
```
blaumeiser-at-bosch: Please do not do this, download only attachments needed for the antenna process

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
neubs-bsi: It only gets references to the attachment. For a download another call must be made

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/attachments/SW360SparseAttachment.java
neubs-bsi: Yes it is

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherImpl.java
neubs-bsi: I did this with the following thought:
- we want to eventually have attachments for the source, the binary and the clearing state. this totals to 3 kind of attachments we can map on the artifacts. 
- if you check the content of the method it shows that per attachment I call a function `mapAttachmentOnArtifact()`. This checks if the attachment has a type that is something we can map on an artifact and *only then* does it actually download and map the attachment. this is currently just "SOURCE" since the binary and clearing state part is still missing. 
- I can rename the method to something like  `checkAndDownloadAttachments` or something. 

Conclusion: I am not downloading all attachments

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/SW360AttachmentAwareClient.java
neubs-bsi: The parameter T is for the HalResource. One gets an attachment of any object type (release, component or project).
Though, you are right. It is not named accordingly, leading to confusion. I'll check this again

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
neubs-bsi: This would result in an NPE, since at the time of instantiation the context is null for the enricher

bs-ondem: Yes, you can do it in `configure`.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ReleaseClientAdapter.java
```diff
@@ -150,4 +153,18 @@ public SW360Release createRelease(SW360Release releaseFromArtifact, boolean uplo
...
+        return releaseClient.getItemAttachments(release, header);
+    }
+
+    public Optional<Path> downloadAttachment(SW360Release release, SW360SparseAttachment attachment, Path downloadPath, HttpHeaders header) {
+        Self self = release.get_Links().getSelf();
```
blaumeiser-at-bosch: Nicer would be
`return Optional.of(release).map(...get_Links()).map(...getSelf()).map(...getHref()).map(...releaseClient...);`

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/attachments/SW360AttachmentType.java
```diff
@@ -0,0 +1,85 @@
...
+            case 14:
+                return LICENSE_AGREEMENT;
+            case 15:
+                return SCREENSHOT;
+            case 16:
+                return OTHER;
+            case 17:
+                return README_OSS;
+            default:
+                return null;
```
blaumeiser-at-bosch: Potentially an exceptional case, returning null is something that should not be done, because it can lead into NPEs in other places. Alternatively there is a default, like "OTHER"?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/attachments/SW360SparseAttachment.java
```diff
@@ -0,0 +1,76 @@
...
+        this.attachmentType = attachmentType;
+        return this;
+    }
+
+    @JsonIgnore
+    public String getAttachmentId() {
+        return Optional.ofNullable(get_Links())
+                .map(LinkObjects::getSelf)
+                .flatMap(SW360HalResourceUtility::getLastIndexOfSelfLink)
+                .orElse(null);
```
blaumeiser-at-bosch: null as return value always make me cringe, can't we use "" as return value?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360AttachmentAdapterUtils.java
```diff
@@ -0,0 +1,37 @@
...
+        Map<Path, SW360AttachmentType> attachments = new HashMap<>();
+
+        getSourceFile(artifact).map(sourceFile -> attachments.put(sourceFile, SW360AttachmentType.SOURCE));
+
+        return attachments;
+    }
+
+    private static Optional<Path> getSourceFile(Artifact artifact) {
+        Optional<ArtifactSourceFile> artifactSourceFile = artifact.askFor(ArtifactSourceFile.class);
+        return artifactSourceFile.map(ArtifactFactWithPayload::get);
```
bs-ondem: Just use `askForGet`
```suggestion
        return artifact.askForGet(ArtifactSourceFile.class);
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360AttachmentAdapterUtils.java
```diff
@@ -0,0 +1,37 @@
...
+import org.eclipse.sw360.antenna.sw360.rest.resource.attachments.SW360AttachmentType;
+
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class SW360AttachmentAdapterUtils {
+
+    public static Map<Path, SW360AttachmentType> getAttachmentsFromArtifact(Artifact artifact) {
```
bs-ondem: This method will always return a Map with only one entry, correct?
You can use one liner
```suggestion
    public static Map<Path, SW360AttachmentType> getAttachmentsFromArtifact(Artifact artifact) {
        return getSourceFile(artifact)
                            .map(sourceFile -> Collections.singletonMap(sourceFile, SW360AttachmentType.SOURCE)
                            .orElse(Collections.EMPTY_MAP);
    }
```

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -39,6 +46,13 @@ public void configure(Map<String, String> configMap) {
         String sw360ProxyHost = context.getToolConfiguration().getProxyHost();
         int sw360ProxyPort = context.getToolConfiguration().getProxyPort();
 
+        downloadAttachments = getBooleanConfigValue(DOWNLOAD_ATTACHMENTS, configMap);
+        if (downloadAttachments) {
+            downloadPath = Paths.get(getConfigValue(DOWNLOAD_ATTACHMENTS_DIR, configMap,
+                    context.getToolConfiguration().getAntennaTargetDirectory().toString()));
+            downloadPath.toAbsolutePath();
```
bs-ondem: Append it to Paths.get? And probably you should normalize the path.

Issue #326, opened 2019-10-16
Evaluate ORT Downloader usage within Antenna workflow


blaumeiser-at-bosch:
### Summary of the Feature
Idea is that for components which do not have local source information the ORT downloader is used to get the sources and to process them further, e.g., upload to SW360, store them at a local place, ...

### Acceptance Criteria
- [ ] In case of a dependency with no local sources, the sources are downloaded via ORT downloader
- [ ] The information is stored in the antenna model appropriately

### Definition of Done
- [ ] Acceptance criteria fulfilled
- [ ] A PR is created, the CI infrastructure reports green
- [ ] The PR is reviewed and approved
- [ ] No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- [ ] Test cases are created to prove the functionality of the feature
- [ ] Documentation is updated

PR #363, opened 2019-11-27
integrate ORT Downloader as a processor step

bs-ondem:
Resolves https://github.com/eclipse/antenna/issues/326

This PR implements a new processor step, in which the ORT Downloader is programmatically called. The ORT Downloader downloads the sources of the artifacts in a temporary directory if it has no local sources. Afterwards it will be archived to a ZIP file in the hardcoded directory `<AntennaTargetDirectory>/ort-downloader-result` and saved in the artifact model. 

New Dependencies: 
- com.github.heremaps.oss-review-toolkit:downloader:f91202964b

### Request Reviewer
@sschuberth @blaumeiser-at-bosch @neubs-bsi 

### Type of Change
new feature | documentation update

### How Has This Been Tested?
Run `example-projects/example-project` with `mvn clean package` and check the `ort-downloader-result` directory and the `sources.zip` file, which now contains also the sources of the JS components. 

### Checklist

### DIFF
--- a/antenna-documentation/src/site/markdown/processors/ort-downloader.md
+++ b/antenna-documentation/src/site/markdown/processors/ort-downloader.md
@@ -1,0 +1,12 @@
+## ORT Downloader 
+The ORT Downloader processor downloads the sources of each artifact if no `ArtifactSourceFile` fact is present.
+
+### HowTo Use
+Add the following step into the `<processors>` section of your workflow.xml
+
+```xml
+<step>
+    <name>ORT Downloader</name>
+    <classHint>org.eclipse.sw360.antenna.workflow.processors.enricher.OrtDownloaderProcessor</classHint>
+</step>
+```
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactVcsInfo.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactVcsInfo.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.artifact.facts;
+
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class ArtifactVcsInfo implements ArtifactFact<ArtifactVcsInfo> {
+    private final VcsInfo vcsInfo;
+
+    public static class VcsInfo {
+        private final String type;
+        private final String url;
+        private final String revision;
+
+        public VcsInfo(String type, String url, String revision) {
+            this.type = type;
+            this.url = url;
+            this.revision = revision;
+        }
+
+        public String getType() {
+            return type;
+        }
+
+        public String getUrl() {
+            return url;
+        }
+
+        public String getRevision() {
+            return revision;
+        }
+
+        public boolean isEmpty() {
+            return type == null && url == null && revision == null;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            VcsInfo vcsInfo = (VcsInfo) o;
+            return Objects.equals(type, vcsInfo.type) &&
+                    Objects.equals(url, vcsInfo.url) &&
+                    Objects.equals(revision, vcsInfo.revision);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(type, url, revision);
+        }
+
+        @Override
+        public String toString() {
+            return "VcsInfo{" +
+                    "type='" + type + '\'' +
+                    ", url='" + url + '\'' +
+                    ", revision='" + revision + '\'' +
+                    '}';
+        }
+    }
+
+    public ArtifactVcsInfo(String type, String url, String revision) {
+        this(new VcsInfo(type, url, revision));
+    }
+
+    public ArtifactVcsInfo(VcsInfo vcsInfo) {
+        this.vcsInfo = vcsInfo;
+    }
+
+    public VcsInfo getVcsInfo() {
+        return vcsInfo;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ArtifactVcsInfo that = (ArtifactVcsInfo) o;
+        return Objects.equals(vcsInfo, that.vcsInfo);
+    }
+
+    @Override
+    public String toString() {
+        return Optional.ofNullable(vcsInfo)
+                .map(Objects::toString)
+                .orElse("EMPTY");
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(vcsInfo);
+    }
+
+    @Override
+    public String getFactContentName() {
+        return "VCS Information";
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return vcsInfo == null;
+    }
+
+    @Override
+    public String prettyPrint() {
+        return "Set ArtifactVcsInfo to " + toString();
+    }
+}
--- a/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
+++ b/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
@@ -47,6 +47,11 @@ private fun mapSourceUrl(pkg: Package): ArtifactSourceUrl? =
         ArtifactSourceUrl(it)
     }
 
+private fun mapVcsInfo(pkg: Package) : ArtifactVcsInfo? =
+        pkg.vcs.takeUnless { it == VcsInfo.EMPTY }?.let {
+            ArtifactVcsInfo(it.type.toString(), it.url, it.revision)
+        }
+
 private fun mapDeclaredLicense(pkg: Package): DeclaredLicenseInformation? =
     pkg.declaredLicensesProcessed.allLicenses.takeUnless { it.isEmpty() }?.let {
         DeclaredLicenseInformation(LicenseSupport.mapLicenses(it))
@@ -86,6 +91,7 @@ class OrtResultArtifactResolver(result: OrtResult) : Function<Package, Artifact>
             a.addCoordinate(mapCoordinates(pkg))
 
             mapSourceUrl(pkg)?.let { a.addFact(it) }
+            mapVcsInfo(pkg)?.let { a.addFact(it) }
             mapDeclaredLicense(pkg)?.let { a.addFact(it) }
             mapFilename(pkg)?.let { a.addFact(it) }
             mapHomepage(pkg)?.let { a.addFact(it) }
--- a/modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
+++ b/modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2019.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.ort.utils
+
+import com.here.ort.model.*
+import org.eclipse.sw360.antenna.model.artifact.Artifact
+import org.eclipse.sw360.antenna.model.artifact.ArtifactCoordinates
+import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactSourceUrl
+import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactVcsInfo
+import java.util.function.Function
+
+const val MAVEN_TYPE_KEY = "Maven"
+const val NPM_TYPE_KEY = "NPM"
+const val NUGET_TYPE_KEY = "nuget"
+const val UNMANAGED_TYPE_KEY = "Unmanaged"
+
+class ArtifactToPackageMapper : Function<Artifact, Package> {
+    override fun apply(artifact: Artifact): Package {
+        return Package(
+                id = mapIdentifier(artifact) ?: Identifier.EMPTY,
+                declaredLicenses = sortedSetOf(),
+                description = "",
+                homepageUrl = "",
+                binaryArtifact = RemoteArtifact.EMPTY,
+                sourceArtifact = mapSourceRemoteArtifact(artifact) ?: RemoteArtifact.EMPTY,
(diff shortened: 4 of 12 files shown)


antenna-documentation/src/site/markdown/processors/ort-downloader.md
```diff
@@ -0,0 +1,12 @@
+## ORT Downloader 
+The ORT Downloader processor downloads the sources of each artifact if no `ArtifactSourceFile` fact is present.
```
sschuberth: Just for my understanding: I assume that *if* `ArtifactSourceFile` is present, it points to a local existing file, correct? I.e., it does *not* point to a remote file location where the source code is hosted. Othwiser the sentence would not make sense to me, as I would expect the ORT Download to download remote files.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactVcsInfo.java
```diff
@@ -0,0 +1,119 @@
...
+
+import org.eclipse.sw360.antenna.model.artifact.ArtifactFact;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class ArtifactVcsInfo implements ArtifactFact<ArtifactVcsInfo> {
+    private final VcsInfo vcsInfo;
+
+    public static class VcsInfo {
```
sschuberth: Could we simply directly use ORT's VcsInfo class here instead?

modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
```diff
@@ -47,6 +47,11 @@ private fun mapSourceUrl(pkg: Package): ArtifactSourceUrl? =
         ArtifactSourceUrl(it)
     }
 
+private fun mapVcsInfo(pkg: Package) : ArtifactVcsInfo? =
+        pkg.vcs.takeUnless { it?.equals(VcsInfo.EMPTY) }?.let {
```
sschuberth: Could can simplify `it?` to just `it` here as the result is non-nullable.

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+                homepageUrl = "",
+                binaryArtifact = RemoteArtifact.EMPTY,
+                sourceArtifact = mapSourceRemoteArtifact(artifact) ?: RemoteArtifact.EMPTY,
+                vcs = mapVcsInfo(artifact) ?: VcsInfo.EMPTY
+                )
+    }
+
+    private fun mapIdentifier(artifact: Artifact) : Identifier? =
+        artifact.askFor(ArtifactCoordinates::class.java).takeIf { it.isPresent }?.let {
+            var coordinates = it.get()
```
sschuberth: This could be a `val`.

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+            val type = coordinates.mainCoordinate.type.toLowerCase()
+            val namespace = coordinates.mainCoordinate.namespace ?: ""
+            val name = coordinates.mainCoordinate.name ?: ""
+            val version = coordinates.mainCoordinate.version ?: ""
+            when (type) {
```
sschuberth: I'd propose to only map (Antenna) type to (ORT) type here for clarity, and construct the `Identifier` afterwards. Like:

            val ortType = when (type) {
                "maven" -> MAVEN_TYPE_KEY
                "npm" -> NPM_TYPE_KEY
                "nuget" -> NUGET_TYPE_KEY
                else -> UNMANAGED_TYPE_KEY
            }
            return Identifier(type = ortType, namespace = namespace, name = name, version = version)

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+                "nuget" -> Identifier(type = NUGET_TYPE_KEY, namespace = namespace, name = name, version = version)
+                else -> {
+                    Identifier(type = UNMANAGED_TYPE_KEY, namespace = namespace, name = name, version = version)
+                }
+            }
+        }
+
+    private fun mapVcsInfo(artifact: Artifact) : VcsInfo? =
+        artifact.askFor(ArtifactVcsInfo::class.java).takeIf { it.isPresent }?.let {
+            var vcsInfo = it.get().vcsInfo
```
sschuberth: This could be a `val`.

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+            VcsInfo(type = VcsType(vcsInfo.type), url = vcsInfo.url, revision = vcsInfo.revision)
+        }
+
+    private fun mapSourceRemoteArtifact(artifact: Artifact) : RemoteArtifact? =
+        artifact.askFor(ArtifactSourceUrl::class.java).takeIf { it.isPresent }?.let {
+            RemoteArtifact.EMPTY.copy(url = it.get().get())
+        }
+
+
+}
```
sschuberth: Nit: Missing newline at end of file.

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    private fun mapVcsInfo(artifact: Artifact) : VcsInfo? =
+        artifact.askFor(ArtifactVcsInfo::class.java).takeIf { it.isPresent }?.let {
+            var vcsInfo = it.get().vcsInfo
+            VcsInfo(type = VcsType(vcsInfo.type), url = vcsInfo.url, revision = vcsInfo.revision)
+        }
+
+    private fun mapSourceRemoteArtifact(artifact: Artifact) : RemoteArtifact? =
+        artifact.askFor(ArtifactSourceUrl::class.java).takeIf { it.isPresent }?.let {
+            RemoteArtifact.EMPTY.copy(url = it.get().get())
```
sschuberth: Is the second `.get()` semantically the right thing to call here? Or should it better be `.toString()` to clarify we expect to get a string representation here?

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+    companion object {
+        const val ORT_DOWNLOADER_DIR = "ort-downloader-result"
+        private val LOGGER = LoggerFactory.getLogger(OrtDownloaderProcessor::class.java)
+    }
+
+    init {
+        workflowStepOrder = 350
+    }
+
+    private lateinit var ortDownloadFolder: File
```
sschuberth: I believe the variable name is misleading as this is not the directory ORT downloads to, but the directory you ZIP the sources to after ORT has downloaded them. Maybe better: `sourcesZipFolder`. As a side note, I personally prefer the term "directory" over "folder" as the latter is mostly used in the Windows world whereas the former is more neutral.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override fun configure(configMap: MutableMap<String, String>) {
+        ortDownloadFolder = context.toolConfiguration.antennaTargetDirectory.resolve(ORT_DOWNLOADER_DIR).toFile()
+        ortDownloadFolder.takeUnless { it.exists() }?.let { it.mkdirs() }
```
sschuberth: `exists()` is too unspecific here, it should better be `isDirectory`. Also, I find the logic a bit hard to follow as you're not using the non-null return value of `takeUnless` at all. How about simply:

    if (!ortDownloadFolder.isDirectory) ortDownloadFolder.mkdirs()

Or, yet better as you check also for an error (e.g. is already a file of the name name exists):

    if (!ortDownloadFolder.isDirectory && !ortDownloadFolder.mkdirs()) throw ...

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+            it.sourceFile.isPresent
+        }.forEach {
+            val pkg = ArtifactToPackageMapper().apply(it)
+            val name = it.askFor(ArtifactCoordinates::class.java).takeIf { f -> f.isPresent }
+                    ?.let {
+                        it.get().mainCoordinate.canonicalize()
+                    } ?: it.askFor(ArtifactFilename::class.java).takeIf { f -> f.isPresent }
+                    ?.let { f -> f.get().bestFilenameEntryGuess.takeIf { fe -> fe.isPresent }
+                            ?.let { e -> e.get().filename }
+                    } ?: "unknown"
```
sschuberth: Here's a slightly reformatted variant of the above which also has a few redundant `let`s removed (simply apply the IDE inspections):

            val name = artifact.askFor(ArtifactCoordinates::class.java).takeIf {
                it.isPresent
            }?.get()?.mainCoordinate?.canonicalize() ?: artifact.askFor(ArtifactFilename::class.java).takeIf {
                it.isPresent
            }?.let { filename ->
                filename.get().bestFilenameEntryGuess.takeIf {
                    it.isPresent
                }?.get()?.filename
            } ?: "unknown"

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+                    val zipFile = File(
+                            ortDownloadFolder,
+                            "${pkg.id.name.encodeOrUnknown()}-${pkg.id.version.encodeOrUnknown()}.zip"
+                    )
+                    LOGGER.debug("Pack source directory '${downloadResult.downloadDirectory.absolutePath}' to " +
+                            "'${zipFile.absolutePath}.")
+                    downloadResult.downloadDirectory.packZip(zipFile)
+
+                    if (zipFile.exists()) {
+                        it.addFact(ArtifactSourceFile(zipFile.toPath()))
```
sschuberth: `it` here also needs to be renamed to `artifact` if you apply my code above.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+            val name = it.askFor(ArtifactCoordinates::class.java).takeIf { f -> f.isPresent }
+                    ?.let {
+                        it.get().mainCoordinate.canonicalize()
+                    } ?: it.askFor(ArtifactFilename::class.java).takeIf { f -> f.isPresent }
+                    ?.let { f -> f.get().bestFilenameEntryGuess.takeIf { fe -> fe.isPresent }
+                            ?.let { e -> e.get().filename }
+                    } ?: "unknown"
+            try {
+                LOGGER.debug("Download sources via ORT Downloader for '${name}'")
+                val downloadResult = Downloader().download(pkg, createTempDir("ortDownloaderDirectory"))
```
sschuberth: It's probably good practice to explicitly delete temporary directories again after use, or at least when the JVM shuts down, e.g. by using

                val ortDownloadDirectory = createTempDir("ortDownloaderDirectory").also { it.deleteOnExit() }
                val downloadResult = Downloader().download(pkg, ortDownloadDirectory)

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+                if (downloadResult.downloadDirectory.exists()) {
```
sschuberth: Better again: `isDirectory`.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+                    val zipFile = File(
+                            ortDownloadFolder,
+                            "${pkg.id.name.encodeOrUnknown()}-${pkg.id.version.encodeOrUnknown()}.zip"
+                    )
+                    LOGGER.debug("Pack source directory '${downloadResult.downloadDirectory.absolutePath}' to " +
+                            "'${zipFile.absolutePath}.")
+                    downloadResult.downloadDirectory.packZip(zipFile)
```
sschuberth: Note: This zips *all* sources of *all* components into a *single* ZIP file. I.e. you're not creating a ZIP per component. But I guess that is what you want, right?

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -0,0 +1,77 @@
...
+
+                    if (zipFile.exists()) {
```
sschuberth: Better again: `isFile`.

antenna-documentation/src/site/markdown/processors/ort-downloader.md
bs-ondem: Correct `ArtifactSourceFile` points to a local existing file.d

modules/ort/src/main/kotlin/utils/ArtifactToPackageMapper.kt
bs-ondem: The semantically correct usage is to call `askForGet` instead of `askFor`:
```
artifact.askForGet(ArtifactSourceUrl::class.java).takeIf { it.isPresent }?.let {
            RemoteArtifact.EMPTY.copy(url = it.get())
```
I will change it.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
bs-ondem: No I'm creating a ZIP per component, because `downloadResult.downloadDirectory` points to the  downloaded sources directory of the package.

core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactVcsInfo.java
bs-ondem: If we use the ORT's VcsInfo class then the antenna-model has a direct dependency to ORT's model.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -33,39 +34,45 @@ class OrtDownloaderProcessor : AbstractProcessor() {
...
+                            sourcesZipDirectory,
                             "${pkg.id.name.encodeOrUnknown()}-${pkg.id.version.encodeOrUnknown()}.zip"
                     )
                     LOGGER.debug("Pack source directory '${downloadResult.downloadDirectory.absolutePath}' to " +
                             "'${zipFile.absolutePath}.")
                     downloadResult.downloadDirectory.packZip(zipFile)
 
-                    if (zipFile.exists()) {
-                        it.addFact(ArtifactSourceFile(zipFile.toPath()))
+                    if (zipFile.exists() && zipFile.isFile) {
```
sschuberth: `isFile` (and also `isDirectory`, BTW) imply `exists()`, so only either one is needed.

sschuberth:
@bs-ondem, code-wise this looks good to me now, but Git-history-wise I'd expect you'd squash the fixup commits into the original commit before merging.

PR #370, opened 2019-12-05
Improved tests for MavenArtifactResolverImpl

bs-ondem:
The MavenArtifactResolverImplTest used real method calls of the
interface IArtifactRequester, which caused errors in a proxy
environment. The new setup mocks the IArtifactRequester interface and
tests only the behaviour of the resolver.

Furthermore MavenRuntimeRequesterTest class added to test the behaviour
of the requester.

### Request Reviewer

### Type of Change
test improvement

### How Has This Been Tested?
- Setup a proxy environment 
- Build project with `mvn clean install`


modules/maven/src/test/java/org/eclipse/sw360/antenna/maven/MavenRuntimeRequesterTest.java
```diff
@@ -0,0 +1,105 @@
...
+                .when(repositorySystem)
+                .resolve(any());
+
+        Path targetDirectory = temporaryFolder.newFolder("dependencies").toPath();
+        Optional<File> file = mavenRuntimeRequester.requestFile(coordinate, targetDirectory, ClassifierInformation.DEFAULT_JAR);
+
+        assertThat(file).isPresent();
+        Arrays.asList(targetDirectory.toFile().listFiles())
+                .stream()
+                .anyMatch(f -> f.equals(file.get()));
```
neubs-bsi: You could just do `Arrays.stream()`
```
Arrays.stream(targetDirectory.toFile().listFiles())
                .anyMatch(f -> f.equals(file.get()));
```

neubs-bsi: also, the results of this are never checked. What if the isn't any match?

modules/maven/src/test/java/org/eclipse/sw360/antenna/maven/MavenRuntimeRequesterTest.java
```diff
@@ -0,0 +1,105 @@
...
+        doReturn(stubbedMavenArtifact)
+                .when(repositorySystem)
+                .createArtifact(anyString(), anyString(), anyString(), anyString());
+
+        doReturn(result)
+                .when(repositorySystem)
+                .resolve(any());
+
+        Path targetDirectory = temporaryFolder.newFolder("dependencies").toPath();
+        Optional<File> file = mavenRuntimeRequester.requestFile(coordinate, targetDirectory, ClassifierInformation.DEFAULT_JAR);
```
neubs-bsi: This could be extracted into an extra method, since both tests use this.

bs-ondem: I would like to keep this here to keep the readability of the test.

PR #372, opened 2019-12-09
Restructure serialization of SW360HalResources

neubs-bsi:

This uses the object mapper from the jackson databind library to
serialize and deserialize all SW360Objects except the attachments
(they are not compatible with the format yet and this should be done
when the roundtrip pr is merged).
This is adapted to the way the releases already were serialized and
helps make the code cleaner by avoiding set attribute variables in
a separate file and instead have the serialization with the objects.
Should the components or projects get externalIds those will also
be easier to map and test.

Tests have been added for the serialization of the objects.

The SW360Project class has been modified to include a clearing state.
SW360 Licenses are not serialized with the generic function yet because
of uncertainty with the full name short name situation.

### Request Reviewer
@blaumeiser-at-bosch .

*Type of change*:  Improvements

Several unit tests have been added to test the serialization changes.

### Checklist
Must:
    no related issue

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360Attributes.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360Attributes.java
@@ -20,50 +20,14 @@ public class SW360Attributes {
     public static final String PROJECT_RELEASES_TRANSITIVE ="transitive";
     public static final String COMPONENT_SEARCH_BY_NAME = "name";
 
-
-    // Attributes of Sw360Project
-    public static final String PROJECT_ID = "id";
-    public static final String PROJECT_TYPE = "type";
-    public static final String PROJECT_NAME = "name";
-    public static final String PROJECT_VERSION = "version";
-    public static final String PROJECT_PROJECT_TYPE = "projectType";
-    public static final String PROJECT_DESCRIPTION = "description";
-    public static final String PROJECT_EXTERNAL_IDS = "externalIds";
-    public static final String PROJECT_CREATED_ON= "createdOn";
-    public static final String PROJECT_BUSINESS_UNIT = "businessUnit";
-    public static final String PROJECT_CLEARING_TEAM= "clearingTeam";
-    public static final String PROJECT_VISIBILITY = "visbility";
-    public static final String PROJECT_RELEASE_ID_TO_USAGE = "releaseIdToUsage";
-
-    // Attributes of Sw360Component
-    public static final String COMPONENT_ID = "id";
-    public static final String COMPONENT_COMPONENT_NAME = "name";
-    public static final String COMPONENT_COMPONENT_TYPE = "componentType";
-    public static final String COMPONENT_TYPE = "type";
-    public static final String COMPONENT_CREATED_ON = "createdOn";
-    public static final String COMPONENT_CATEGORIES = "categories";
-    public static final String COMPONENT_HOMEPAGE = "homepage";
-
-    // Attributes of Sw360Release
-    public static final String RELEASE_ID = "id";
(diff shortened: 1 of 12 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360Attributes.java
```diff
@@ -20,50 +20,14 @@ private SW360Attributes() {}
...
-    public static final String RELEASE_MAIN_LICENSE_IDS = "mainLicenseIds";
-    public static final String RELEASE_EXTERNAL_IDS = "externalIds";
-
     // Attributes of Sw360Attachment
     public static final String ATTACHMENT_ATTACHMENT_TYPE = "attachmentType";
     public static final String ATTACHMENT_CHECK_STATUS = "checkStatus";
 
     // Attributes of Sw360License
-    public static final String LICENSE_TYPE = "type";
     public static final String LICENSE_TEXT = "text";
```
bs-ondem: Why do you keep still these license attributes?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/RestUtils.java
```diff
@@ -16,19 +16,14 @@
...
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
 
-public class RestUtils {
+public class RestUtils  {
```
bs-ondem: remove whitespace

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ProjectAdapterUtils.java
```diff
@@ -36,10 +37,11 @@ public static void setVersion(SW360Project project, String version) {
...
 
-    public static void setClearingTeam(SW360Project project, String clearingTeam) {
-        if (clearingTeam != null &&
-                !clearingTeam.isEmpty()) {
-            project.setClearingTeam(clearingTeam);
+    public static void setClearingState(SW360Project project, String clearingState) {
+        if (clearingState != null &&
+                !clearingState.isEmpty()) {
+            project.setClearingState(SW360ProjectClearingState.findByValue(0));
+            // TODO implement properly -> which state when?
```
bs-ondem: Is there an issue for the TODO?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ProjectAdapterUtils.java
```diff
@@ -49,6 +51,7 @@ public static void prepareProject(SW360Project sw360Project, String projectName,
         SW360ProjectAdapterUtils.setDescription(sw360Project, projectName + " " + projectVersion);
         sw360Project.setProjectType(SW360ProjectType.PRODUCT);
         sw360Project.setVisibility(SW360Visibility.BUISNESSUNIT_AND_MODERATORS);
+        // TODO set clearing state
```
bs-ondem: Is there an issue for the TODO?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/SW360Attributes.java
neubs-bsi: those are currently still in use. since I have not changed the way Licenses get serialized, since this changes based on whether or not the `SW360License` has a full name or not. I didn't find a way to implement this logic in JsonSerialization. Suggestions are welcome

Issue #371, opened 2019-12-06
SW360 name of components not including namespace


neubs-bsi:
### Summary of the Bug
```
Due to this code in the `SW360ComponentAdapterUtils`:
return artifact.askFor(ArtifactCoordinates.class)
                .map(ArtifactCoordinates::getMainCoordinate)
                .map(Coordinate::getName)
                .orElse(artifact.toString())
```
the name of the component is not set with namespace, since PackageUrl.getName() only returns the name, not the namespace. 
Hence we have Maven components in the database that do not have a namespace. 
Since there are a lot of components in Maven that have the same artifact id, but not the same group-id as well, I think it is important to rework this to reflect a way the component can be more uniquely identified. 

### Steps to reproduce
- run example project and check out created components. 

### Acceptance Criteria
- [ ] names of individual package sources are better represented in our SW360 Component modell

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

Issue #348, opened 2019-11-18
403 ModSecurity Action - SW360ComponentClient.searchByName


neubs-bsi:
### Summary of the Bug
When running the Example project with an activated SW360 Updater, the build fails with this Error message when handling the Artifact "Arbitrary Copied Code":
```

Caused by: org.springframework.web.client.HttpClientErrorException$Forbidden: 403 ModSecurity Action
    at org.springframework.web.client.HttpClientErrorException.create (HttpClientErrorException.java:83)
    at org.springframework.web.client.DefaultResponseErrorHandler.handleError (DefaultResponseErrorHandler.java:97)
    at org.springframework.web.client.DefaultResponseErrorHandler.handleError (DefaultResponseErrorHandler.java:79)
    at org.springframework.web.client.ResponseErrorHandler.handleError (ResponseErrorHandler.java:63)
    at org.springframework.web.client.RestTemplate.handleResponse (RestTemplate.java:777)
    at org.springframework.web.client.RestTemplate.doExecute (RestTemplate.java:735)
    at org.springframework.web.client.RestTemplate.execute (RestTemplate.java:669)
    at org.springframework.web.client.RestTemplate.exchange (RestTemplate.java:607)
    at org.eclipse.sw360.antenna.sw360.rest.SW360Client.doRestCall (SW360Client.java:58)
    at org.eclipse.sw360.antenna.sw360.rest.SW360Client.doRestGET (SW360Client.java:66)
    at org.eclipse.sw360.antenna.sw360.rest.SW360ComponentClient.searchByName (SW360ComponentClient.java:69)
    at org.eclipse.sw360.antenna.sw360.adapter.SW360ComponentClientAdapter.getComponentByName (SW360ComponentClientAdapter.java:62)
    at org.eclipse.sw360.antenna.sw360.adapter.SW360ReleaseClientAdapter.getReleaseByNameAndVersion (SW360ReleaseClientAdapter.java:125)
    at org.eclipse.sw360.antenna.sw360.adapter.SW360ReleaseClientAdapter.getSparseRelease (SW360ReleaseClientAdapter.java:104)
    at org.eclipse.sw360.antenna.sw360.adapter.SW360ReleaseClientAdapter.getRelease (SW360ReleaseClientAdapter.java:108)
    at org.eclipse.sw360.antenna.sw360.adapter.SW360ReleaseClientAdapter.getOrCreateRelease (SW360ReleaseClientAdapter.java:46)
[...]

PR #380, opened 2019-12-16
Including namespace of Coordinate in SW360Component name

bs-ondem:
Fixes https://github.com/eclipse/antenna/issues/371
Creating the name of the SW360Component with the namespace and name of
the Coordinate fact of the artifact using a '/' as the delimiter
identical to the PURLs.

The SW360ComponentAdapterUtils::createComponentName method tries to
extract the component name from the ArtifactCoordinate fact first. If
it doesn't exist, it extracts it from the ArtifactFilename fact.
Otherwise it returns an empty string.

### Request Reviewer

### Type of Change
*improvements*

### Checklist
Must:

Optional:

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
@@ -18,6 +18,8 @@ import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360Component;
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360SparseComponent;
 import org.eclipse.sw360.antenna.sw360.utils.SW360ComponentAdapterUtils;
 import org.eclipse.sw360.antenna.util.ProxySettings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpHeaders;
 
 import java.util.ArrayList;
@@ -26,6 +28,8 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class SW360ComponentClientAdapter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SW360ComponentClientAdapter.class);
+
     private final SW360ComponentClient componentClient;
 
     public SW360ComponentClientAdapter(String restUrl, ProxySettings proxySettings) {
@@ -52,9 +56,14 @@ public class SW360ComponentClientAdapter {
     }
 
     public Optional<SW360Component> getComponentByArtifact(Artifact artifact, HttpHeaders header) {
-        String componentName = SW360ComponentAdapterUtils.createComponentName(artifact);
-
-        return getComponentByName(componentName, header);
+        try {
+            String componentName = SW360ComponentAdapterUtils.createComponentName(artifact);
+            return getComponentByName(componentName, header);
+        } catch (ExecutionException e) {
+            LOGGER.debug("No component found for {}. Reason: {}", artifact.prettyPrint(), e.getMessage());
+            LOGGER.debug("Error: ", e);
+            return Optional.empty();
+        }
     }
 
     public Optional<SW360Component> getComponentByName(String componentName, HttpHeaders header) {
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
@@ -10,19 +10,31 @@
  */
 package org.eclipse.sw360.antenna.sw360.utils;
 
+import org.eclipse.sw360.antenna.api.exceptions.ExecutionException;
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
 import org.eclipse.sw360.antenna.model.artifact.ArtifactCoordinates;
+import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactFilename;
+import org.eclipse.sw360.antenna.model.artifact.facts.ArtifactFilename.ArtifactFilenameEntry;
 import org.eclipse.sw360.antenna.model.coordinates.Coordinate;
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360Component;
 import org.eclipse.sw360.antenna.sw360.rest.resource.components.SW360ComponentType;
 import org.eclipse.sw360.antenna.sw360.rest.resource.releases.SW360Release;
 
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 public class SW360ComponentAdapterUtils {
     public static String createComponentName(Artifact artifact) {
         return artifact.askFor(ArtifactCoordinates.class)
                 .map(ArtifactCoordinates::getMainCoordinate)
-                .map(Coordinate::getName)
-                .orElse(artifact.toString()); // TODO: ugly hack
+                .map(o -> Stream.of(o.getNamespace(), o.getName()))
+                .map(o -> o.collect(Collectors.joining("/")))
+                .orElseGet(() -> artifact.askFor(ArtifactFilename.class)
+                        .flatMap(ArtifactFilename::getBestFilenameEntryGuess)
+                        .map(ArtifactFilenameEntry::getFilename)
+                        .orElseThrow(() -> new ExecutionException("Artifact " + artifact.prettyPrint() + " does not have " +
+                                "enough facts to create a component name. Please provide more information " +
+                                "by an analyzer (" + artifact.getAnalysisSource() + ")")));
     }
 
     public static String createComponentVersion(Artifact artifact) {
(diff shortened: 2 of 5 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
```diff
@@ -12,17 +12,26 @@
...
         return artifact.askFor(ArtifactCoordinates.class)
                 .map(ArtifactCoordinates::getMainCoordinate)
-                .map(Coordinate::getName)
-                .orElse(artifact.toString()); // TODO: ugly hack
+                .map(o -> Stream.of(o.getNamespace(), o.getName()))
+                .map(o -> o.collect(Collectors.joining("/")))
+                .orElse(artifact.askFor(ArtifactFilename.class)
+                        .flatMap(ArtifactFilename::getBestFilenameEntryGuess)
+                        .map(ArtifactFilenameEntry::getFilename)
+                        .orElse(""));
```
blaumeiser-at-bosch: We could use something like a UUID for the name, an empty name is perhaps not the right option

neubs-bsi: I am not really happy with an empty string. This is the name that will stand in SW360 and there shouldn't be multiple ones with empty string. Additionally by this name the release with be searched, making wrong assignments to releases with the same version and also an empty string name very likely.

neubs-bsi: I think, if there is no name here, we maybe shouldn't have this in SW360 at all. 
@blaumeiser-at-bosch what do you think?

blaumeiser-at-bosch: That was my though as well, if we do not have enough meat to at least have a simple name, we should inform the user that there is something strange here, and we should not push arbitrary data to SW360. At least a filename should be there.

bs-ondem: Yes the user should be informed, but not in the adapter utils class? The component client will fail in creating a component with an empty string and will log an error.

neubs-bsi: Do you mean when the request is made?
I feel like we should try to avoid making requests we know will fail to save resources.

bs-ondem: We are trying to avoid it with the `CoordinatesValidator`

blaumeiser-at-bosch: I totally agree that here is not the point of informing the user here. This is a helper who does something useful. The question is, whether no information at all should return an error aka an exception instead of an empty string?

bs-ondem: Just for clarity, is that the flow you want?
```
if AritfactCoordinate then
   return <component name extracted from Coordinates>
else if ArtifactFilename then
   return <component name extracted from ArtifactFilenameEntry>
else 
   throw new ExecutionException(..);
```

blaumeiser-at-bosch: Yes, I think this makes sense. This exception can be caught and it is easier to distinguish between make the right thing and do something exceptional

neubs-bsi:
Just as a question. Does this PR also coincidentally solve this issue: #348 ?
Since it is also related to the method your PR is changing

bs-ondem:
@neubs-bsi I'm not sure, but I will have a look at it after improving my changes based on your comments.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
```diff
@@ -10,19 +10,29 @@
...
         return artifact.askFor(ArtifactCoordinates.class)
                 .map(ArtifactCoordinates::getMainCoordinate)
-                .map(Coordinate::getName)
-                .orElse(artifact.toString()); // TODO: ugly hack
+                .map(o -> Stream.of(o.getNamespace(), o.getName()))
+                .map(o -> o.collect(Collectors.joining("/")))
+                .orElseGet(() -> artifact.askFor(ArtifactFilename.class)
+                        .flatMap(ArtifactFilename::getBestFilenameEntryGuess)
+                        .map(ArtifactFilenameEntry::getFilename)
+                        .orElseThrow(() -> new ExecutionException("No suitable component name found.")));
```
blaumeiser-at-bosch: Error message is misleading. Should be something like artifact does not have enough information to build a name + artifact.toString() or something like this, Enough information that allows us to understand what is going on, if a customer comes back to us with this problem!

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/adapter/SW360ComponentClientAdapter.java
```diff
@@ -52,9 +56,13 @@ public SW360Component createComponent(SW360Component component, HttpHeaders head
...
 
     public Optional<SW360Component> getComponentByArtifact(Artifact artifact, HttpHeaders header) {
-        String componentName = SW360ComponentAdapterUtils.createComponentName(artifact);
-
-        return getComponentByName(componentName, header);
+        try {
+            String componentName = SW360ComponentAdapterUtils.createComponentName(artifact);
+            return getComponentByName(componentName, header);
+        } catch (ExecutionException e) {
+            LOGGER.debug("No component found for {}. Reason: {}", artifact, e.getMessage());
```
blaumeiser-at-bosch: Always log the exception, especially debug messages are meant for us to understand what is going on, if a user comes back to us, so don't be shy and print out whatever you think could help us in identifying the problem.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
```diff
@@ -49,19 +53,23 @@ public SW360UpdaterImpl(SW360MetaDataUpdater sw360MetaDataUpdater,
...
+                if (sw360MetaDataUpdater.isUploadSources()
+                        && sw360ReleaseFinal.get_Links().getSelf() != null
+                        && !sw360ReleaseFinal.get_Links().getSelf().getHref().isEmpty()) {
+                    Map<Path, SW360AttachmentType> attachments = SW360AttachmentAdapterUtils.getAttachmentsFromArtifact(artifact);
+                    if (!attachments.isEmpty()) {
+                        sw360ReleaseFinal = sw360MetaDataUpdater.uploadAttachments(sw360ReleaseFinal, attachments);
+                    }
                 }
+                releases.add(sw360ReleaseFinal);
+            } catch(Exception e) {
```
blaumeiser-at-bosch: Never catch a generic Exception, you also catch all RuntimeExceptions and the like, which should not be caught.

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
```diff
@@ -49,19 +53,23 @@ public SW360UpdaterImpl(SW360MetaDataUpdater sw360MetaDataUpdater,
...
+                LOGGER.warn("Release for {} will not be created in SW360 due to missing facts. Reason: {}",
```
blaumeiser-at-bosch: Also log the exception as debug entry! Actually I think this should even be logged as an error because the user might ignore warnings, but with this issue here, he will never get an approval for the project, he has to do something and this is necessary to address with him

Actually, the log message should contain a hint about what needs to be done. Without the component in SW360, the compliance office will not work on the component. What can he do to solve this? Good question, isn't it, can he do something in the config file?

I have to admit, I miss the situation as of now, where this case can happen at all because we will detect either a coordinate because we get information from a package manager, or we detect a file so we have at least a filename. Is there any way, in which we can detect a dependency without any information? I have no idea how. Since we do not expect that, perhaps it is even feasible to break completely, so no caught exception here. If something happens, that cannot be, it is best to not react at all, so that the user is coming to us in order to solve the issue.

neubs-bsi: Yes, when a component is added via the antennaconf.xml this can happen. 
Especially then it is important that the user adds enough information so this component can be used correctly.

blaumeiser-at-bosch: THen we should break at reading in the config file if we cannot detect enough informationen for a coordinate.

bs-ondem: Follow up issue for config.xml https://github.com/eclipse/antenna/issues/384

PR #382, opened 2019-12-17
Add createdOn variable to SW360Release

neubs-bsi:
Added createdOn variable to SW360Release

### Request Reviewer
@blaumeiser-at-bosch 

*Type of change*:  Other

Unit tests still pass


neubs-bsi:
@blaumeiser-at-bosch 
It is used in the CSV Exporter only. Due to the de-serialization it gets filled when getting a Release from SW360 and we can then use it to sort it in the Exporter. It is not needed or used anywhere else, which is also the reason why I didn't include in the equals or hashCode methods

blaumeiser-at-bosch:
> @blaumeiser-at-bosch
> It is used in the CSV Exporter only. Due to the de-serialization it gets filled when getting a Release from SW360 and we can then use it to sort it in the Exporter. It is not needed or used anywhere else, which is also the reason why I didn't include in the equals or hashCode methods

You are right with equals and hashCode, but where is it created, is this a SW360 property? I was wondering, because only the getter has a JSON annotation and I would have expected only the setter to have an annotation.

Issue #367, opened 2019-12-02
Logger in antenna cli not working


blaumeiser-at-bosch:
### Summary of the Bug
Currently, the command line interface of Antenna is not giving output above Error Level

### Steps to reproduce
- Run Antenna CLI, observe, not output is given, except Error level messages

### Acceptance Criteria
- [ ] Log output is given as in other clients

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

Issue #385, opened 2019-12-18
Harmonize Logging to use Log4j2 or at least one common logging


blaumeiser-at-bosch:
### Summary of the Bug

Currently we have a configuration for log4j and log4j2, Ort seems to use another logging technology which does some logging in between, ...

This should be harmonized to have only one logging used throughout the tooling.

In addition there is a log message from the ORT Result analyzer which seems to come from the used ORT libraries, this information should not be logged.

### Steps to reproduce
- Observe core/runtime/src/main/resources/log4j*.properties
- Observe when runnig the antenna cli that there is a ort log message

### Acceptance Criteria
- [x] One configuration is setup with one common backend
- [x] No special ORT log messages, they should be configured out

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

sschuberth:
> In addition there is a log message from the ORT Result analyzer which seems to come from the used ORT libraries, this information should not be logged.

Actually, why not? If the ORT analyzer is used, you might want to know if something goes wrong, and log output can be really helpful then. I'd rather configure the logger in a way that the FQN of the class is being logged, so that it becomes evident whether the log output originates from ORT, or Antenna directly.

blaumeiser-at-bosch:
> > In addition there is a log message from the ORT Result analyzer which seems to come from the used ORT libraries, this information should not be logged.
> 
> Actually, why not? If the ORT analyzer is used, you might want to know if something goes wrong, and log output can be really helpful then. I'd rather configure the logger in a way that the FQN of the class is being logged, so that it becomes evident whether the log output originates from ORT, or Antenna directly.

I reduced the normal logging to a minimum, so that if you run antenna, you get basically progress information on which step is currently active, if no warn or error is created each step logs not much except for long lasting activities. In this output sequence the ORT log information looks somewhat detailed compared to the rest, so pure aesthetics.

oheger-bosch:
[...]

Issue #383, opened 2019-12-18
Configure Log Level for CLI


blaumeiser-at-bosch:
### Summary of the Feature
It should be possible to configure the log level in the command line interface. This is needed so that customers can run in DEBUG mode to provide us with detailed information of their problem.

### Acceptance Criteria
- [ ] A configuration parameter allows to run a higher log leven than INFO, if is ok, to only allow to switch on DEBUG level instead of offering the whole range of different log levels.
- [ ] If run with this parameter, logging information on the specified level is reported by the tool,. e.g. DEBUG information if log level is set to debug

### Definition of Done
- Acceptance criteria fulfilled
- A PR is created, the CI infrastructure reports green
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Test cases are created to prove the functionality of the feature
- Documentation is updated

sschuberth:
FYI, for Log4j 2 the log level can be set at runtime via [Configurator.setRootLevel()](https://github.com/heremaps/oss-review-toolkit/blob/0bcabe6850c0d6ee3cf3be07a6d1e13fe8d534e6/cli/src/main/kotlin/OrtMain.kt#L127-L130).

blaumeiser-at-bosch:
Review 2020-02-20

PR #386, opened 2019-12-18
Fix immediate effect of antenna cli not providing log messages

blaumeiser-at-bosch:
Fix immediate effect of antenna cli not providing log messages

- Add a configuration to log4j2.properties to configure the log backend for antenna cli which uses log4j2 as log backend
- Remove properties from log4j.properties that have no effect on the logging of other frontends

Still open:
- The different antenna frontends use different logging configurations (log4j vs. log4j2), that needs to be harmonized -> #385
- There is a log message coming from the ORT Result Analyzer that isn't configured out, this should be done to clean the output -> #385
- The CLI should have a parameter to set the log level to default -> #383

### Request Reviewer
@bs-ondem or
@neubs-bsi or
@sschuberth 

### Type of Change

*Type of change*:  bug fix

### How Has This Been Tested?
Run different ways like CLI and Maven frontend 

### Checklist
Must:

closes: #367

### DIFF
--- a/core/runtime/src/main/resources/log4j2.properties
+++ b/core/runtime/src/main/resources/log4j2.properties
@@ -8,4 +8,14 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-log4j2.simplelogStatusLoggerLevel=INFO
+name = PropertiesConfig
+
+appenders = console
+
+appender.console.type = Console
+appender.console.name = STDOUT
+appender.console.layout.type = PatternLayout
+
+rootLogger.level = info
+rootLogger.appenderRefs = stdout
+rootLogger.appenderRef.stdout.ref = STDOUT
--- a/core/runtime/src/main/resources/log4j.properties
+++ b/core/runtime/src/main/resources/log4j.properties
@@ -12,19 +12,10 @@
 # Root logger option
 log4j.rootLogger=INFO, stdout
 
-
 #  Logging level
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.Target=System.out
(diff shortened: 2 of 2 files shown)


core/runtime/src/main/resources/log4j2.properties
```diff
@@ -8,4 +8,14 @@
...
+
+appenders = console
+
+appender.console.type = Console
+appender.console.name = STDOUT
+appender.console.layout.type = PatternLayout
+
+rootLogger.level = info
+rootLogger.appenderRefs = stdout
+rootLogger.appenderRef.stdout.ref = STDOUT
```
sschuberth: Nit: Missing final new line at end of file.

blaumeiser-at-bosch:
> To me, it's unclear from reading the commit message _why_ it's fine to delete those properties from `log4j.properties` without any substitution. Could you please elaborate?
> 
> Similarly, _why_ do all these other properties need to be added to `log4j2.properties`?
> 
> Also, as this is described as a clean-up commit, _why_ do we actually have _both_ `log4j.properties` and `log4j2.properties` in the same project? I would have expected a clean-up to address exactly that, and get rid of one in favor of the other.
> 
> Finally, I guess we should not mention "Tina" as part of the commit message.

Hope the new information solves your request for information better, also fixed the missing new line.

sschuberth:
Thanks, the new message is much better! Some final commit message nits:

> There is a log essage coming from the ORT Result Analyzer that uses a different logger, this information should be prevented from being logged

s/essage/message/

And ORT also uses Log4j 2, you may want to name that explicitly instead of referring to "a different logger".

PR #390, opened 2020-01-09
Fix ORTDownloaderProcessor in case of the procotol of url is SSH

bs-ondem:
In some cases, the protocol of a VcsInfo.url is SSH, which leads to
hanging builds, because it's prompting for a password. To avoid this,
we are replacing the protocol from ssh to https for github.com hosts.

### Type of Change
*bug fix*

### DIFF
--- a/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
+++ b/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
@@ -48,7 +48,7 @@ private fun mapSourceUrl(pkg: Package): ArtifactSourceUrl? =
     }
 
 private fun mapVcsInfo(pkg: Package) : ArtifactVcsInfo? =
-        pkg.vcs.takeUnless { it == VcsInfo.EMPTY }?.let {
+        pkg.vcsProcessed.takeUnless { it == VcsInfo.EMPTY }?.let {
             ArtifactVcsInfo(it.type.toString(), it.url, it.revision)
         }
 
--- a/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
+++ b/modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
@@ -47,7 +47,7 @@ private fun mapSourceUrl(pkg: Package): ArtifactSourceUrl? =
         ArtifactSourceUrl(it)
     }
 
-private fun mapVcsInfo(pkg: Package) : ArtifactVcsInfo? =
+private fun mapToArtifactVcsInfo(pkg: Package) : ArtifactVcsInfo? =
         pkg.vcsProcessed.takeUnless { it == VcsInfo.EMPTY }?.let {
             ArtifactVcsInfo(it.type.toString(), it.url, it.revision)
         }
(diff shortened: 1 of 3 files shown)


modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -61,6 +61,13 @@ class OrtDownloaderProcessor : AbstractProcessor() {
             try {
                 LOGGER.debug("Download sources via ORT Downloader for '${name}'")
                 val ortDownloadDirectory = createTempDir("ortDownloaderDirectory").also { it.deleteOnExit() }
+                // In some cases, the protocol of a VcsInfo.url is SSH, which leads to hanging builds,
```
blaumeiser-at-bosch: I am not sure if this is needed here.

@sschuberth, if  I understood you right, this is exactly what processed already does, right? What about Gitlab and Bitbucket, the same applies for that, or not?

sschuberth: It is needed; that's the thing I wanted to discuss with @bs-ondem yesterday.

While the `url` in `vcsProcessed` is normalized from an [scp-like Git URL](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) like [git@github.com:kaelzhang/node-ignore.git](https://github.com/kaelzhang/node-ignore/blob/aa4e6b6bd9a22061364d8816d39ec0c15fc7cd47/package.json#L28) to a standard `ssh`-URL like `ssh://git@github.com/kaelzhang/node-ignore.git`, the normalization deliberately does *not* change the protocol.

It's a common mistake of package maintainers to use a clone URL that requires authorization in meta-data fields that just demand read-only access. Ideally, this should be reported upstream, but this would not fix the meta-data for already released packages. In plain ORT, the correct way to address this issue would be to create a [curation](https://github.com/heremaps/oss-review-toolkit/blob/master/docs/config-file-curations-yml.md) for the package. But for the sake of simplicity, and as it's like said a common mistake, my suggestion was to hard-code a generic "curation" that maps `ssh`- to `https`-URLs here.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt
```diff
@@ -61,6 +61,13 @@ class OrtDownloaderProcessor : AbstractProcessor() {
...
+                // because it's prompting for a password. To avoid this, we are replacing the protocol from ssh
+                // to https for github.com hosts.
+                if (pkg.vcsProcessed.url.startsWith("ssh://git@github.com")){
+                    val newUrl = pkg.vcsProcessed.url.replace("ssh://git@github.com", "https://git@github.com")
```
sschuberth: We need to drop the generic `git` username from the `https` URL as

1. in contrast to `ssh`, `https` uses the GitHub username for authentication, and
2. we do not want / need to authenticate at all, as we only clone, but not push.

modules/ort/src/main/kotlin/resolver/OrtResultArtifactResolver.kt
```diff
@@ -47,8 +47,8 @@ private fun mapSourceUrl(pkg: Package): ArtifactSourceUrl? =
         ArtifactSourceUrl(it)
     }
 
-private fun mapVcsInfo(pkg: Package) : ArtifactVcsInfo? =
-        pkg.vcs.takeUnless { it == VcsInfo.EMPTY }?.let {
+private fun mapToArtifactVcsInfo(pkg: Package) : ArtifactVcsInfo? =
```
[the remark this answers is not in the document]
bs-ondem: I split the commit to separate them.

PR #396, opened 2020-01-13
refactor sw360 workflow items

neubs-bsi:
extract methods from process to clean code a bit.
this makes reading the code easier.

@bs-ondem 

All tests pass

### DIFF
--- a/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
+++ b/modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
@@ -45,29 +45,10 @@ public class SW360UpdaterImpl {
     public Map<String, IAttachable> produce(Collection<Artifact> intermediates) {
         List<SW360Release> releases = new ArrayList<>();
         for (Artifact artifact : intermediates) {
-            List<License> availableLicenses = ArtifactLicenseUtils.getFinalLicenses(artifact).getLicenses();
-
-            Set<SW360License> detectedLicenses = sw360MetaDataUpdater.getLicenses(availableLicenses);
-            Set<String> licenseIds = Collections.emptySet();
-            if (detectedLicenses.size() == availableLicenses.size()) {
-                licenseIds = detectedLicenses.stream()
-                        .map(SW360License::getShortName)
-                        .collect(Collectors.toSet());
-            }
             try {
-                final SW360Release sw360ReleaseFromArtifact = SW360ReleaseAdapterUtils.convertToRelease(artifact);
-                sw360ReleaseFromArtifact.setMainLicenseIds(licenseIds);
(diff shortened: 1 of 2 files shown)


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
```diff
@@ -45,35 +40,41 @@ public SW360UpdaterImpl(SW360MetaDataUpdater sw360MetaDataUpdater,
...
-                        && sw360ReleaseFinal.get_Links().getSelf() != null
-                        && !sw360ReleaseFinal.get_Links().getSelf().getHref().isEmpty()) {
-                    Map<Path, SW360AttachmentType> attachments = SW360AttachmentAdapterUtils.getAttachmentsFromArtifact(artifact);
-                    if (!attachments.isEmpty()) {
-                        sw360ReleaseFinal = sw360MetaDataUpdater.uploadAttachments(sw360ReleaseFinal, attachments);
-                    }
-                }
-                releases.add(sw360ReleaseFinal);
-            } catch(ExecutionException e) {
-                LOGGER.error("Release will not be created in SW360. Reason: {}", e.getMessage());
```
bs-ondem: Where are these logs in the refactoring?

modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterImpl.java
```diff
@@ -45,35 +40,41 @@ public SW360UpdaterImpl(SW360MetaDataUpdater sw360MetaDataUpdater,
...
-                LOGGER.debug("Error: ", e);
-            }
+            SW360Release sw360ReleaseFinal = artifactToReleaseInSW360(artifact);
```
bs-ondem:
```suggestion
        List<SW360Release> releases = intermediates.stream()
                .map(this::artifactToReleaseInSW360)
                .collect(Collectors.toList());
```

neubs-bsi: @bs-ondem does this seem suitable to you?
```
List<SW360Release> releases = intermediates.stream()
                .map(artifact -> {
                    try {
                        return artifactToReleaseInSW360(artifact);
                    } catch (ExecutionException e) {
                        LOGGER.error("Release will not be created in SW360. Reason: {}", e.getMessage());
                        return null;
                    }
                })
                .filter(Objects::nonNull)
                .collect(Collectors.toList());
```

bs-ondem: Hmmm... my suggestion was for your changes without the logging.. So in this case you can keep the for loop and put the logging inside the loop... I don't like streams with try-blocks

Issue #404, opened 2020-01-20
Combine CSV Writer and Generator into a CSVReader


neubs-bsi:
### Summary of the Improvement
Currently there are two places in the code that use the same CSV formatting.
One in the `CSVAnalyzerImpl`, where a format is used for reading and once in the compliance tool `SW360ExporterCSVWriter` that writes a csv file with the same format. 

This should be unified to a CSV Reader that allows to update the format in one and not two places. 

### Acceptance Criteria
- [ ] Writting and Reading both happen in a CSV Reader (compare to: JsonReader)
- [ ] Code is cleaned up

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

neubs-bsi:
Already implemented

Issue #405, opened 2020-01-20
Create roundtrip: CSVAnalyzer provides SourceFilePath -> SW360Updater upload sources


neubs-bsi:
### Summary of the Improvement
Currently CSVAnalyzer creates ArtifactPathnames which isn't used for in the SW360Updater to upload sources. 

The CSVAnalyzer should provide all the fields we want to use for creating a release in SW360 and all its corresponding data. 

### Acceptance Criteria
- [ ] CSV Analyzer - Sw360 Updater workflow is complete with current data model

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

PR #406, opened 2020-01-21
CsvReader from csvwriter and csvanalyzer

neubs-bsi:
#404 
This combines the csv analyzer functionality of reading a csv file and mapping it on an artifact with the SW360ExporterCSVWriter functionality to produce one. 

Several places needed adjustment to create a roundtrip. 

@blaumeiser-at-bosch @bs-ondem 

Tests were moved to new corresponding classes. 

### Checklist
Must:


modules/sw360/pom.xml
```diff
@@ -159,6 +159,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
```
blaumeiser-at-bosch: Why do you need the simple slf4j backend implementation? Is this a test dependency?

neubs-bsi: This gets rid of the warning of the missing SLFJ4 warning when executing the tests. 
I can remove this though and do it in a separate pull request.

blaumeiser-at-bosch: So you could add it as a testing dependency?

neubs-bsi: I think this occurs when executing the classes in general, it just is visible when executing the tests.
I'd rather check this out later again and remove the change for now, if you don't mind.

neubs-bsi:
TODO in code refers to #405

PR #408, opened 2020-01-23
Sources handling csvreader and compliance tool

neubs-bsi:
#405 

Added sources handling to csvreader and compliance tool

@blaumeiser-at-bosch 
https://github.com/oheger-bosch

*Type of change*:  Improvement

Unit tests added

### Checklist
Must:


neubs-bsi:
@oheger-bosch

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SW360Exporter.java
```diff
@@ -49,7 +53,15 @@ public void execute() {
         Collection<SW360Release> sw360ReleasesNotApproved = getNonApprovedReleasesFromSpareReleases(sw360SparseReleases, headers);
 
         List<Artifact> artifacts = sw360ReleasesNotApproved.stream()
-                .map(release -> SW360ReleaseAdapterUtils.convertToArtifactWithoutSourceFile(release, new Artifact("SW360")))
+                .map(release -> {
```
blaumeiser-at-bosch: This is very much code for a lambda. Could you extract a meaningful method.

assembly/compliance-tool/src/test/resources/test-source.txt
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
blaumeiser-at-bosch: Interesting question, is this a copyrightable content? 😀

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/csvreader/CSVReader.java
```diff
@@ -395,10 +399,16 @@ private static String mapCPEIdToString(Artifact artifact) {
     }
 
     private static String mapFileNameToString(Artifact artifact) {
```
blaumeiser-at-bosch: I think the naming here should be adapted, actually it is more something like getFilepathAsString(), right?

core/core-workflow-steps/src/test/resources/CsvAnalyzerTest/test_source.txt
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
blaumeiser-at-bosch: Perhaps if you claim, that it is minimalistic art 😀

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SW360Exporter.java
```diff
@@ -49,7 +53,15 @@ public void execute() {
         Collection<SW360Release> sw360ReleasesNotApproved = getNonApprovedReleasesFromSpareReleases(sw360SparseReleases, headers);
 
         List<Artifact> artifacts = sw360ReleasesNotApproved.stream()
-                .map(release -> SW360ReleaseAdapterUtils.convertToArtifact(release, new Artifact("SW360")))
+                .map(release -> {
+                    Artifact artifact = SW360ReleaseAdapterUtils.convertToArtifactWithoutSourceFile(release, new Artifact("SW360"));
+                    Set<SW360SparseAttachment> sparseAttachments = getSparseAttachmentsSource(release);
+                    sparseAttachments.forEach(sparseAttachment -> {
+                        Optional<Path> path = connectionConfiguration.getSW360ReleaseClientAdapter().downloadAttachment(release, sparseAttachment, configuration.getSourcesPath(), headers);
+                        path.map(pth -> artifact.addFact(new ArtifactSourceFile(pth)));
```
oheger-bosch: Better use path.ifPresent(); map() is more to apply a function to the value which does not have a side effect.

assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SW360ExporterTest.java
```diff
@@ -105,5 +108,6 @@ public void testExporter() throws IOException {
         assertThat(csvFile).exists();
         CSVParser csvParser = SW360TestUtils.getCsvParser(csvFile);
         assertThat(csvParser.getRecords().size()).isEqualTo(1);
+        verify(releaseClientAdapterMock, atLeast(1)).downloadAttachment(any(), any(),any() ,any());
```
oheger-bosch: Maybe some concrete expected values can be used instead of the any()s to make the test more meaningful.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/csvreader/CSVReader.java
```diff
@@ -255,15 +255,19 @@ private void addOptionalArtifactFacts(CSVRecord record, Artifact artifact) {
         }
         if (checkIfRecordIsMappedAndNotEmptyForParameter(record, PATH_NAME)) {
             String pathName = record.get(PATH_NAME);
-            String absolutePathName = makePathAbsolute(pathName);
-            artifact.addFact(new ArtifactPathnames(absolutePathName));
+            Path path = getPathFromPathname(pathName);
+            if (path.toFile().exists()) {
```
oheger-bosch: You could use java.nio.file.Files.exists(path) to check whether the path exists without having it to convert to a File.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/csvreader/CSVReader.java
```diff
@@ -394,11 +399,17 @@ private static String mapCPEIdToString(Artifact artifact) {
...
-    private static String mapFileNameToString(Artifact artifact) {
-        return artifact.askFor(ArtifactSourceFile.class)
+    private static String getFilepathAsString(Artifact artifact) {
+        Optional<Path> path = artifact.askFor(ArtifactSourceFile.class)
                 .map(ArtifactFactWithPayload::get)
                 .map(Path::toAbsolutePath)
-                .map(Path::toString)
-                .orElse("");
+                .filter(pth -> pth.toFile().exists());
+        if (path.isPresent()) {
```
blaumeiser-at-bosch: How about
.map(Path::toString)
.orElseGet(log + "")

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/csvreader/CSVReader.java
```diff
@@ -394,11 +399,17 @@ private static String mapCPEIdToString(Artifact artifact) {
...
+            return path.get().toString();
+        } else {
+            artifact.getMainCoordinate().ifPresent(coordinate ->
+                    LOGGER.debug("The given source file {} for artifact {} does not exist.", path, coordinate));
```
blaumeiser-at-bosch: Äh, path.isPresent() is false, what does this log message give, it uses an empty optional.toString, and what in case of a non existing main coordinate, there is still an artifact, how to deal with that?

core/runtime/src/main/java/org/eclipse/sw360/antenna/util/LicenseSupport.java
```diff
@@ -98,4 +98,14 @@ public static LicenseInformation fromSPDXExpression(SpdxExpression spdxExpressio
         }
         throw new ExecutionException("SPDX expression=[" + spdxExpression.toString() + "] could not be parsed");
     }
+
```
blaumeiser-at-bosch: The following code will collide with Onur, please talk

bs-ondem: It's just a new method and I think my changes will be merged after this PR, so i can adapt it during the rebase.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/csvreader/CSVReader.java
```diff
@@ -394,11 +399,16 @@ private static String mapCPEIdToString(Artifact artifact) {
...
-    private static String mapFileNameToString(Artifact artifact) {
-        return artifact.askFor(ArtifactSourceFile.class)
+    private static String getFilepathAsString(Artifact artifact) {
+        Optional<Path> path = artifact.askFor(ArtifactSourceFile.class)
                 .map(ArtifactFactWithPayload::get)
                 .map(Path::toAbsolutePath)
-                .map(Path::toString)
-                .orElse("");
+                .filter(pth -> pth.toFile().exists());
+        return path.map(Path::toString)
```
blaumeiser-at-bosch: Is there a reason why you have an obsolete path variable here? You could start with return artifact.askFor (...);

If I get you right, you should stop prior to the filter method. There are two cases for logging.
1. There is no ArtifactSourceFileFact. This is a problem, because we require to have a source file linked. This should be mentioned in output.
2. THere is an ArtifactSourceFileFact, but the the path does not reference an existing file, this is what you want to log below.

This requires that you break the optiona cascade prior to the .filter call. THere you can check whether you have a path and if you have a path to check if it exists. If both is good, you can continue by mapping the path to a string.

neubs-bsi: > Is there a reason why you have an obsolete path variable here? You could start with return artifact.askFor (...); 

 I found that the statement itself was getting a bit to complex to be just one return statement. 

> There is no ArtifactSourceFileFact. This is a problem, because we require to have a source file linked. This should be mentioned in output.

This is just a csv reader. This function is part of _writing_ artifacts to a csv file. In our direct use case this wouldn't be a problem, it would only mean that an artifact exported from SW360 does not have a source file, which a team could fill in the updater part of the compliance tool. It is just like any other missing fact and it does not require any additional output.  

My suggestion:
```
Optional<Path> path = artifact.askFor(ArtifactSourceFile.class)
                .map(ArtifactFactWithPayload::get);
        if (path.isPresent()) {
            return path.map(Path::toAbsolutePath)
                    .filter(pth -> pth.toFile().exists())
                    .map(Path::toString)
                    .orElseGet(() -> {
                                artifact.getMainCoordinate().ifPresent(coordinate ->
                                        LOGGER.debug("The given source file for artifact {} does not exist", coordinate));
                                return "";
                            });
        } else {
            return "";
        }
```

This way we don't give a wrong warning when the fact isn't present. and give a logger debug when the fact is present but the file doesn't exist

blaumeiser-at-bosch: How about
`return artifact.askFor(ArtifactSourceFile.class)
    .map(ArtifactFactWithPayload::get)
    .map(this::getPathAsString)
    .orElse("");`

And the rest is then in this second method getPathAsString. This mixing of procedural and functional programming looks always a bit ugly.

neubs-bsi: that is possible, however I still prefer the above method.
Since we would have to give the artifact, or at least coordinate of the artifact to the `getPathAsString` method as well for the log message. and since we give a path and not n optional would lose the easy functional programming and make it procedural. For such a small method that seems over the top

neubs-bsi:
```
private static String getFilepathAsString(Artifact artifact) {
        return artifact.askFor(ArtifactSourceFile.class)
                .map(ArtifactFactWithPayload::get)
                .map(pth -> CSVReader.getPathAsString(pth, artifact))
                .orElse("");
    }

    private static String getPathAsString(Path path, Artifact artifact) {
        File file = path.toAbsolutePath().toFile();
        if (file.exists()) {
            return file.toString();
        } else {
            artifact.getMainCoordinate().ifPresent(coordinate ->
                    LOGGER.debug("The given source file for artifact {} does not exist", coordinate));
            return "";
        }
    }
```

Issue #409, opened 2020-01-24
New license model which replaces the through JAXB generated model with actual Java classes


bs-ondem:
### Summary of the Improvement
Currently the license model is specified in XSD, which generates the Java classes with JAXB. This limits our flexibility to support e.g. SPDX expressions. 

It would be better, to replace the license model in XSD with actual Java classes, which supports better SPDX expressions. 

### Acceptance Criteria
- [x]  License information is stored in pure Java classes which can be created from LicenseInformation items,
- [x] adapt sources and tests to the new license model,
- [x] our classes are build to represent SPDX license expressions,
- [x] the SPDX expressions can be simplified,
- [x] the SPDX expressions can be rendered,
- [ ] make LicenseKnowledgeBaseResolver configurable in the workflow configuration 

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

PR #413, opened 2020-02-03
New license model implementation

bs-ondem:
Fixes https://github.com/eclipse/antenna/issues/409
This PR introduces a new license model with actual Java classes, which replaces the XSD license model. 

### Request Reviewer
@blaumeiser-at-bosch 

### Type of Change
*improvements*

### Checklist
Must:

Optional:


antenna-csv-license-knowledge-base/src/main/java/org/eclipse/sw360/antenna/knowledgebase/csv/CSVBasedLicenseKnowledgeBase.java
```diff
@@ -48,15 +47,15 @@
     private static final String KEY_OSS = "OpenSource";
     private static final String KEY_DELIVER_SRC = "DeliverSources";
     private static final String KEY_DELIVER_LICENSE = "DeliverLicense";
-    private static final String KEY_CLASSIFICATION = "CoveredByINSTStandardProcess";
+    private static final String KEY_CLASSIFICATION = "Classified";
```
blaumeiser-at-bosch: What does classified mean here, isn't it something like Classification?

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+package org.eclipse.sw360.antenna.model.license;
+
+import java.util.*;
+
+public class License extends LicenseInformation {
+    private static final String THREAT_GROUP_KEY = "threatGroup";
+    private static final String CLASSIFICATION_KEY = "classification";
+
+    private String name;
```
blaumeiser-at-bosch: I would prefer to have id and name instead of name and longName. What is now the name is typically the SPDX identifiert or something similar and it is an id, so the code should reflect that.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseInformation.java
```diff
@@ -0,0 +1,24 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.license;
+
+import java.util.List;
+
+public abstract class LicenseInformation {
```
blaumeiser-at-bosch: Why didn't you use interface here?

antenna-csv-license-knowledge-base/src/main/java/org/eclipse/sw360/antenna/knowledgebase/csv/CSVBasedLicenseKnowledgeBase.java
bs-ondem: Yes is something like Classification. I named it classified because it only knew the values ​​yes, no or empty string from the enum `LicenseClassification`. 
Do you have a better name for the column?

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
bs-ondem: I know that, but I wanted to keep the old naming, but if it's okay to change the field/method names, I will do it.

What about the Configuration syntax? Should i keep the old:
```
<license>
   <name/>
   <longName/>
</license>
```
or should i change it to:
```
<license>
   <id/>
   </name/>
</license>
```

blaumeiser-at-bosch: That is more or less an internal naming, so I would prefer to have it changed to the second proposal. Could we somehow handle it in a way that in the tinaConfig we allow both somehow, but this collides with the name field, right? Perhaps we could phrase it fullName, this way we have to disjunctive sets.    Or we keep longName, actually we could use id and longName this would be a minimal change and allow also name to be used in tina config to reference the id.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseInformation.java
bs-ondem: Because i referenced this class in the bindigs.xjb file and the licenseInformation type in the config.xsd is an abstract complexType. 
I will change this abstract class to an interface and introduce an abstract class in the `FromXmlLicenseInformationBuilder` and reference this in the bindings.xjb.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
blaumeiser-at-bosch: I do not see this remark being addressed? Do I get something wrong.

Idea is to name the two properties "id" and "longName" and allow also in the config to map name -> id.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseInformation.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.model.license;
+
+import java.util.List;
+
+public interface LicenseInformation {
+    String evaluate();
+    String evaluateLong();
```
blaumeiser-at-bosch: Here we have the chance to drop the long. I think this should be renamed to evaluateUseNames. Is this method really needed? I would expect that evaluate should be enough, especially since SPDX-Identifier are quite unterstandable.

PR #419, opened 2020-02-10
Get rid of obsolete stuff in antenna

blaumeiser-at-bosch:
Removal of drools policy engine and p2 module. Details see commit comments

### Request Reviewer
@bs-ondem 

### Type of Change

*Other*: Cleanup of obsolete functionality 

### How Has This Been Tested?
All existing tests need to run through green

### Checklist
- [x] Internal issue: https://github.com/bosch-io/osm-antenna/issues/886


blaumeiser-at-bosch:
The failing check of SonarCoud is due to the fact, that the code changed is part of a test framework. Because of Maven this is in the src/main/java hierarchy of the module but the module itself is only a test module used elsewhere for testing. The code coverage seems to not understand that the code is executed in the context of the other tests, but not explicitely tested here.

From my point of view, therefore this is allowable, that the code coverage is lower than expected.

antenna-documentation/src/site/markdown/processors/manifest-resolver.md
```diff
@@ -1,16 +1,17 @@
 ## Manifest resolver
-The Manifest resolver checks the first element of the list of path names for each artifact. If a JAR file is found 
-and it contains a MANIFEST.MF file with the properties `Bundle-SymbolicName` and `Bundle-Version`, it will add a 
-`BundleCoordinates` fact to the artifact. It's deactivated in the default antenna workflow configuration, i.e. it 
-has to be activated in the workflow.xml. 
- 
+The Manifest resolver checks for each artifact if a JAR file is found as fact
+and the jar filecontains a MANIFEST.MF file with the properties `Bundle-SymbolicName` and `Bundle-Version`.
```
bs-ondem: space in filecontains.

blaumeiser-at-bosch: p2 is still supported, but not as good as in the past, so having checks that still find p2 components makes sense.

blaumeiser-at-bosch:
p2 is still supported, but not as good as in the past, so having checks that still find p2 components makes sense.

blaumeiser-at-bosch:
The SonarCloud analysis failure is still acceptable as stated above

The Travis CI failure is a bug in Github Connection to Travis CI. If you follow the link you see that the build indeed was successful after restarting the failing partial build, but the new result was not updated.

blaumeiser-at-bosch:
@bs_ondem: I think I covered all your points, please reevaluate

PR #422, opened 2020-02-11
Updating ort version

neubs-bsi:
This fixes the issue of compatibility with result files created
with the latest master of ort with a new mapping of "has_errors"
to "has_issues"

@sschuberth 

*Type of change*:  version update

### DIFF
--- a/core/frontend-stubs-testing/pom.xml
+++ b/core/frontend-stubs-testing/pom.xml
@@ -161,7 +161,11 @@
             <artifactId>spdx-utils</artifactId>
             <scope>compile</scope>
         </dependency>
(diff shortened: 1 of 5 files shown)


neubs-bsi:
Current State:
Dependency `com.jcraft.jsch.agentproxy.jsch'  used in the ort downloader is not found when running antenna, leading to a `ClassNotFoundException' 
Current Investigations are into checking were the problem with the dependency stems from (ort itself or antenna dependecy problems)

sschuberth:
Superseded by https://github.com/eclipse/antenna/pull/482.

PR #436, opened 2020-02-25
Fix naming issue with coordinates without namespace for SW360 comp name

blaumeiser-at-bosch:
Small bugfix to prevent names with null as part

### Request Reviewer
@neubs-bsi @oheger-bosch 

### Type of Change

*bugfix*

### Checklist
Must:


modules/sw360/src/main/java/org/eclipse/sw360/antenna/sw360/utils/SW360ComponentAdapterUtils.java
```diff
@@ -20,21 +21,25 @@
...
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 public class SW360ComponentAdapterUtils {
     public static String createComponentName(Artifact artifact) {
         return artifact.askFor(ArtifactCoordinates.class)
                 .map(ArtifactCoordinates::getMainCoordinate)
-                .map(o -> Stream.of(o.getNamespace(), o.getName()))
-                .map(o -> o.collect(Collectors.joining("/")))
+                .map(SW360ComponentAdapterUtils::createNameForCoordinate)
```
neubs-bsi: I would like a test for this

Issue #423, opened 2020-02-12
Separate SW360 module into a sw360 client library and an Antenna connector


blaumeiser-at-bosch:
### Summary of the Improvement
In order to make the SW360 module the Java VM based REST Client for SW360, the module has to be separated into a library that does not make use of other Antenna dependencies and the Antenna embedding that connects the SW360 client to the antenna workflow engine.

### Acceptance Criteria
- [x] SW360 module is separated into two components
- [x] 1. component has no dependencies to Antenna and implements the rest client functionality
- [x] 2. component embeds the 1. component into the Antenna workflow engine

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

blaumeiser-at-bosch:
Review 2020-03-05

PR #442, opened 2020-02-26
Oheger bosch/core/#423 http library

oheger-bosch:
Issue #423

This PR creates a new http-support module that provides a generic HTTP client implementation. The idea is to use this client everywhere in Antenna where HTTP requests are executed. The concrete reason for this is the extraction of the SW360 client library, which also needs to send HTTP requests. As the library should be usable outside of the Antenna context, more flexibility is needed:
* While in Antenna all HTTP requests are blocking, for other use cases asynchronous requests may be more suitable.
* The library should minimize dependencies to a concrete HTTP client implementation, as client code may already use a specific one.

The new http-support module contains a set of interfaces defining the API of a simple, asynchronous HTTP client, plus an implementation of these interfaces based on the popular OkHttpClient library. 

As a proof that the new APIs are effective, the new module has been integrated with the SW360 client library, at least as far as the concrete client implementations have been reworked to use the new HTTP client. As part of this integration, the handling of the authorization header (fetching the access tokens) has been reworked; this is now handled internally by the SW360 client library, and callers no longer have to provide initialized headers for each request.

Apart from that the API of the SW360 client library to the outside - in form of the several adapter classes - has not changed. The asynchronous capabilities of the http-support library are not yet exposed; this will be done in another PR.

### Request Reviewer
@neubs-bsi @blaumeiser-at-bosch 

*Type of change*:  
improvements

The new HTTP client implementation is covered by new unit and integration tests based on a mock server. The same applies to the modified client implementations in the SW360 client module.

### Checklist
Must:


modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProviderTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Test
+    public void testAccessTokenIsCached() {
+        CompletableFuture<String> tokenFuture = CompletableFuture.completedFuture(TOKEN);
+        expectTokenRequest(tokenFuture);
+        AccessToken accessToken1 = tokenProvider.obtainAccessToken().join();
+
+        AccessToken accessToken2 = tokenProvider.obtainAccessToken().join();
+        assertThat(accessToken2).isSameAs(accessToken1);
+        verify(authClient).getOAuth2AccessToken();  // exactly once
```
neubs-bsi: I don't know if this comment is necessary, we use this method in this way several time. 
Just a thought though

neubs-bsi:
@oheger-bosch I still need to test this with a working SW360 instance. Will do so this afternoon

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/RequestBuilderImpl.java
```diff
@@ -0,0 +1,174 @@
...
+    public RequestBuilder bodyJson(Object payload) {
+        try {
+            return bodyString(mapper.writeValueAsString(payload), "application/json");
+        } catch (JsonProcessingException e) {
+            throw new HttpExecutionException(e);
+        }
+    }
+
+    @Override
+    public RequestBuilder bodyPart(String name, RequestProducer partProducer) {
```
blaumeiser-at-bosch: I see that is where the multi part builder is used. This looks a bit ugly, it is a lot of complexity for calling a single method, or do I get something wrong here. Is it to fulfill the produceRequest interface?

http-support/src/main/java/org/eclipse/sw360/antenna/http/utils/FailedRequestException.java
```diff
@@ -0,0 +1,83 @@
...
+ * </p>
+ * <p>
+ * This exception class stores the response status code, so that it is
+ * available for further investigation. A cause is not available because the
+ * request completed normally. In order to identify the request that failed, a
+ * tag can be provided that is added to the exception message. This is useful
+ * if a future is constructed from the results of multiple requests.
+ * </p>
+ */
+public class FailedRequestException extends IOException {
```
blaumeiser-at-bosch: Do we use this really, as I wrote I do not like new exception types but you added a semantic here, which I haven't found caught and used anywhere

http-support/src/test/java/org/eclipse/sw360/antenna/http/impl/HttpClientFactoryImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+        HttpClientImpl client = createClient(config);
+        Proxy proxy = client.getClient().proxy();
+        assertThat(proxy).isNotNull();
+        assertThat(proxy.address()).isInstanceOf(InetSocketAddress.class);
+        InetSocketAddress address = (InetSocketAddress) proxy.address();
+        assertThat(address.getHostName()).isEqualTo(host);
+        assertThat(address.getPort()).isEqualTo(port);
+        assertThat(proxy.type()).isEqualTo(Proxy.Type.HTTP);
+    }
+}
```
blaumeiser-at-bosch: Please a newline

http-support/src/test/java/org/eclipse/sw360/antenna/http/impl/HttpClientImplTest.java
```diff
@@ -0,0 +1,172 @@
...
+    @Test
+    public void testCallbackErrorHandling() {
+        CompletableFuture<Object> future = new CompletableFuture<>();
+        IOException exception = new IOException("Request failed miserably");
+
+        Callback callback = httpClient.createCallback(createProcessorMock(), future);
+        callback.onFailure(createCallMock(), exception);
+        expectFailure(future, exception);
+    }
+}
```
blaumeiser-at-bosch: Please add newline

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param token     the current access token
+     * @param <T>       the type of the result generated
+     * @return the result of the {@code ResponseProcessor}
+     */
+    private static <T> ResponseProcessor<T> checkUnauthorized(ResponseProcessor<T> processor, String tag,
+                                                              AccessToken token) {
+        return response -> {
+            try {
+                return processor.process(response);
+            } catch (FailedRequestException e) {
```
blaumeiser-at-bosch: Really necessary, does this specialized exception bring any value?

oheger-bosch:
>     * I am not a big fan of using getter methods within the class were the variable is located, the way it is done in the client adapter classes with the clients. But that is probably just a style choice and as mentioned by you in a conversation will make potential changes easier. The only thing I would give as a consideration is that there are situations were the getters are changed to contain more than just a simple return this.variable to make changes or checks on the variable. While this is also a code style worse discussing this should be kept in the  back of the mind.

This is the result of the _Encapsulate field_ refactoring of the IDE, which I use to apply to make fields accessible for unit tests. Accessing fields only via methods increases flexibility, but this is not a big thing and I have no strong feelings about it. From a performance point of view it probably makes no difference as the JIT compiler will probably inline the direct field access anyway.

> 
>     * the tags used in the clients are only used in the exception handling and not actually necessary for the request, I believe. Is this not a bit much additional for such a small functionality. Or do you think it is important to understanding the exceptions when thrown?

The tags are indeed useful if something goes wrong. As CompletableFutures can be chained to execute multiple requests, it is not always directly obvious, which of the requests failed. They are also logged with debug level before executing a request.

> 
> 
> Things I have seen that have been introduced in preparation of this PR:
> 
>     * there is a new dependency to wiremock, has the license of this been checked?
Wiremock is ASL 2.0 licensed: https://github.com/tomakehurst/wiremock/blob/master/LICENSE.txt

> 
>     * the json files in the test resources of the sw360 client module contain data to an internal sw360 instance. this should probably be changed (probably in a different PR)
Do you think this is a problem? If so, I can change the URLs and internal IDs.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
```diff
@@ -13,7 +13,16 @@
 package org.eclipse.sw360.antenna.sw360.workflow.processors;
 
 import org.eclipse.sw360.antenna.model.artifact.Artifact;
-import org.eclipse.sw360.antenna.model.artifact.facts.*;
```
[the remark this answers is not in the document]
oheger-bosch: It seems that we have different code style settings for imports. In my settings *-imports are disabled. What is the convention for this project?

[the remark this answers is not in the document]
neubs-bsi: so far I think the default settings of intellij: usage >= 5 we use *-imports

neubs-bsi:
> This is the result of the Encapsulate field refactoring of the IDE, which I use to apply to make fields accessible for unit tests. Accessing fields only via methods increases flexibility, but this is not a big thing and I have no strong feelings about it. From a performance point of view it probably makes no difference as the JIT compiler will probably inline the direct field access anyway.

Yeah, in this case we can leave it. 

> The tags are indeed useful if something goes wrong. As CompletableFutures can be chained to execute multiple requests, it is not always directly obvious, which of the requests failed. They are also logged with debug level before executing a request.

When testing I saw that indeed in the debug level those outputs were extremly helpful, so I take back my previous comment. 

> Wiremock is ASL 2.0 licensed: https://github.com/tomakehurst/wiremock/blob/master/LICENSE.txt

That is fine then :)

> Do you think this is a problem? If so, I can change the URLs and internal IDs.

I don't know if it is a problem, but for example the project names of the waivers projects is used and I don't think antenna is the place for this data. I personally would prefer dummy data, especially with urls and IDs and project names.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: We had this discussion in the past. I do not like the .*, but priority had in the past that we do not require any configuration of the editor you are using. I, therefore accepted the unlucky default setup in IntelliJ. But I appreciate every time, where we get rid of a .*!

http-support/src/test/java/org/eclipse/sw360/antenna/http/config/ProxySettingsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    public void testIsProxyUseNullHost() {
+        ProxySettings settings = ProxySettings.useProxy(null, -1);
+
+        assertThat(settings.isProxyUse()).isFalse();
+    }
+
+    @Test
+    public void testIsUseProxyUndefinedPort() {
```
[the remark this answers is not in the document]
oheger-bosch: Fixed

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360ProjectClientIT.java
```diff
@@ -64,33 +68,33 @@ private static void checkTestProjects(Collection<? extends SW360Project> project
...
-        List<SW360Project> projects = projectClient.searchByName(projectName, new HttpHeaders());
+        List<SW360Project> projects = waitFor(projectClient.searchByName(projectName));
         checkTestProjects(projects);
     }
 
     @Test
     public void testSearchByNameNoContent() {
         wireMockRule.stubFor(get(urlPathEqualTo("/projects"))
-                .willReturn(aResponse().withStatus(HttpStatus.SC_NO_CONTENT)));
+                .willReturn(aResponse().withStatus(202)));
```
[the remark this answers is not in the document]
oheger-bosch: Done

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360ProjectClientIT.java
```diff
@@ -64,33 +68,33 @@ private static void checkTestProjects(Collection<? extends SW360Project> project
...
-        List<SW360Project> projects = projectClient.searchByName("foo", new HttpHeaders());
-        assertThat(projects).hasSize(0);
+        extractException(projectClient.searchByName("foo"), IOException.class);
     }
 
     @Test
     public void testSearchByNameError() {
         wireMockRule.stubFor(get(urlPathEqualTo("/projects"))
-        .willReturn(aJsonResponse(HttpStatus.SC_BAD_REQUEST)));
+                .willReturn(aJsonResponse(HttpStatus.SC_BAD_REQUEST)));
```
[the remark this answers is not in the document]
oheger-bosch: Done

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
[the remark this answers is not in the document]
oheger-bosch: Ok, then I can leave it as it is now.

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360LicenseClientIT.java
```diff
@@ -58,32 +61,33 @@ private static void checkLicenses(List<? extends SW360SparseLicense> licenses) {
...
+                expectFailedRequest(licenseClient.getLicenses(), HttpConstants.STATUS_ERR_BAD_REQUEST);
+        assertThat(exception.getTag()).isEqualTo(SW360LicenseClient.TAG_GET_LICENSES);
     }
 
     @Test
-    public void testGetLicenseByName() {
+    public void testGetLicenseByName() throws IOException {
         final String licenseName = "tst";
         wireMockRule.stubFor(get(urlPathEqualTo("/licenses/" + licenseName))
                 .willReturn(aJsonResponse(HttpStatus.SC_OK)
```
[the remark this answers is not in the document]
oheger-bosch: Done

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/SW360ConnectionConfigurationFactory.java
```diff
@@ -10,155 +10,174 @@
...
+     * The name of the configuration property defining the password of the
+     * OAuth 2 client to be passed to the authorization server when requesting
+     * an access token.
+     */
+    public static final String CLIENT_PASSWORD_KEY = "client.password";
+
+    /**
+     * The name of the configuration property that determines whether an HTTP
+     * proxy should be used. Only if this property has a value of
+     * <strong>true</strong>, the configured proxy host and port are used. The
```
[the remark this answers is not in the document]
oheger-bosch: The JavaDocs now mention that the configuration settings are read from the Antenna tool configuration.

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360ClientIT.java
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * Checks whether the given list of projects contains the expected test
+     * data.
+     *
+     * @param projects the list with projects to be checked
+     */
+    private static void checkTestProjects(List<SW360Project> projects) {
+        String[] expectedProjects = {
+                "Central-Waivers_UNKNOWN-License", "Central-Waivers_No-Sources", "SheriffToolTest",
+                "Central-Waivers_No-Source-License"
```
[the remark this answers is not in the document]
oheger-bosch: The test data has been changed; the URLs and IDs in them were replaced by dummy data.

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProviderTest.java
oheger-bosch: I tend to forget this and interpret such constructs as "verify at least once". Normally I do not use this comment. But the purpose of this test is to check that a cached token is not fetched a second time. Therefore, I prefer to make this explicit.

neubs-bsi: that's fine then :)

http-support/src/main/java/org/eclipse/sw360/antenna/http/api/RequestProducer.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+@FunctionalInterface
+public interface RequestProducer {
+    /**
+     * Defines a request to be executed by configuring the
+     * {@code RequestBuilder} passed as argument.
+     *
+     * @param builder the builder to define the request
+     */
+    void produceRequest(RequestBuilder builder);
```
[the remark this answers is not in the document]
oheger-bosch: The goal here is not to have a pure functional design, but to offer a DSL which can be used conveniently by clients. The interface allows constructs like the following one:
```
httpClient.execute(builder -> builder.method(RequestBuilder.Method.POST)
                                .uri(endpointUri())
                                .bodyString(CONTENT, CONTENT_TEXT_PLAIN),
                                ...);
```
This mimics a bit the Kotlin concept of lambdas with receivers, which are also used to update properties of their receiver objects.

But maybe the naming is not ideal for the intended purpose. _produceRequest()_ sounds as if something should be returned. Maybe _defineRequest()_?

http-support/src/main/java/org/eclipse/sw360/antenna/http/config/ProxySettings.java
```diff
@@ -0,0 +1,147 @@
...
+    public static final String UNDEFINED_HOST = "";
+
+    /** Constant representing an undefined port. */
+    public static final int UNDEFINED_PORT = -1;
+
+    private static final ProxySettings EMPTY_SETTINGS = new ProxySettings(UNDEFINED_HOST, UNDEFINED_PORT);
+
+    private final String proxyHost;
+    private final int proxyPort;
+
```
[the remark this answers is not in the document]
oheger-bosch: Proxies with authentication are supported by the underlying OkHttpClient library, but as the implementation is not fully trivial, I would prefer adding this feature in a separate PR.

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/HttpClientImpl.java
```diff
@@ -0,0 +1,117 @@
...
+ * Copyright (c) Bosch.IO GmbH 2020.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.sw360.antenna.http.impl;
```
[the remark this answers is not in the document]
oheger-bosch: Actually,I was not sure whether there will only be a single implementation. A design goal of the HTTP library was to provide an abstraction from a concrete client library. When used outside of the Antenna context it may well be the case that a different HTTP client library is in use, e.g. Apache Http Client or the Java 9+ HTTP client. Then it is at least in theory possible that the interfaces are implemented on top of this other library (as they are not too complex).

However, supporting this scenario would require more work to separate the implementation into another module.

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/RequestBuilderImpl.java
oheger-bosch: In Antenna there is the use case to generate multipart requests for file uploads. Such requests have a part with the binary content of the file and one JSON part with meta data. RequestBuilder needs to support such requests, and the idea was to do this via the _bodyPart()_ method that can simply be called multiple times for each part to be added - rather than calling one of the other methods to set a body.

There is actually a bigger overlap between a builder for a request part and one for a whole request. In both cases you can set headers and the body in different flavours (text, file, JSON).  The builder for a request part does not support setting a request URI, a method, and another body part though; therefore, these methods throw UnsupportedOperationExceptions in this implementation.

I thought I could be lazy and reuse the same interfaces in both cases due to the overlap in the properties that can be defined. But Stephanie commented on this as well; so I will think of an alternative.

http-support/src/main/java/org/eclipse/sw360/antenna/http/utils/FailedRequestException.java
oheger-bosch: In this case, the new exception class brings additional value by storing the status code received from the server. This is used for different purposes in the SW360 client library. For instance, it is checked whether the status code is 401, indicating that the access token has expired. When using a standard exception for this purpose, we would somehow have to parse the status code out of the exception message - which I would really like to avoid.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: see my other comment

http-support/src/main/java/org/eclipse/sw360/antenna/http/api/RequestProducer.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: I do not get your remark, where is produceRequest used in your snippet?

http-support/src/main/java/org/eclipse/sw360/antenna/http/config/ProxySettings.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: Ok

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/HttpClientImpl.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: I do not think that a meta framework is really needed here, right. I think a small use case specific framework that allows to make easy use of a REST API with HATEOAS implemented in the technology we think appropriate would do it for now. Btw, you can still implement this based on a different technology, you are only carrying the dead code in the package, which you do in the current implementation as well, since as you mentioned, the implementation is not in another module.

My principle concern is, that you can get an awful lot of advantages with restricting access to classes if they are packaged right. The common style of making many packages contradicts this advantage and this makes me feel sad.

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/RequestBuilderImpl.java
blaumeiser-at-bosch: Thanks, the code is quite hard to understand as it is.

http-support/src/main/java/org/eclipse/sw360/antenna/http/utils/FailedRequestException.java
blaumeiser-at-bosch: Ok

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Checks whether a request needs to be retried because the access token
+     * may have expired. The method checks whether the request failed with the
+     * special {@code RequestUnauthorizedException} exception.
+     *
+     * @param exception the exception the request failed with if any
+     * @return a flag whether a retry should be attempted
+     */
+    private boolean checkIfRetry(Throwable exception) {
```
blaumeiser-at-bosch: This is a heavy side effect, a check method that invalidates the token.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
```diff
@@ -0,0 +1,263 @@
...
+     * chained to the original future.
+     *
+     * @param producer  the {@code RequestProducer}
+     * @param processor the {@code ResponseProcessor}
+     * @param tag       a tag to identify the request
+     * @param canRetry  a flag whether a retry is possible
+     * @param <T>       the type of the result
+     * @return a future with the result of the request
+     */
+    private <T> CompletableFuture<T> executeWithExpiredTokenCheck(RequestProducer producer,
```
blaumeiser-at-bosch: I am having issues with this method. First the name indicates a special execution of requests, but it is the standard approach to run requests. The original intention of running the request is hard to understand because of this embedding into the token check. I have no idea how to improve, but it took a while until I understood. Also the name of the withConditionalFallback method is from the name not helping to be honest. Perhaps some name like createFutureWrapperForConditionalFallback or something like that.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
blaumeiser-at-bosch: I still do not know whether this accesstoken thing is needed. This is a lot of complexity to transfer the access token through the call stack so that the invalidate side effect is security protected. Why does invalid need this in the first place, because an superfluous invalidate causes a new token to be aquired. So if called slightly to often this creates too much token renewals but so what?

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: Yup, and suggestions for improvements are welcome ;-)

Fact is that we need to manage a state - the current access token. At some point this state has to be updated.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: Sorry, when I see a race condition, I feel the urge to fix it.

The problem will occur in practice if there are many concurrent requests. Without special handling, for each of them the invalidation is done, and a new token is fetched. When this level of concurrency is used performance is certainly an issue. It would be a pity if the library then performs bad.

http-support/src/main/java/org/eclipse/sw360/antenna/http/api/RequestProducer.java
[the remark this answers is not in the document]
oheger-bosch: It is implicitly fulfilled by the lambda expression that defines the request.
But never mind, while reworking the request builder interfaces I got rid off this interface; now generic Consumers are used to define requests.

http-support/src/main/java/org/eclipse/sw360/antenna/http/api/HttpExecutionException.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * <p>
+ * A runtime exception thrown by the Antenna HTTP library if an error during
+ * internal processing occurs.
+ * </p>
+ * <p>
+ * This exception class is used to report errors that are not related to I/O
+ * problems (which are represented by {@code IOException} exceptions).
+ * Typical examples are problems with the serialization or de-serialization of
+ * JSON payload. Such problems should normally not occur when using the library
```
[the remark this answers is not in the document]
oheger-bosch: The exception class was removed. The code now throws an IllegalStateException.

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/HttpClientImpl.java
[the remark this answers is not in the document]
oheger-bosch: The api and impl packages have been removed. The classes are now in the top-level http package.

http-support/src/main/java/org/eclipse/sw360/antenna/http/impl/RequestBuilderImpl.java
oheger-bosch: The builder interfaces have been reworked. The bodies of regular requests and parts of a multipart request are now defined using a new RequestBodyBuilder interface. MultiPartRequestBuilder is no longer needed. I hope this is a clearer design.

http-support/src/test/java/org/eclipse/sw360/antenna/http/impl/HttpClientFactoryImplTest.java
oheger-bosch: Done

http-support/src/test/java/org/eclipse/sw360/antenna/http/impl/HttpClientImplTest.java
oheger-bosch: Done

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: Found an alternative solution. The access tokens are now fully managed by the AccessTokenProvider class. It detects itself when a token is expired; so the code of SW360Client could be simplified. The access token no longer has to be passed to the chain of futures.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: This has been reworked. Token invalidation is now handled by the AccessTokenProvider class.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
oheger-bosch: The methods have been renamed, so that their intension is hopefully clearer.

PR #443, opened 2020-02-27
Remove remaining references to P2 functionality

sschuberth:
P2 functionality was removed in e3316f7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,6 @@ The Antenna project consists of a core and multiple modules in `./modules`, whic
 
 Some of these folders contain their own `README.md`, like e.g. `./modules/sw360/README.md` which contain module specific information.
 
-#### Optional Modules
-Most of the modules are activated and used by default, but the p2-resolver in `./modules/p2/p2-resolver/`, which resolves OSGi sources via P2 repositories, is excluded from the build (since it complicates the build and is unnecessary in most cases).
-To enable it, one can call the corresponding prepare script `./modules/p2/prepareDependenciesForP2.sh` (without Bash support one has to follow the steps in the script by hand).
-To remove the P2 dependencies again you can use the script `./modules/p2/cleanupDependenciesForP2.sh`.
-
-
 ### Install and build Antenna
 
 Please note that some dependencies of SW360antenna are only available for Java 8. So you need to use Java 8 to build the project.
--- a/antenna-documentation/src/site/markdown/dev-guide.md.vm
+++ b/antenna-documentation/src/site/markdown/dev-guide.md.vm
@@ -52,27 +52,11 @@ If your machine is located behind a proxy keep in mind to configure your proxy s
 
 #[[##]]# How to build
 
-If you don't want to build the p2 parts of antenna, just use
+Just use
 ```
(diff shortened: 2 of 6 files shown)


sschuberth:
Thanks for the link, I've restored the `example-projects/p2-example-project` and `example-projects/p2-example-tycho-project` projects. But shouldn't we then also keep `.ci-scripts/test-p2-end2end.sh` as it operates on `example-projects/p2-example-tycho-project`?

bs-ondem:
Yes we should keep the end2end test, but we can remove all parts from the bash scripts regarding the P2 Module e.g. run_all_tests.sh.

.github/workflows/pull_request_build.yml
```diff
@@ -59,9 +59,6 @@ jobs:
     - name: Gradle Frontend Test
       run: bash .ci-scripts/test-ExampleTestProject-with-gradle.sh
 
-    - name: p2Test
```
bs-ondem: We should keep this end2end test.

.travis.yml
```diff
@@ -76,10 +76,6 @@ jobs:
     - name: gradle frontend test
       script: .ci-scripts/test-ExampleTestProject-with-gradle.sh
     
-    # test P2
-    - name: p2Test
-      script: .ci-scripts/test-p2-end2end.sh
```
bs-ondem: We should keep this end2end test.

core/frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testing/AbstractAntennaFrontendTest.java
```diff
@@ -66,8 +66,7 @@ protected void setRunExecutionTest(boolean runExecutionTest) {
                 {(Supplier<AbstractTestProjectWithExpectations>) MinimalTestProject::new, "minimal configuration"},
                 {(Supplier<AbstractTestProjectWithExpectations>) BasicTestProject::new, "basic configuration"},
                 {(Supplier<AbstractTestProjectWithExpectations>) ExampleTestProject::new, "example configuration"},
-                {(Supplier<AbstractTestProjectWithExpectations>) MavenTestProject::new, "maven configuration"},
-                {(Supplier<AbstractTestProjectWithExpectations>) P2TestProject::new, "p2 configuration"}
```
bs-ondem: If we keep the p2 example tests, we should also keep the Test classes for it.

.github/workflows/pull_request_build.yml
sschuberth: Done.

.travis.yml
sschuberth: Done.

core/frontend-stubs-testing/src/main/java/org/eclipse/sw360/antenna/frontend/testing/AbstractAntennaFrontendTest.java
sschuberth: Done.

Issue #437, opened 2020-02-25
Configuration of product name for SW360


blaumeiser-at-bosch:
### Summary of the Bug
There are inconsistencies on how the SW360 name of the project is configured between maven plugin and cli at least. There is a description on page https://eclipse.github.io/antenna/SNAPSHOT/config-configuration.html which proposes to use a config in the antenna config file. This configuration seems to not work in the cli, so the cli still creates a project 1.0. The product name in the tool configuration seems to be used in the Maven plugin, but not in the cli. In general this is not nicely setup

### Steps to reproduce
- see above, experimentation necessary to get the full impact

### Acceptance Criteria
- [x] There should only be one name to be configured for the Disclosure Document and for SW360
- [x] The configuration should not be in the antenna config file, since this is not data but general configuration
- [x] Documentation is done according to the mechanism

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

blaumeiser-at-bosch:
Review 2020-03-05

PR #444, opened 2020-02-28
SW360ProjectCoordinates removed and using tool config values now

neubs-bsi:
Issue: #437 

This PR removes the old way of setting the SW360ProjectCoordinates via the antennaconf.xml all together and instead takes the coordinate names from the tool configuration variables `fullName`and `version`. 

Additionally, since requested in the issue, it also makes it possible that the `AttributionDocumentGenerator` uses the above-mentioned values if no values are set in the workflow steps configuration. 

Tests were adapted and documentation changed. 

@blaumeiser-at-bosch 

*Type of change*:  bug fix

Changed unit tests

### Checklist
Must:


neubs-bsi:
Code Coverage is ignored for this pull request, since the attribution document generator, even though having test is not recognized by sonar cube.

PR #448, opened 2020-03-04
Reduce dependencies

sschuberth:

### DIFF
--- a/modules/attribution-document/attribution-document-generator/src/main/java/org/eclipse/sw360/antenna/attribution/document/workflow/generators/AttributionDocumentGenerator.java
+++ b/modules/attribution-document/attribution-document-generator/src/main/java/org/eclipse/sw360/antenna/attribution/document/workflow/generators/AttributionDocumentGenerator.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.codehaus.plexus.util.FileUtils;
 import org.eclipse.sw360.antenna.api.Attachable;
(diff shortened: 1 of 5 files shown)


modules/attribution-document/attribution-document-generator/src/main/java/org/eclipse/sw360/antenna/attribution/document/workflow/generators/AttributionDocumentGenerator.java
```diff
@@ -110,12 +109,10 @@ public void configure(Map<String, String> configMap) {
 
    private File createWorkDir(Path antennaDir) {
       File workDir = new File(antennaDir.toFile(), WORKING_DIR_NAME);
-      try {
-         FileUtils.forceMkdir(workDir);
-         return workDir;
-      } catch (IOException e) {
-         throw new ConfigurationException("Unable to create working directory in path " + workDir, e);
+      if (!workDir.mkdirs()) {
```
oheger-bosch: Just a remark: This class uses a mix of File and Path APIs and conversions between each other. It would make sense to clean this up and only use one of the two. But this is probably out of scope for this PR. (And this is probably not the only class doing this.)

sschuberth: Good point. Personally, I always find it hard to choose between the `File` and `Path` APIs in Java: I find the former more convenient to use, but the latter offers more modern / performant functionality. In the end, I also often end up with a mix of both APIs :grimacing:

PR #449, opened 2020-03-04
Explain what the flatten plugin is used for

sschuberth:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -551,6 +551,10 @@
             </plugins>
         </pluginManagement>
         <plugins>
+            <!--
+                The flatten plugin is required to resolve the "revision" property used in the version tag of POMs in a
+                multi-module setup, see https://maven.apache.org/maven-ci-friendly.html#Install_.2F_Deploy.
+            -->
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>flatten-maven-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -556,7 +556,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>flatten-maven-plugin</artifactId>
-                <version>1.1.0</version>
+                <version>1.2.1</version>
                 <configuration>
                     <updatePomFile>true</updatePomFile>
                     <flattenMode>resolveCiFriendliesOnly</flattenMode>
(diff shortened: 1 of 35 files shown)


pom.xml
```diff
@@ -551,14 +549,17 @@
...
         </pluginManagement>
         <plugins>
+            <!--
+                The flatten plugin is required to resolve the "revision" property used in the version tag of POMs in a
+                multi-module setup, see https://maven.apache.org/maven-ci-friendly.html#Install_.2F_Deploy.
+            -->
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>flatten-maven-plugin</artifactId>
                 <version>1.1.0</version>
```
oheger-bosch: The most recent version of the flatten-maven-plugin is 1.2.1. It will probably make sense to upgrade the plugin.

pom.xml
```diff
@@ -16,15 +16,13 @@
...
+    <version>${revision}</version>
     <packaging>pom</packaging>
     <name>Antenna</name>
 
     <properties>
         <!-- >This is the main version applied throughout the project. Change this to increase all versions<-->
-        <revision>1.0.0</revision>
-        <!-- >The qualifier can be set to any value and will set the corresponding qualifier in all projects<-->
-        <qualifier>-SNAPSHOT</qualifier>
+        <revision>1.0.0-SNAPSHOT</revision>
```
neubs-bsi: could we then not just not use a properties variable at all and instead use project.version?

pom.xml
sschuberth: Good point, done.

oheger-bosch:
I agree that your changes are in-line with the documentation referenced in the commits. But the comment in the _DependencyTest_ class (and the whole test itself) is still a mystery to me. Do you have any insights here?

sschuberth:
@neubs-bsi FYI, also see [this question](https://stackoverflow.com/q/10582054/1127485) and the accepted answer for a good explanation of the issue.

sschuberth:
> But the comment in the _DependencyTest_ class (and the whole test itself) is still a mystery to me. Do you have any insights here?

Not (yet) really. In a recent call @maxhbr, who added that comment, agreed that it's misleading as "the structure of the repository" as nothing to do with it (it's more about the project being a multi-module project). And maybe the reference to the *flatten* plugin is wrong here, too. Either it should have been the *shade* plugin to begin with, or maybe the whole code is not needed anymore with my recent changes. In fact all of the clean-up work I'm doing has the goal to get rid of this awkward test case and the need to specify transitive dependencies explicitly. We should never need to do that.

PR #453, opened 2020-03-11
Oheger bosch/core/#423 http library integration

oheger-bosch:
This is still related to issue #423.

This PR implements a further integration of the HTTP client library with Antenna. It reworks the places in the codebase, in which Apache HTTP client was used directly, so that now the HTTP client from the HTTP support library is used. The dependencies to Apache HTTP components could be removed.

The AntennaContext class has been extended to store a central HTTP client and JSON object mapper. These objects are thread-safe and can be shared by all Antenna components; they can now be accessed from the context. When setting up an SW360 client the objects from the context are now used.

The compliance tool was affected by these changes as well in the way the connection to SW360 was created. It now creates a partial AntennaContext, from which it can obtain the HTTP client and the JSON mapper.

### Request Reviewer
@neubs-bsi 

*Type of change*:  
improvements

Existing tests have been extended, new tests have been created where necessary. A manual test run with the compliance tool was done to check whether the connection to SW360 could be established.

### Checklist
Must:


modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
```diff
@@ -58,13 +73,13 @@ public void configure(Map<String, String> configMap) {
...
         SW360Connection connection =
-                configurationFactory.createConnection(key -> getConfigValue(key, configMap),
-                        key -> getBooleanConfigValue(key, configMap),
-                        sw360ProxyHost, sw360ProxyPort);
+                getConnectionFactory().createConnection(key -> getConfigValue(key, configMap),
+                        context.getHttpClient(), context.getObjectMapper());
         return new SW360MetaDataReceiver(connection);
     }
+
+    SW360ConnectionConfigurationFactory getConnectionFactory() {
```
neubs-bsi: Getter isn't necessary, if this if only used here

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
```diff
@@ -101,15 +107,38 @@ SW360MetaDataReceiver createMetaDataReceiver(Map<String, String> configMap) {
...
-        verify(toolConfigMock).getProxyHost();
-        verify(toolConfigMock).getProxyPort();
+
+        @SuppressWarnings("unchecked")
+        ArgumentCaptor<SW360ConnectionConfigurationFactory.Getter<String>> captor =
+                ArgumentCaptor.forClass(SW360ConnectionConfigurationFactory.Getter.class);
+        verify(connectionFactory).createConnection(captor.capture(), eq(httpClient), eq(mapper));
+        SW360ConnectionConfigurationFactory.Getter<String> getter = captor.getValue();
+        for (Map.Entry<String, String> e : createStandardConfigMap().entrySet()) {
+            assertThat(getter.apply(e.getKey())).isEqualTo(e.getValue());
```
neubs-bsi: could this ever assert to false? 
and if you are testing the getter.apply method, is this the right place?

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterTest.java
```diff
@@ -77,15 +94,23 @@ public void testConfigure() {
...
-        verify(toolConfigMock, times(1)).getProxyHost();
-        verify(toolConfigMock, times(1)).getProxyPort();
+        @SuppressWarnings("unchecked")
+        ArgumentCaptor<SW360ConnectionConfigurationFactory.Getter<String>> captor =
+                ArgumentCaptor.forClass(SW360ConnectionConfigurationFactory.Getter.class);
+        verify(connectionFactory).createConnection(captor.capture(), eq(httpClient), eq(mapper));
+        SW360ConnectionConfigurationFactory.Getter<String> getter = captor.getValue();
+        for (Map.Entry<String, String> e : configMap.entrySet()) {
+            assertThat(getter.apply(e.getKey())).isEqualTo(e.getValue());
+        }
```
neubs-bsi: could this ever assert to false?
and if you are testing the getter.apply method, is this the right place?

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/ComplianceFeatureUtils.java
```diff
@@ -96,4 +98,85 @@ public static String mapEnvironmentVariable(String value) {
...
+        ToolConfiguration.ConfigurationBuilder configurationBuilder =
+                new ToolConfiguration.ConfigurationBuilder();
+        if (Boolean.parseBoolean(properties.get("proxyUse"))) {
+            configurationBuilder.setProxyHost(properties.get("proxyHost"))
+                    .setProxyPort(Integer.parseInt(properties.get("proxyPort")));
+        }
+        AntennaContext context = new AntennaContext.ContextBuilder()
+                .setToolConfiguration(configurationBuilder.buildConfiguration())
+                .buildContext();
+        return new MappedConfigurationData(properties, context);
```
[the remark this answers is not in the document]
oheger-bosch: Honestly, I do not like this solution myself. I think the basic problem is that the compliance tool is a tool on its own, but should use some core functionality provided by Antenna. This is not only limited to the HTTP client and the object mapper but includes things like logging and the handling of certain default command line options. So there needs to be a balance between code reuse and duplication.

The approach to reuse the whole context is obviously not satisfying as it has too many other dependencies. I will try to come up with another solution; but it won't be easy to find the right balance.

core/runtime/pom.xml
```diff
@@ -108,6 +108,10 @@
             <artifactId>system-rules</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock-jre8</artifactId>
```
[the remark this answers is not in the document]
oheger-bosch: The scope is inherited from the dependency management section in the top-level pom.

assembly/gradle-plugin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Mar 27 15:55:39 UTC 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-bin.zip
```
sschuberth: Any reason not to go straight to the current version 6.2.2?

oheger-bosch: I was thinking that using an older version might give us some insights about the minimum gradle version that is supported. (However, you are right that version 5.1.1 is a bit arbitrary.)

blaumeiser-at-bosch: If it as an effect on the usability within a customer Gradle build, the older the version the better. It is not a good habit to push our customers to use the latest Gradle only to run the compliance tooling.

sschuberth: It has no effect on the customer. This is just the Gradle version used to build the Gradle plugin.

oheger-bosch: Probably, as long as the Gradle plugin only uses the stable Gradle API, it will be compatible with multiple Gradle versions, even older ones - and I think that our plugin only uses a rather small part of the API.

However, if my investigations of the Kotlin incompatibility are correct, the dependencies the Antenna plugin drags in restrict the Gradle version that can be used. So because of our dependency to OkHttpClient and (transitive) Kotlin 1.3, the plugin will not be compatible with Gradle 4.x.

sschuberth: Ah, I was indeed mistaken that the Gradle version used for building the plugin would have no influence on the consumer, because of the `gradleApi()` call [here](https://github.com/eclipse/antenna/blob/1502d6ef07abcae23c9179c5189cb06708fe5b62/assembly/gradle-plugin/build.gradle#L15) (i.e. the plugin is built against the API of the Gradle version used for building). I somehow thought we'd build the Gradle plugin with pure Maven (as explained [here](https://stackoverflow.com/a/27556212/1127485)), but in that case we wouldn't need the Gradle wrapper at all anyway.

blaumeiser-at-bosch: We once had a bigger issue when we updated the Maven dependency and got feedback on projects using quite old versions. But moving towards Gradle 5.1.1 as base version sounds okish, especially when we have good reason like a technical dependency.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/ComplianceFeatureUtils.java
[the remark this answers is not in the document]
blaumeiser-at-bosch: From my point of view, there is no big issue since the compliance tool is also an Antenna based tooling. Each frontend can of course use the basic infrastructure provided by Antenna.

But of course having a very small need and dragging in a whole framework typically does not feel good with good reason, so I would appreciate if there is a simpler solution, but I would also accept, if you say this is conceptually the right way to go.

core/model/src/main/java/org/eclipse/sw360/antenna/api/configuration/AntennaContext.java
```diff
@@ -41,6 +60,9 @@ private AntennaContext(ContextBuilder builder) {
         this.contextExtension = builder.contextExtension;
 
         this.processingReporter = builder.processingReporter;
+
+        objectMapper = builder.getObjectMapper();
+        httpClient = builder.createHttpClient();
```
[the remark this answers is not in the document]
blaumeiser-at-bosch: I still try to find out who this guy was, that proposed this this. thing. I agree to be consistent, but would rather like to see no this.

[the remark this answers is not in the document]
oheger-bosch: Code is now uniform with the other fields.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/ComplianceFeatureUtils.java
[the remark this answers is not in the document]
oheger-bosch: This part has been reworked. The AntennaContext is no longer used by the compliance tool. Rather, the logic to create shared service objects has been extracted into a new ServiceFactory class. This one is now used by the compliance tool, which is a significant simplification.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/ComplianceFeatureUtils.java
```diff
@@ -96,4 +98,85 @@ public static String mapEnvironmentVariable(String value) {
...
+    }
+
+    /**
+     * A class representing configuration data for the compliance tool that has
+     * already been processed.
+     */
+    public static class MappedConfigurationData {
```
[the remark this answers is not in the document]
oheger-bosch: The class no longer exists.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360Updater.java
```diff
@@ -80,4 +88,8 @@ private String getProjectName() {
     public void setUpdaterImpl(SW360UpdaterImpl updaterImpl) {
         this.updaterImpl = updaterImpl;
     }
+
+    SW360ConnectionConfigurationFactory getConnectionFactory() {
+        return connectionFactory;
+    }
```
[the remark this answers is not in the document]
oheger-bosch: It is used by tests.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360Enricher.java
oheger-bosch: It is used by tests.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/processors/SW360EnricherTest.java
oheger-bosch: What is tested here is whether a correct getter for configuration settings has been passed to the connection factory.  The factory uses this getter to read the configuration properties. So it is checked whether these read operations yield the expected values.

I don't know why this mechanism with the Getter objects is used; I did not change this.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/workflow/generators/SW360UpdaterTest.java
oheger-bosch: see my other comment

assembly/gradle-plugin/gradle/wrapper/gradle-wrapper.properties
sschuberth: Any objection then @blaumeiser-at-bosch if I also move the Gradle version used for our dependencies from 4.1 to 5.1.1 to align?

blaumeiser-at-bosch: Please do so

At @bs-ondem: We should keep this somewhere, so that we point that out in the release notes.

core/model/src/main/java/org/eclipse/sw360/antenna/api/service/ServiceFactory.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * Application of the Initialization-on-demand holder idiom for the JSON
+     * object mapper.
+     */
+    private static class LazyMapperHolder {
+        private static final ObjectMapper MAPPER = new ObjectMapper()
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    }
```
neubs-bsi: I think this class is superfluous and only used once (if I haven't missed an occurrence) and could just be in the "getObjectMapper()" method.

oheger-bosch: This is tricky, but the class plays indeed an important role for the safe initialization of the ObjectMapper. This is the so-called [Initialization-on-demand holder idiom](https://de.wikipedia.org/wiki/Initialization-on-demand_holder_idiom)

Issue #421, opened 2020-02-10
Clean up & Improvement of License Knowledge Base Resolver


bs-ondem:
### Summary of the Improvement
After the discussion with @blaumeiser-at-bosch about https://github.com/bosch-io/osm-antenna/issues/886, we decided to clean up and improve the `LicenseKnowledgeBaseResolver` and all existing `ILicenseManagementKnowledgeBase` implementations. 

### Acceptance Criteria
- [x] Make `LicenseKnowledgeBaseResolver` work with multiple priority based `ILicenseManagementKnowledgeBase` implementations. 
- [x] Extend the `ILicenseManagementKnowledgeBase` interface with new method, which returns the unique identifier of the implementation
- [x] The `LicenseKnowledgeBaseResolver` has a new config entry for the configuration of the prioritised `ILicenseManagementKnowledgeBase` implementation.
- [x] ~~Move the one method of `LicenseKnowledgeBaseFactory` to `LicenseKnowledgeBaseResolver` and delete unnecessary factory class.~~
- [x] Move `CSVBasedLicenseKnowledgeBase` to core-workflow-steps and delete it on root. 

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

PR #456, opened 2020-03-11
CleanUp & Improve LicenseKnowledgeBaseResolver

bs-ondem:
Fixes https://github.com/eclipse/antenna/issues/421 
Summary of PR in the issue above. 

### Request Reviewer
wip

### Type of Change
*improvements*

### Checklist
Must:


assembly/configuration/src/main/resources/workflow.xml
```diff
@@ -37,6 +37,9 @@
         <step>
             <name>License Knowledgebase Resolver</name>
             <classHint>org.eclipse.sw360.antenna.workflow.processors.LicenseKnowledgeBaseResolver</classHint>
+            <configuration>
```
blaumeiser-at-bosch: I do not think, the standard workflow should state a configuration here. Instead the standard priorities should be followed.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolver.java
```diff
@@ -45,104 +44,40 @@ public LicenseKnowledgeBaseResolver(ILicenseManagementKnowledgeBase knowledgeBas
...
+        }
+
+        List<ILicenseManagementKnowledgeBase> licenseManagers = knowledgeBaseSupplier.get();
+
+        this.knowledgeBase = Optional.ofNullable(knowledgeBase)
+                .orElseGet(() -> licenseManagers.stream()
+                        .filter(l -> chosenManager.equals(l.getId()))
+                        .findFirst()
+                        .orElseGet(() -> licenseManagers.stream()
+                                .sorted(Comparator.comparing(ILicenseManagementKnowledgeBase::getPriority).reversed())
```
blaumeiser-at-bosch: The idea is that the CSV knowledgebase should check whether it can provide information, i.e., whether there is a csv file giving license information. If this is not the case, it should not be used as license knowledgebase. So Antenna chooses the knowledge bases according to the priority which is csv knowledgebase first and then the SPDX knowledgebase, but without a csv configuration the csv knowledgebase has to indicate somehow, that it cannot be used, so here the natural choice would be the SPDX knowledgebase. My proposal is that the getPriority method returns a value < 0 to indicate that the knowledgebase cannot be used.

assembly/spdx-license-knowledge-base/pom.xml
```diff
@@ -73,6 +73,22 @@
             <artifactId>jsonld-java</artifactId>
```
blaumeiser-at-bosch: Sorry, that I come up with this comment so late, but I think we should also move the spdx license knowledgebase to the core workflow steps and get rid of another module.

PR #457, opened 2020-03-12
Remove Spring references from POMs

sschuberth:
The last dependency on Spring was removed in 8890038.


oheger-bosch:
Seems that the build is broken :-(

PR #459, opened 2020-03-12
HttpHelperTest: Download to a different location that the source file

sschuberth:
Previously, the downloaded file was written to the same location from
where the file was served. This coincidentally used to work before
63781ba because FileUtils.copyInputStreamToFile() truncated an existing
file instead of deleting it like Files.copy() does, but with 63781ba the
test started to fail on Windows.

Also, the created test file was empty, and the final assertion only
compared the file paths, not their contents.

Improve this by

- using separate directories for source and target files,
- writing some content to the test file,
- compare the contents of source and target files.


sschuberth:
@oheger-bosch, I know `HttpHelperTest` will get rewritten by your https://github.com/eclipse/antenna/pull/453, but I wanted to make this quick-fix anyway as an intermediate step to document my findings.

PR #465, opened 2020-03-24
Fix master

blaumeiser-at-bosch:

There was an issue in the example project.

### Request Reviewer
@sschuberth 

### Type of Change

*Type of change*: Bugfix 

### How Has This Been Tested?
Current testing

### Checklist
Must:


example-projects/example-project/pom.xml
```diff
@@ -122,11 +122,6 @@
                 </executions>
                 <dependencies>
                     <!-- The antenna mojo requires the kbase to fetch the license information -->
```
sschuberth: If the `spdx-license-knowledge-base` dependency is removed, then this comment should be removed too, or?

PR #472, opened 2020-03-25
Build the Antenna Gradle plugin with Maven instead of Gradle run

### LINKED ISSUE #473: Build Gradle Plugin with Maven means
Build Antenna Gradle Plugin with Maven instead of an embedded Gradle build, ensure that tests are still running, especially the acceptance tests using the Gradle Plugin.

### Acceptance Criteria
- [ ] Build Gradle Plugin with Maven
- [ ] Ensure that docu and acceptance testing is still working

### Definition of Done
- Documentation is updated

blaumeiser-at-bosch:
Removed the Gradle call from the build of the gradle plugin and build it with maven now. Removed also the gradle build infrastructure from the plugin.

Adapted the tests to run as far as possible. One test is lost because testing gradle plugins is restricted in a Maven environment.

Changed the example project to be able to run with a self contained gradle wrapper

### Request Reviewer
@sschuberth 

### Type of Change
*Type of change*:  CI

Fixes #473 

### How Has This Been Tested?
Basic regression testing by build including the gradle acceptance test in the build

### Checklist
Must:


core/frontend-stubs/maven-frontend-stub/src/test/java/org/eclipse/sw360/antenna/frontend/stub/mojo/DependencyTest.java
```diff
@@ -62,6 +62,7 @@
             dependency("org.gradle", "gradle-base-services"),
             dependency("org.gradle", "gradle-core-api"),
             dependency("org.gradle", "gradle-logging"),
+    	    dependency("org.gradle", "gradle-base-services-groovy"),
```
sschuberth: This (and all other `gradle-base-services-groovy` occurrences) should not be needed anymore (I've just recently removed them in the upgrade to Gradle 5.6).

blaumeiser-at-bosch: I explicitely had to add this due to a build problem 🤔 . The build told me that this dependency is missing.

blaumeiser-at-bosch:
Code coverage of 0% is ok for this change, since only one line of productive code is affected and in this line a string replacement was done.

core/frontend-stubs/maven-frontend-stub/src/test/java/org/eclipse/sw360/antenna/frontend/stub/mojo/DependencyTest.java
sschuberth: As CI passed for my PR, could you please double-check you've got a clean build?

blaumeiser-at-bosch: Whatever happened locally, now I do not need it. You were right, change pushed

assembly/gradle-plugin/src/main/java/org/eclipse/sw360/antenna/frontend/gradle/AntennaGradleTask.java
```diff
@@ -14,6 +15,6 @@
 
 public class AntennaGradleTask extends AnalyzeTask {
     protected String getPluginDescendantArtifactIdName() {
-        return "antenna-maven-plugin";
+        return "gradle-plugin";
```
sschuberth: Is this renaming necessary as part of the Gradle to Maven conversion? If so, why?

assembly/gradle-plugin/src/test/java/org/eclipse/sw360/antenna/frontend/gradle/AntennaGradlePluginTest.java
```diff
@@ -81,6 +80,7 @@ public void testWithoutGradle() {
         assertThat(projectRoot.resolve("antenna").toFile()).exists(); // TODO
     }
 
+    @Ignore("Gradle plugin tests without a Gradle running have issues")
```
sschuberth: What *kind* of issues?

assembly/gradle-plugin/src/test/resources/pom.xml
```diff
@@ -80,14 +81,6 @@
                         <phase>package</phase>
                     </execution>
                 </executions>
-                <dependencies>
```
sschuberth: Why is this not required anymore?

blaumeiser-at-bosch: This dependency is not existing anymore, the functionality has been integrated by Onur into the core workflow steps.

assembly/gradle-plugin/src/test/java/org/eclipse/sw360/antenna/frontend/gradle/AntennaGradlePluginTest.java
blaumeiser-at-bosch: I remove the test and the other ignored test, there is no value in ignored tests, since they can always be added again out of version control. Both tests were not sceduled to be fixed soon. We have principle tests that execute the gradle task in the same way as gradle would run the task and the pricinple gradle test is executed as acceptance test in the build, so there is no test gap that is created by the removal.

assembly/gradle-plugin/src/main/java/org/eclipse/sw360/antenna/frontend/gradle/AntennaGradleTask.java
blaumeiser-at-bosch: I'll isolate this change into a second commit which does the name fix.

PR #477, opened 2020-03-30
cyclone-dx: Get rid of the (undeclared) dependency on Apache Commons

sschuberth:


blaumeiser-at-bosch:
Code coverage 0.0 is ok since change does not really change any semantics

PR #481, opened 2020-03-31
Improvements on the license model implementation

blaumeiser-at-bosch:
#### Major changes:
- Add method to return in LicenseStatement the next level of the expression, this enables the traversal through a license expression
- Reimplement the semantics of a with license with details:
  - Change the semantics to represent one license as WithLicense with two parts
  - Base WithLicense on a single license, all methods of License are adapted to meet the needs of a with license
  - The basic concept of a with license is that two SPDX identifier are matched by the term ' WITH ', this is also true for the name of the license not only the id
  - The license text is also attached to each other in the form of license text with exception: exception text and corresponding newlines
  - WithLicense can be built from two parts reflecting original license and exception and from already combined terms, e.g., 'licenseID WITH exceptionID
    - Corresponding Constructors and setters have been defined, getters are added for both as convenience 
  - Tests are added to check for the new behavior
  - Adapt places in the code that requires a change due to the changed understanding of WithLicense

#### Minor changes:
- Replace use of List by use of Collection, because a LicenseStatement is a commutative operation, so order must not be important
- Improve parameter handling to ensure non-null parameters, this prevents NullPointerExceptions in use due to immediate feedback
- Small improvement of local stream handling, using the joining collector with pre- and suffix
- Adaptations of code places where interface changes mattered

### Request Reviewer
@neubs-bsi @sschuberth 

### Type of Change

### How Has This Been Tested?
Unittests have been added, tests have been adapted to reflect the changed semantics

### Checklist
Must:

- [x] I have updated the documentation accordingly to my changes as javadoc

### DIFF
--- a/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolverImpl.java
+++ b/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolverImpl.java
@@ -105,16 +105,16 @@ public class LicenseKnowledgeBaseResolverImpl {
     private void setThreatGroup(License license) {
         Optional<String> threatGroupOfLicense = license.getThreatGroup();
         if (!threatGroupOfLicense.isPresent() || threatGroupOfLicense.get().isEmpty()) {
-            String threatGroupOfKb = this.knowledgeBase.getThreatGroupForId(license.getId());
-            license.setThreatGroup(threatGroupOfKb);
+            Optional.ofNullable(this.knowledgeBase.getThreatGroupForId(license.getId()))
+                    .ifPresent(license::setThreatGroup);
         }
     }
 
     private void setClassification(License license) {
         Optional<String> classificationOfLicense = license.getClassification();
         if (!classificationOfLicense.isPresent() || classificationOfLicense.get().isEmpty()) {
-            String classificationOfKb = this.knowledgeBase.getClassificationById(license.getId());
-            license.setClassification(classificationOfKb);
+            Optional.ofNullable(this.knowledgeBase.getClassificationById(license.getId()))
+                    .ifPresent(license::setClassification);
         }
     }
 }
--- a/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolverImpl.java
+++ b/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolverImpl.java
@@ -46,7 +46,7 @@ public class LicenseKnowledgeBaseResolverImpl {
                 .flatMap(List::stream)
                 .map(ArtifactLicenseInformation::get)
                 .map(LicenseInformation::getLicenses)
-                .flatMap(List::stream)
+                .flatMap(Collection::stream)
                 .forEach(license -> {
                     aliasToIdentifier(license);
                     setText(license);
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
@@ -17,6 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+/**
+ * Helper class for creation of license objects out of config xml statements, not for normal use outside
+ * of config parsing.
+ */
 public class FromXmlLicenseInformationBuilder {
     public interface ILicenseInformationBuilder {
         LicenseInformation build();
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
@@ -114,7 +114,7 @@ public class FromXmlLicenseInformationBuilder {
 
         @Override
         public LicenseInformation build() {
-            return new WithLicense((License) license.build(), (License) exception.build());
+            return new WithLicense(license.id, license.commonName, license.text, exception.id, exception.commonName, exception.text);
         }
     }
 }
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
@@ -68,11 +68,14 @@ public class License implements LicenseInformation {
         if (properties == null) {
             return new HashMap<>();
         }
-        return properties;
+        return Collections.unmodifiableMap(properties);
     }
 
     public void setProperties(Map<String, String> properties) {
-        this.properties = properties;
+        this.properties = new HashMap<>();
+        if (properties != null) {
+            this.properties.putAll(properties);
+        }
     }
 
     public void setProperty(String key, String property) {
--- a/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
+++ b/core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
@@ -22,22 +22,25 @@ public class License implements LicenseInformation {
     private Map<String, String> properties;
 
     public License() {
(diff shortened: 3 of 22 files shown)


core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/processors/LicenseKnowledgeBaseResolverImpl.java
```diff
@@ -46,7 +46,7 @@ public LicenseKnowledgeBaseResolverImpl(ILicenseManagementKnowledgeBase knowledg
                 .flatMap(List::stream)
                 .map(ArtifactLicenseInformation::get)
                 .map(LicenseInformation::getLicenses)
-                .flatMap(List::stream)
+                .flatMap(Collection::stream)
```
sschuberth: Just a general comment: Using bullet points in a commit message is a good hint that you're probably doing more in a single commit than you should do. If you need to enumerate things, move each point into a separate commit instead.

core/model/src/test/java/org/eclipse/sw360/antenna/model/license/LicenseTest.java
```diff
@@ -39,6 +39,15 @@ public void testLicenseWithOnlyId() {
...
 
+    @Test
+    public void testEqualLicense() {
+        License license1 = new License ("Apache-2.0");
+        License license2 = new License ("Apache-2.0", "Apache License Version 2.0", "Some license text");
+        License license3 = new License ("MIT");
+
+        assertThat(license1.equalLicense(license2)).isTrue();
+        assertThat(license1.equalLicense(license3)).isFalse();
+    }
```
sschuberth: Nit: Missing blank line after this one.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
```diff
@@ -114,7 +114,7 @@ public void setException(LicenseBuilder exception) {
 
         @Override
         public LicenseInformation build() {
-            return new WithLicense((License) license.build(), (License) exception.build());
+            return new WithLicense(license.id, license.commonName, license.text, exception.id, exception.commonName, exception.text);
```
sschuberth: A general remark: We should watch out to not re-implement too much of either https://github.com/oss-review-toolkit/ort/tree/master/spdx-utils/src/main/kotlin or https://github.com/spdx/Spdx-Java-Library/tree/master/src/main/java/org/spdx/library/model/license here.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
```diff
@@ -23,83 +34,167 @@
...
 
+    /**
+     * Return the value of a property or an empty Optional.
+     *
+     * @param key The key of the queried property, must not be null.
+     * @return The property value stored or an empty Optional.
+     */
     public Optional<String> getProperty(String key) {
-        return Optional.ofNullable(properties.get(key));
+        return Optional.ofNullable(properties.get(requireNonNull(key)))
```
sschuberth: This is more than just adding comments.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
```diff
@@ -130,8 +225,15 @@ public boolean equals(Object o) {
...
+    /**
+     * Compares the license to another license. The licenses are equals, if the license ids are the same,
+     * independent whether one license is complete or not.
+     *
+     * @param license The license to compare this license to.
+     * @return True, if both licenses have the same license id.
+     */
     public boolean equalLicense(License license) {
-        return Objects.equals(evaluate(), license.evaluate());
+        return Objects.equals(getId(), license.getId());
```
sschuberth: This is more than just adding comments.

core/model/src/test/java/org/eclipse/sw360/antenna/model/license/WithLicenseTest.java
```diff
@@ -88,6 +88,33 @@ public void testWithLicenseBuildFromAssembledInput() {
         assertThat(withLicense1.equalLicense(withLicense2)).isTrue();
     }
 
+    @Test(expected=IllegalArgumentException.class)
```
sschuberth: A general remark: According to the commit message, only this file should be changes, and all changes in other files should be separated out to other commit with proper commit messages.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
blaumeiser-at-bosch: I do not get your point, the commit message also says, fix code smells, the fixes below are doing that, fixing the code smells that were reported by Sonarqube.

core/model/src/test/java/org/eclipse/sw360/antenna/model/license/WithLicenseTest.java
blaumeiser-at-bosch: See my answer above, the other stuff were code smells reported by SonarQube which I fixed as the commit message says. You could argue, that both things could be separated into two commits, but I do not see here the advantage of having this separated, because the changes are not really changing a lot.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/License.java
blaumeiser-at-bosch: I'll change this.

core/model/src/test/java/org/eclipse/sw360/antenna/model/license/WithLicenseTest.java
sschuberth: The advantage of separating semantically different changes into different commits for me as a reviewer is that I can skip over commits that say to change things which are either unimportant to me, or that are things I cannot comment on.

So, if a commit says to just add a test, that hardly breaks the main app, and I might skip the review if I'm short of time. Similarly, if a commit says to just add comments, that is a non-functional change which I can just skim over and proceed to the next commit.

However, for that to work I need to be able to rely on the commit message to really only change what is stated here. At least when the commit comes from a trusted contributor. For untrusted parties I of course review everything, thoroughly :-)

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/FromXmlLicenseInformationBuilder.java
blaumeiser-at-bosch: In principle I agree, but there has to be a step-by-step move towards the intended solution. In the meanwhile, we need this step to continue with "smaller" steps.

blaumeiser-at-bosch:
@sschuberth This is it, no commit found, that could be smaller.

core/model/src/main/java/org/eclipse/sw360/antenna/model/license/LicenseStatement.java
```diff
@@ -10,10 +10,7 @@
  */
 package org.eclipse.sw360.antenna.model.license;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
sschuberth, line 13: Nit: The commit message just repeats in plain word what's obvious from the diff, but lacks context *why* this change makes sense.

core/runtime/src/test/java/org/eclipse/sw360/antenna/configuration/ConfigurationTest.java
```diff
@@ -62,7 +62,7 @@ public void test() {
 
         ArtifactIdentifier identifier = new ArtifactFilename("overrideName.jar");
         assertThat(configuration.getFinalLicenses().keySet().stream().anyMatch(k -> k.matches(identifier))).isTrue();
-        List<LicenseInformation> list = configuration.getFinalLicenses().entrySet().stream()
+        List<License> list = configuration.getFinalLicenses().entrySet().stream()
```
sschuberth, line 65: To the reader who's not so familiar with the code base (like me) it's surprising that this actually just works. It would be nice to read in the commit message *why* this is a fix, and why it "worked" (i.e. compiled) also before, but what's better now.

sschuberth:
> @sschuberth This is it, no commit found, that could be smaller.

Thank you, this is *much* more pleasant to review now!

Issue #480, opened 2020-03-31
SW360 204 No_Content Responses are processed as a succesfull request and causes Exception


neubs-bsi:
### Summary of the Bug
[A SW360 commit](https://github.com/eclipse/sw360/commit/bd996419b959fd57a17080290323f1d86529316c#diff-fc8bf46fc1e710f969b71bcd6141aacf) introduced a `204 NO_CONTENT` response to some request, causing Antenna to wrongly process it as successful since it only checks if the response is 200...300, as is usual for the [okhttp3.isSuccessfull check](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/is-successful/) 

### Steps to reproduce
- Run antenna with a SW360 instance with no projects.

### Acceptance Criteria
- [ ] Response needs to be properly checked and reacted to
- [ ] No Exception is caused
- [ ] All changed request responses of above mentioned commit are checked and changed 

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

neubs-bsi:
> <@U6L2MTT71> Unfortunately the latest master of Antenna still has the same/similar issues. When running against SW360 8.0.1, SW360 returns a response with status code 204 (NO_CONTENT) with a null body. Antenna tries to process this body and throws an exception. What it should be doing is recognising a null body as “no projects match that name” and trying to create a project instead.
```<https://sw360.verifa.io/resource/api/projects?name=SpringBootExampleProject>
get_projects_by_name
HttpUtils:checkResponse(...): response.statusCode(): 204
HttpUtils:checkResponse(...): response.body().available(): 0
[ERROR] Failed HTTP request GET <https://sw360.verifa.io/resource/api/projects?name=SpringBootExampleProject>
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (okio.RealBufferedSource$inputStream$1); line: 1, column: 0]
    at com.fasterxml.jackson.databind.exc.MismatchedInputException.from (MismatchedInputException.java:59)
    at com.fasterxml.jackson.databind.ObjectMapper._initForReading (ObjectMapper.java:4360)
[...]

PR #484, opened 2020-04-08
Oheger bosch/sw360/bug 480 no content handling

oheger-bosch:
Issue #480 
This PR deals with changes in the SW360 REST API related to status codes for query requests: If a request does not yield results, the response has status code 204 (no content) and no response body.

The solution assumes that a no content response is valid only for queries that result in empty lists. Other requests, e.g. POSTs for creating new entities, should always return a body. Therefore, the solution does not implement a central handling of the 204 status, but there is now a new method in SW360Client for sending requests which checks for the 204 response status and returns a default object (typically an empty list) in this case. The concrete SW360 client implementations are responsible of using the correct method for their queries; they have been updated accordingly in this PR.

@neubs-bsi 

*Type of change*:  
bug fix

### How Has This Been Tested?
For each query of a client that may yield a 204 response, a test case has been added.

### Checklist
Must:

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
@@ -23,6 +23,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import static org.eclipse.sw360.antenna.http.utils.HttpConstants.URL_PATH_SEPARATOR;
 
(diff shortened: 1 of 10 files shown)


modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360Client.java
```diff
@@ -166,7 +193,7 @@ protected String resourceUrl(String... paths) {
 
         CompletableFuture<T> futRequest = getTokenProvider().doWithToken(accessToken ->
                 getClientConfig().getHttpClient().execute(accessToken.tokenProducer(producer),
-                HttpUtils.checkResponse(processor, tag)));
+                        HttpUtils.checkResponse(processor, tag)));
```
blaumeiser-at-bosch, line 196: Small thing: This seems to be a reformatting issue

oheger-bosch: Fixed.

PR #488, opened 2020-04-21
Allow unverified access to ssl resource

blaumeiser-at-bosch:
Needed for example to access SW360 in a local instance that has a self certificate.

### Request Reviewer
@oheger-bosch 

### Type of Change
*Type of change*:  Improvement

### How Has This Been Tested?
Additional test has been added.

### Checklist
Must:

### DIFF
--- a/http-support/src/test/java/org/eclipse/sw360/antenna/http/HttpClientFactoryImplTest.java
+++ b/http-support/src/test/java/org/eclipse/sw360/antenna/http/HttpClientFactoryImplTest.java
@@ -74,4 +74,19 @@ public class HttpClientFactoryImplTest {
         assertThat(address.getPort()).isEqualTo(port);
         assertThat(proxy.type()).isEqualTo(Proxy.Type.HTTP);
     }
+
+    @Test
+    public void testNewClientWithoutCertificateCheck() {
+        try {
+            System.setProperty(HttpClientFactoryImpl.CLIENT_ACCESS_UNVERIFIED_PROPERTY, "true");
+            ObjectMapper mapper = mock(ObjectMapper.class);
+            HttpClientConfig config = HttpClientConfig.basicConfig()
+                    .withObjectMapper(mapper);
+
+            HttpClientImpl client = createClient(config);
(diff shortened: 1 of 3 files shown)


oheger-bosch:
This is for sure a nice addition to the library.
However, turning off the verification globally could be risky. How about passing a hostname verifier function into the HttpClientConfig? Then the checks could be enabled or disabled on a more fine-granular base. We could also provide some standard verifier functions, e.g. one that actually disables verification or one that only handles specific hosts.

blaumeiser-at-bosch:
> This is for sure a nice addition to the library.
> However, turning off the verification globally could be risky. How about passing a hostname verifier function into the HttpClientConfig? Then the checks could be enabled or disabled on a more fine-granular base. We could also provide some standard verifier functions, e.g. one that actually disables verification or one that only handles specific hosts.

In principle you are right. But this is a clear debug feature for local settings. So it should not be enabled in any productive setting. I think that it is too much to provide a fine granular setup because it is either a problem locally or not. A fine granular feature would only make sense, if the library is used in the same tool for different purposes and only one of the servers for the testing application has a self signed certificate. Perhaps we should add an issue for this and work on it if someone sees this as an issue. What do you think?

http-support/src/test/java/org/eclipse/sw360/antenna/http/HttpClientFactoryImplTest.java
```diff
@@ -74,4 +74,15 @@ public void testNewClientWithProxySettings() {
         assertThat(address.getPort()).isEqualTo(port);
         assertThat(proxy.type()).isEqualTo(Proxy.Type.HTTP);
     }
+
+    @Test
+    public void testNewClientWithoutCertificateCheck() {
+        System.setProperty(HttpClientFactoryImpl.CLIENT_ACCESS_UNVERIFIED_PROPERTY, "true");
```
oheger-bosch, line 80: The system property should be unset after the test - best in a try-finally-block -; otherwise it might impact other tests, as this is a VM-wide setting.

oheger-bosch:
Ok, fine with me, except for this minor nit in the test case.

Issue #494, opened 2020-04-28
SW360 client: uploading attachments causes 500 Respone due to duplicate attachment


neubs-bsi:
### Summary of the Bug
When update_release is deactivated in the SW360Updater the following error is thrown in the example-project run. 

```
[ERROR] Failed HTTP request POST http://localhost:8880/resource/api/releases/703df9c3bdb1ac909ddbb7544104f979/attachments
org.eclipse.sw360.antenna.http.utils.FailedRequestException: The request 'post_upload_attachment' failed with status code 500
    at org.eclipse.sw360.antenna.http.utils.HttpUtils.lambda$checkResponse$0 (HttpUtils.java:143)
    at org.eclipse.sw360.antenna.http.HttpClientImpl$1.onResponse (HttpClientImpl.java:87)
    at okhttp3.RealCall$AsyncCall.run (RealCall.kt:140)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:624)
    at java.lang.Thread.run (Thread.java:748)
[DEBUG] Clean up workflow ...
[DEBUG] Clean up workflow done
[DEBUG] Workflow execution done
[DEBUG] Create Analysis Report...
[DEBUG] Writing report to /home/stephanie/Projects/antenna_projects/antenna/example-projects/example-project/target/antenna/Antenna_3rdPartyAnalysisReport.txt
[DEBUG] Create Analysis Report...done
[DEBUG] Attaching artifacts to project
[DEBUG] Attaching artifact /home/stephanie/Projects/antenna_projects/antenna/example-projects/example-project/target/antenna/Antenna_3rdPartyAnalysisReport.txt
[DEBUG] Attaching artifacts to project..done
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15.072 s
[INFO] Finished at: 2020-04-28T15:21:22+02:00
[INFO] ------------------------------------------------------------------------
[...]

PR #495, opened 2020-04-28
Add attachment merging to SW360Release mergeWith()

neubs-bsi:
Issue: #494

Since attachments are part of a release they should be
merged along with everything else in the mergeWith method.

Now when checking the merged release in the
hasDuplicateAttachments method, attachments won't be empty
if there were any in the fetched release.

*Type of change*:  bug fix

Unit tests were added
Additionally this should be tested with the entry update_releases set to false in the SW360UpdaterImpl. The example project should be run twice, since the error is one that occurs when trying to upload a duplicate attachment. 

### Checklist
Must:

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
@@ -341,8 +341,8 @@ public class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects, SW36
             get_Links().setSelfComponent(componentIdWithPrecedence);
         }
         final List<SW360SparseAttachment> releaseWithPrecedenceAttachments = releaseWithPrecedence.get_Embedded().getAttachments();
-        if (releaseWithPrecedenceAttachments != null && !releaseWithPrecedenceAttachments.isEmpty()) {
-            if (get_Embedded().getAttachments() == null || get_Embedded().getAttachments().isEmpty()) {
+        if (!releaseWithPrecedenceAttachments.isEmpty()) {
+            if (get_Embedded().getAttachments().isEmpty()) {
                 get_Embedded().setAttachments(releaseWithPrecedenceAttachments);
             } else {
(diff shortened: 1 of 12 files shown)


modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -338,12 +340,26 @@ public SW360Release mergeWith(SW360Release releaseWithPrecedence) {
         if (componentIdWithPrecedence != null && !componentIdWithPrecedence.getHref().isEmpty()) {
             get_Links().setSelfComponent(componentIdWithPrecedence);
         }
+        final List<SW360SparseAttachment> releaseWithPrecedenceAttachments = releaseWithPrecedence.get_Embedded().getAttachments();
+        if (releaseWithPrecedenceAttachments != null && !releaseWithPrecedenceAttachments.isEmpty()) {
+            if (get_Embedded().getAttachments() == null || get_Embedded().getAttachments().isEmpty()) {
```
oheger-bosch, line 345: SW360ReleaseEmbedded.getAttachments() never returns null, so the null checks can be removed.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/rest/resource/releases/SW360Release.java
```diff
@@ -338,12 +340,26 @@ public SW360Release mergeWith(SW360Release releaseWithPrecedence) {
...
 
         return this;
     }
 
+    private List<SW360SparseAttachment> mergeAttachments(List<SW360SparseAttachment> attachments, List<SW360SparseAttachment> releaseWithPrecedenceAttachments) {
+        return Stream.concat(attachments.stream(), releaseWithPrecedenceAttachments.stream())
+                .distinct()
+                .collect(Collectors.toList());
+    }
+
```
oheger-bosch, line 362: One proposal: As duplicate attachments are obviously not allowed, it could be an option to store them in SW360ReleaseEmbedded in a Set rather than a List. Then a merge is straight-forward: just construct a union set from both releases involved.

Issue #490, opened 2020-04-23
PDF File not closed in Attribution Document Generator


blaumeiser-at-bosch:
### Summary of the Bug
If running an Antenna workflow, the following issue is happening:

[INFO] Run org.eclipse.sw360.antenna.attribution.document.workflow.generators.AttributionDocumentGenerator
[INFO] Generating pdf report.
[INFO] Run org.eclipse.sw360.antenna.workflow.generators.SourceZipWriter
[WARNING] Warning: You did not close a PDF Document

### Steps to reproduce
- Run Antenna Workflow with the AttributionDocumentGenerator
- Observe the issue

### Acceptance Criteria
- [x] The warning is not happening anymore

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

PR #503, opened 2020-05-05
Close all opened PDDocument objects

bs-ondem:
Fixes https://github.com/eclipse/antenna/issues/490

The class Templates wrapped all PDDocuments, which are needed for the
generation of the Attribution Document. After opening the PDDocuments
during the initialization of the Templates object, they were not
closed again which caused the Warning described in the issue. After
closing all PDDocument objects explicitly, the warning disappeared.

### Request Reviewer
@blaumeiser-at-bosch 

### Type of Change
*bug fix*

### Checklist
Must:

### DIFF
--- a/modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/AttributionDocumentGeneratorImpl.java
+++ b/modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/AttributionDocumentGeneratorImpl.java
@@ -1,5 +1,6 @@
 /**
  * Copyright (c) Robert Bosch Manufacturing Solutions GmbH 2019.
+ * Copyright (c) Bosch.IO GmbH 2020.
  * <p>
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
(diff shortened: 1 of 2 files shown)


modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/AttributionDocumentGeneratorImpl.java
```diff
@@ -72,15 +72,29 @@ public AttributionDocumentGeneratorImpl(String documentName, File workingDir, St
...
+            File copyright = writeCopyright(templates);
+            File backPage = writeBackPage(templates.getBackPage());
+            File artifactPages = writeArtifacts(templates, artifacts, templates.getContent());
+            File intermediateDoc = mergePages(title, copyright, artifactPages, backPage);
+
+            return postProcess(templates, intermediateDoc, documentName);
+        } finally {
+            if (templates != null) {
+                try {
+                    templates.getBackPage().close();
```
blaumeiser-at-bosch, line 89: I think, we should close each page in an own try catch, and only log, if there is an IOException here. It is not critical to the execution if the closing fails, but it should be remarked.

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/AttributionDocumentGeneratorImpl.java
```diff
@@ -72,15 +77,41 @@ public AttributionDocumentGeneratorImpl(String documentName, File workingDir, St
...
+            File artifactPages = writeArtifacts(templates, artifacts, templates.getContent());
+            File intermediateDoc = mergePages(title, copyright, artifactPages, backPage);
+
+            return postProcess(templates, intermediateDoc, documentName);
+        } finally {
+            if (templates != null) {
+                try {
+                    templates.getTitle().close();
+                } catch (IOException e) {
+                    LOG.warn("Failed to close the PDDocument object.", e);
```
blaumeiser-at-bosch, line 96: One small improvement here, the warning should not contain the exception, but should refer to the document that could not be closed. In addition, there should be a debug log with the exception. A normal user should not be confronted with exceptions, but in debug log, enough information should be given.

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/AttributionDocumentGeneratorImpl.java
```diff
@@ -72,15 +77,41 @@ public AttributionDocumentGeneratorImpl(String documentName, File workingDir, St
...
-        File title = writeTitle(templates);
-        File copyright = writeCopyright(templates);
-        File backPage = writeBackPage(templates.getBackPage());
-        File artifactPages = writeArtifacts(templates, artifacts, templates.getContent());
-        File intermediateDoc = mergePages(title, copyright, artifactPages, backPage);
-
-        return postProcess(templates, intermediateDoc, documentName);
+        Templates templates = null;
+        try {
+            templates = TemplateLoaderUtil.load(templateKey);
```
blaumeiser-at-bosch, line 82: If you would add a close method to Templates class and let it implement Closable, you could ease this here with try (Templates templates = ...) and move the close complexity to Templates.

PR #510, opened 2020-05-07
Get rid of critical sca warnings

blaumeiser-at-bosch:
Each commit is looking into warnings from one file raised during the SonarCloud analysis

### Request Reviewer
@bs-ondem @neubs-bsi @oheger-bosch 

### Type of Change

### How Has This Been Tested?
Standard regression tests

### Checklist
Must:


oheger-bosch:
The changes look good to me. There are, however, two test failures - probably related to the XML-related changes.

blaumeiser-at-bosch:
@oheger-bosch: I removed the failing commit, there seem to be an error in the SCA rule, because the advice does not work. It is refered to as a vulnerability, but I think the risk is acceptable for an command line tool to have this issue, so I will close the issue in SonarCloud.

PR #512, opened 2020-05-12
Simplify Jenkinsfile for Eclipse Infrastructure

blaumeiser-at-bosch:

There were one repository creation to much in the build script

### Request Reviewer
@bs-ondem @oheger-bosch 

### Type of Change

### How Has This Been Tested?
Regression Test

### Checklist
Must:

### DIFF
--- a/Jenkinsfile.eclipse
+++ b/Jenkinsfile.eclipse
@@ -48,9 +48,6 @@ pipeline {
                     IS_SNAPSHOT_VERSION = (params.VERSION ==~ /.*-SNAPSHOT/)
                 }
                 sh "echo \"VERSION is: ${params.VERSION} with IS_SNAPSHOT_VERSION=${IS_SNAPSHOT_VERSION}\""
-
-                sh 'rm -rf localRepository'
-                sh 'mkdir -p localRepository'
             }
         }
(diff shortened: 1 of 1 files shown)


Jenkinsfile.eclipse
```diff
@@ -60,8 +60,6 @@ pipeline {
             steps {
                 // build antenna and also deploy it to an output repository
                 sh """
-                    rm -rf repository
-                    mkdir -p repository
                     mvn -Dmaven.repo.local=\$(readlink -f localRepository) \
```
oheger-bosch, line 63: These statements have at least the effect that an already existing repository folder is cleaned, so that it does not contain any older files. However, I don't know whether this action is necessary in the context of the build.

blaumeiser-at-bosch: @oheger-bosch You are absolutely right, but if you open up the lines above, you see that this is already done in the previous step. But you are right, it is better to change it at the first call of these two tools.

blaumeiser-at-bosch: Ups, I think, now the structure is easier to understand. I overlooked that it were two different repositories used.

PR #515, opened 2020-05-14
Get rid of incompatible xml property

blaumeiser-at-bosch:
There is a mismatch on this properties. They are not known to the xml components we are using, although, e.g., SonarCloud marks the missing code as vulnerability. The vulnerability is not an issue in our tooling, so I think removing is ok.

### Request Reviewer
@bs-ondem 

### Type of Change
*Type of change*:  bug fix

### How Has This Been Tested?
Regression Test and example project test in Tina later

### Checklist
Must:


blaumeiser-at-bosch:
The introduced vulnerability was expected, the removed code was there to get rid of it. Unfortunately, the proposed solution does not work, it creates a runtime exception. The vulnerability is already marked in SonarCloud, so the missing check here is acceptable.

PR #518, opened 2020-05-25
Oheger bosch/sw360/#426 tweak SW360 release client API

oheger-bosch:
Issue 426

This PR is part of the redesign of the SW360 client library. It focuses on the API of the client for releases with the goal to remove methods specific to Antenna use cases and replace them with more generic methods. Antenna-specific logic therefore has been moved to the classes in the workflow and compliance-tool packages.

Methods considered Antenna-specific are for instance getOrCreateRelease() with an integrated update of an existing release entity, or the methods that lookup releases based on a passed in release object.

A missing update() method has been added, and the API to upload attachments has been improved.

Javadoc has been added to all methods in the interfaces.

@neubs-bsi @blaumeiser-at-bosch 

*Type of change*:  
improvements

### How Has This Been Tested?
Where necessary, existing unit tests have been adapted where logic has been moved. For new classes, new unit tests have been created.

### Checklist
Must:

- [X] I have provided tests for the changes (if there are changes that need additional tests)

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/AttachmentUploadRequest.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/AttachmentUploadRequest.java
@@ -10,8 +10,8 @@
  */
 package org.eclipse.sw360.antenna.sw360.client.adapter;
 
+import org.eclipse.sw360.antenna.sw360.client.rest.resource.SW360HalResource;
 import org.eclipse.sw360.antenna.sw360.client.rest.resource.attachments.SW360AttachmentType;
-import org.eclipse.sw360.antenna.sw360.client.rest.resource.releases.SW360Release;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -30,12 +30,13 @@ import java.util.Objects;
  * is created using a builder that offers methods to add the items to be
  * uploaded.
  * </p>
+ * @param <T> the type of the entity to upload attachments to
  */
-public final class AttachmentUploadRequest {
+public final class AttachmentUploadRequest<T extends SW360HalResource<?, ?>> {
     /**
-     * The release that is the target for uploads.
+     * The entity that is the target for uploads.
      */
-    private final SW360Release target;
+    private final T target;
 
     /**
      * Stores the items to be uploaded.
@@ -49,20 +50,20 @@ public final class AttachmentUploadRequest {
      * @param target the target entity of the uploads
      * @param items  a list with the items to be uploaded
      */
-    private AttachmentUploadRequest(SW360Release target, List<Item> items) {
+    private AttachmentUploadRequest(T target, List<Item> items) {
         this.target = target;
         this.items = Collections.unmodifiableList(new ArrayList<>(items));
     }
 
     /**
      * Returns a new {@code Builder} to define a request to upload attachments
-     * to the given release.
+     * to the given entity.
      *
      * @param target the target of the upload operation
      * @return the builder to define the upload request
      */
-    public static Builder builder(SW360Release target) {
-        return new Builder(target);
+    public static <T extends SW360HalResource<?, ?>> Builder<T> builder(T target) {
+        return new Builder<>(target);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -104,6 +106,16 @@ public boolean isUploadSources() {
...
-        return releaseClientAdapter.uploadAttachments(sw360Release, attachments);
+        AttachmentUploadRequest.Builder builder = AttachmentUploadRequest.builder(sw360Release);
+        for (Map.Entry<Path, SW360AttachmentType> e : attachments.entrySet()) {
+            builder = builder.addAttachment(e.getKey(), e.getValue());
+        }
+
+        AttachmentUploadResult result = releaseClientAdapter.uploadAttachments(builder.build());
+        LOGGER.debug("Result of attachment upload operation: {}", result);
+        if (!result.isSuccess()) {
+            LOGGER.error("Failed to upload attachments: {}", result.failedUploads());
```
neubs-bsi, line 117: When the error is due to a duplicate attachment, this error message will be logged many times. And in case of the duplicate attachments it should not really be an error, since the attachment in the end does exist in SW360 with the release. 
While a bit laborious, could for example the exception be checked for the word duplicate or sth similar? because from a users perspective error messages should only be thrown when there is an acutal error. Dupliactes should be a warning at most

oheger-bosch: Not sure whether this is true in all use cases. The local attachment file may be different from the one on the server. If the library does not return error information in this case, the user would expect that it had been replaced on the server, while it had not.

neubs-bsi: Is there a way to determine if it is the same? because i think printing an error message each time might also lead the user to not take the message seriously in case it actually is a different file.

oheger-bosch: SW360 maintains a hash value for attachments that have been uploaded. So by calculating the hash on the local file (we could add an utility method to support this), one can determine whether files have changed.

It would also be possible to treat duplicates not as errors in general. The upload result object could contain a list with duplicates that have been detected. Then the caller can decide how to treat this situation.

I am afraid that the API related to attachments seems to be very restricted in SW360. A once uploaded attachment file cannot be deleted or overwritten. So if you once uploaded a wrong attachment file, you are stuck with it.

neubs-bsi: The above suggestion sounds promising (and probably worth another isuue / pr) 

maybe we can also check out what changes could be done in sw360.

The for now this should be fine and we can find a solution in a separate PR

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/AttachmentUploadRequest.java
```diff
@@ -0,0 +1,216 @@
...
+ * It is possible to upload multiple attachments in a single request. A request
+ * is created using a builder that offers methods to add the items to be
+ * uploaded.
+ * </p>
+ */
+public final class AttachmentUploadRequest {
+    /**
+     * The release that is the target for uploads.
+     */
+    private final SW360Release target;
```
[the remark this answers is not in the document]
oheger-bosch: _AttachmentUploadRequest_ and _AttachmentUploadResult_ are now generic and support different entity types. The logic to upload and download attachments has been extracted into an utility class.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/AttachmentUploadRequest.java
```diff
@@ -0,0 +1,216 @@
...
+    public SW360Release getTarget() {
+        return target;
+    }
+
+    /**
+     * Returns a list with the items that are to be uploaded.
+     *
+     * @return a list with the items to be uploaded
+     */
+    public List<Item> items() {
```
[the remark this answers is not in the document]
oheger-bosch: Method was renamed.

Issue #517, opened 2020-05-19
Content Reporter for compliance tool


neubs-bsi:
### Summary of the Feature
When using the compliance tool it would be good to get more information from the running sw360 instance than the exporter provides.
Therefore a status reporter with different switches for different desired information would be good

### Acceptance Criteria
- [ ] New mode "reporter"
- [ ] Reporter gives back in csv form
    - [ ] releases_cleared
    - [ ] release_not_cleared
    - [ ] projects
    - [ ] projects_that_use_release_<id>
    - [ ] releases_of_project_<name>
    - [ ] components
    - [ ] releases_of_component_<id>
    - [ ] ...
- [ ] should be given via command line

### Definition of Done
- Acceptance criteria fulfilled
- A PR is created, the CI infrastructure reports green
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Test cases are created to prove the functionality of the feature
- Documentation is updated

PR #522, opened 2020-05-28
Neubs bsi/ct/reporter start

neubs-bsi:
This introduces a status reporter to the compliance tool to get reports on different data in sw360.

It works with the same properties file as the other modes and is choosable by a switch, just like the others. 
With this PR parameter get introduced to the CLI Options since they are used to determine which information is requested from sw360 by the user 

Issue: #517 

@oheger-bosch 

*Type of change*:  new feature

Some tests have been written. 

### Checklist
Must:


assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/main/AntennaComplianceToolOptions.java
```diff
@@ -55,20 +75,26 @@
...
      * @param debugLog           flag whether debug log should be active
      * @param showHelp           flag whether the help message should be printed
      * @param valid              flag whether the command line is valid
      */
-    AntennaComplianceToolOptions(String propertiesFilePath, String complianceMode, boolean debugLog, boolean showHelp, boolean valid) {
+    AntennaComplianceToolOptions(String propertiesFilePath, String complianceMode, Set<String> parameters, boolean debugLog, boolean showHelp, boolean valid) {
         super(debugLog, showHelp, valid);
         this.propertiesFilePath = propertiesFilePath;
         this.complianceMode = complianceMode;
+        this.parameters = parameters;
```
oheger-bosch, line 97: To make sure that the AntennaComplianceToolOptions class remains immutable, the set with parameters should not be directly exposed to clients, so that it cannot be manipulated after an instance was created. So create a defensive copy here and in _getParameters()_, e.g. 
`this.parameters = new HashSet(parameters);`
Alternatively, the set can be stored as an unmodifiable set in the field, then it can be returned by _getParameters()_ directly, e.g.
`this.parameters = Collections.unmodifiableSet(new HashSet(parameters));`

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/main/AntennaComplianceToolOptions.java
```diff
@@ -151,13 +189,16 @@ static String helpMessage() {
     private static Optional<String> getComplianceModeFromSwitches(Set<String> switches) {
         boolean updater = hasSwitch(switches, SWITCH_UPDATER_SHORT) || hasSwitch(switches, SWITCH_UPDATER_LONG);
         boolean exporter = hasSwitch(switches, SWITCH_EXPORTER_SHORT) || hasSwitch(switches, SWITCH_EXPORTER_LONG);
+        boolean reporter = hasSwitch(switches, SWITCH_REPORTER);
 
         if (updater && exporter) {
```
oheger-bosch, line 194: The checking logic is now more complex; the new switch has to be taken into account as well.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/SW360StatusReporterParameters.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Class responsible for parsing status reporter parameters
+ * and creating the {@code InfoRequest} the status reporter
+ * uses.
+ */
+class SW360StatusReporterParameters {
```
oheger-bosch, line 17: The class name is a bit misleading. Maybe something like ParametersHelper or ParameterParser?

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/SW360StatusReporterParameters.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        if (parameters.stream().filter(p -> p.contains(REPORTER_PARAMETER_PREFIX)).count() != 1) {
+            LOGGER.error("Too many information requests were made in this status report. ");
+            LOGGER.info(helpMessage());
+            throw new IllegalArgumentException("Too many information requests were made in this status report. ");
+        }
+        return parameters.stream().filter(p -> p.contains(REPORTER_PARAMETER_PREFIX)).findFirst().orElse("");
+    }
+
+    /**
```
oheger-bosch, line 60: Parsing of parameters is pretty inefficient as for each parameter the whole set is traversed. Alternatively, the set could be processed once and transformed into a map with key-value pairs. From the map it is then trivial to find the value of a specific parameter. During this transformation, it is also possible to detect duplicate parameter keys.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/SW360StatusReporterParameters.java
```diff
@@ -0,0 +1,139 @@
...
+     */
+    static String helpMessage() {
+        String cr = System.lineSeparator();
+        return "Usage: java -jar compliancetool.jar " + AntennaComplianceToolOptions.SWITCH_REPORTER + "[options] <complianceMode> <propertiesFilePath>" + cr + cr +
+                "The reporter create a csv file for every run with the requested information statement. " + cr + cr +
+                "The reporter info statements: (only one can be set)" + cr +
+                new IRGetClearedReleases().getInfoParameter() + ":   Gives a list of all releases in a given sw360 instances that are cleared." + cr +
+                new IRGetNotClearedReleases().getInfoParameter() + ":   Gives a list of all releases in a given sw360 instances that are not cleared." + cr +
+                new IRGetReleasesOfProjects().getInfoParameter() + ":   Gives a list of all releases of a given project in a given sw360 instances." + cr;
+    }
```
oheger-bosch, line 120: The information which concrete reports are supported is distributed over multiple classes; here the parameters class has to know all supported report types in order to generate the help message. Only the InfoRequestFactory class should know this. So either InfoRequestFactory could generate the help message or it could provide a method which returns a collection of all supported InfoRequest objects; from this collection the help text can then be generated.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/InfoRequest.java
```diff
@@ -0,0 +1,55 @@
+package org.eclipse.sw360.antenna.frontend.compliancetool.sw360.status;
```
oheger-bosch, line 1: Should the package name be something like "report" or "reporter"? "status" is pretty generic.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/InfoRequest.java
```diff
@@ -0,0 +1,55 @@
...
+abstract public class InfoRequest<T> {
+    abstract public String getInfoParameter();
+
+    boolean hasAdditionalParameters() {
+        return getAdditionalParameters().size() > 0;
+    }
+
+    abstract String helpMessage();
+    abstract boolean isValid();
+    abstract Set<String> getAdditionalParameters();
```
oheger-bosch, line 18: Idea: _getAdditionalParameters()_ could return more information than just parameter names, e.g. data objects with the parameter name, the help text, and a flag whether the parameter is optional/mandatory. Then a major part of the parsing could be already implemented by the base class, also the generation of the help message.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/ReporterOutput.java
```diff
@@ -0,0 +1,13 @@
...
+import org.eclipse.sw360.antenna.frontend.compliancetool.sw360.SW360Configuration;
+
+import java.util.Collection;
+
+/**
+ * Central interface for the report output of the status reporter
+ * @param <T>
+ */
+public interface ReporterOutput<T> {
+    void printFile(Collection<T> result, SW360Configuration configuration);
```
oheger-bosch, line 12: The use of generics is strange here. A concrete implementation will have to deal with multiple entity types. So the type parameter should be on the method rather than the interface. Alternatively, there could be multiple _printFile()_ methods for different entities.

Or is it expected to have implementations per format and entity class, e.g. ReporterOutputCSVProject, ReporterOutputHtmlRelease, ...?

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/ReporterOutputFactory.java
```diff
@@ -0,0 +1,18 @@
...
+ * Factory class for creating a {@link ReporterOutput} implementation
+ */
+class ReporterOutputFactory {
+    /**
+     * create function to create an object implementing the {@link ReporterOutput}
+     * @param outputFormat String representation of the type of the object
+     * @return {@link ReporterOutput} of the output format
+     */
+    static ReporterOutput getReporterOutput(String outputFormat) {
+        if (outputFormat.equals("CSV") || outputFormat.equals("csv") || outputFormat.equals("Csv")) {
```
oheger-bosch, line 13: if (outputFormat.equalsIgnoreCase("csv")) { ... }

assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/SW360StatusReporterTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SW360StatusReporterTest {
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+
+    @Mock
```
oheger-bosch, line 23: Annotation is not needed.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
```diff
@@ -386,4 +386,34 @@ public int hashCode() {
...
+    private String collectionToString(Collection<String> collection) {
+        String separator = "";
+        StringBuilder mainLicenses = new StringBuilder();
+        for (String item : collection) {
+            mainLicenses.append(separator);
+            separator = ",";
+            mainLicenses.append(item);
+        }
+        return mainLicenses.toString();
+    }
```
oheger-bosch, line 418: I am not sure whether these format-specific output methods should be part of the data model. I would rather keep this separate, e.g. as part of the ReportOutputCSV implementation. When we support other output formats, would we add corresponding methods as well?

neubs-bsi: oh I agree. I should have mentioned that I regarded this as an intermediate solution until i come up with a better idea for the demo.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/status/ReporterOutput.java
neubs-bsi: I have not yet implemented the visitor pattern we talked about. 
I felt it better to do these changes in a separate PR when we know exactly what we want.

assembly/compliance-tool/src/site/markdown/updater.md
```diff
@@ -0,0 +1,28 @@
...
+### Properties
+- `targetDir`: Target directory in which artificial clearing reports are created in. 
+- `sourcesDirectory`: Directory where the sources for uploading are stored
+- `basedir`: Base directory of the execution
+- `csvFilePath`: Path and name to the csv file with the release information
+- `delimiter`: Delimiter used in the csv file to separate columns (by default it is `,`)
+- `encoding`: Encoding of the csv file, normally `UTF-8` 
+- `removeClearedSources`: A boolean property that controls whether the updater should remove the source attachments of a release from the local sources directory once the release has been cleared. Cleared releases are no longer relevant for the workflow of the compliance tool; so by setting this property to *true*, an automatic cleanup of the sources directory can be enabled. The default value is *false*.
+- `removeClearingDocuments`: A boolean property that controls whether clearing documents for releases should be removed after the release has been cleared. The default value is *false*.
+- `proxyHost`: If a proxy is in use, supply the host name
```
oheger-bosch, line 18: Idea: To get rid off duplication, should the properties common to all execution modes (e.g. SW360 connection details) be described on the top-level site for the compliance tool?

assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/reporter/IRGetClearedReleasesTest.java
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(clearedReleases).containsExactly(release);
+    }
+
+    /**
+     * create a connection with all mock objects that get
+     * called when executing getting the releases.
+     *
+     * @return mocked sw360 connection
+     */
+    static SW360Connection getSW360Connection(SW360Release release) {
```
oheger-bosch, line 57: Maybe a candidate to be refactored into a test helper class, as the method is also used by the test class for not cleared releases.

assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/reporter/ReporterUtils.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static String releaseCsvPrintRow(SW360Release release) {
+        return String.format("%s;%s;%s;%s;%s;%s;%s;%s;%s;%s;%s;%s;%s;%s",
+                release.getId(), release.getName(), release.getVersion(),
+                collectionToString(release.getCoordinates().values()),
+                collectionToString(release.getMainLicenseIds()),
+                release.getOverriddenLicense(), release.getDeclaredLicense(), release.getObservedLicense(),
+                release.getCopyrights(), release.getClearingState(), release.getChangeStatus(),
+                release.getDownloadurl(), release.getReleaseTagUrl(), release.getSoftwareHeritageId());
+    }
```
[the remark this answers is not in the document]
neubs-bsi: The csv printer only takes complete records ( meaning rows). If I would get the fields as Object[] I would have to iterate over the arrays of fields and give them individually as records, making the chance high to have mismatched placements. 
Did I understand your suggestion correctly?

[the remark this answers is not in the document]
oheger-bosch: I was referring to the printRecord() method of CSVPrinter which expects an Object[] (as var args) as argument. It can be passed an object of fields and prints them as a single row. This method is already used currently, but only a single string (with the already generated row) is passed in. The generation of this string is done manually, and I was hoping that it could be delegated to CSVPrinter.

PR #524, opened 2020-06-03
Removed Antenna runtime from attribution-document-core

bs-ondem:
The Antenna runtime was removed from the attribution-document-core due
to it was a unused dependency. As a result, there is no circular
dependency in the new ORT PDF reporter.

Signed-off-by: Onur Demirci <onur.demirci@bosch.io>

### Request Reviewer
@sschuberth

### DIFF
--- a/modules/attribution-document/attribution-document-core/pom.xml
+++ b/modules/attribution-document/attribution-document-core/pom.xml
@@ -47,10 +47,6 @@
             <artifactId>commons-lang3</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.github.package-url</groupId>
-            <artifactId>packageurl-java</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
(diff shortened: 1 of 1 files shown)


sschuberth:
To verify that the `org.eclipse.sw360.antenna:runtime` dependency is not needed I ran

    mvn -B -f modules/attribution-document/attribution-document-core/pom.xml dependency:analyze

which gave

    Used undeclared dependencies found:
       org.junit.jupiter:junit-jupiter-api:jar:5.5.2:test
    Unused declared dependencies found:
       org.eclipse.sw360.antenna:runtime:jar:1.0.0-SNAPSHOT:compile
       com.github.package-url:packageurl-java:jar:1.1.1:compile
       org.junit.jupiter:junit-jupiter:jar:5.5.2:test
       org.eclipse.sw360.antenna:antenna-license-provider:zip:licenseinfo:1:provided

So it seems we could at least remove the dependency on `com.github.package-url:packageurl-java` safely, too. Would you mind adding another commit to this PR doing so?

bs-ondem:
Thank you for the useful command :) I also removed packageurl-java.

PR #528, opened 2020-06-04
Separate OrtDownloadProcessor into Impl

neubs-bsi:
This to decouple the functionality from the
AntennaContext.

### DIFF
--- a/modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessorImpl.kt
+++ b/modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessorImpl.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) Bosch.IO GmbH 2020.
+ *
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessorImpl.kt
```diff
@@ -0,0 +1,67 @@
+package workflow.processors.enricher
```
sschuberth, line 1: Missing copyright header.

modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessorImpl.kt
```diff
@@ -0,0 +1,67 @@
...
+                        artifact.addFact(ArtifactSourceFile(zipFile.toPath()))
+                    }
+                }
+            } catch (e: DownloadException) {
+                LOGGER.warn("Failed to download sources for '${name}'")
+            }
+        }
+        return intermediates
+    }
+}
```
sschuberth, line 67: Missing final new line.

PR #530, opened 2020-06-08
Oheger bosch/compliance/#529 workflow support exporter

oheger-bosch:
Issue 529.

This PR extends the exporter of the compliance tool by the functionality to remove unreferenced sources from the sources directory. The functionality can be enabled by a new setting in the configuration properties file. If enabled, as a final step, the exporter iterates over the files in the sources directory and checks whether they belong to one of the current components. If not, they are deleted.

The export logic related to source attachments has been moved into a new class _SourcesExporter_. It is invoked from the main export process. The download of source attachments now also makes use of the asynchronous capabilities of the SW360 client library; this enables a concurrent download of attachments.

The documentation of the compliance tool has been extended to cover the new configuration property.

@neubs-bsi 

*Type of change*:  
new feature

### How Has This Been Tested?
A new test class for the new _SourcesExporter_ class has been added. The test cases for the exporter have been extended to check the collaboration with the sources exporter and the processing of the new configuration property.

### Checklist
Must:

### DIFF
--- a/assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SourcesExporter.java
+++ b/assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SourcesExporter.java
@@ -1,0 +1,218 @@
+/*
+ * Copyright (c) Bosch.IO GmbH 2020.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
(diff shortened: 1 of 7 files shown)


assembly/compliance-tool/src/main/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/exporter/SourcesExporter.java
```diff
@@ -0,0 +1,218 @@
...
+ * An internally used helper class that handles the download of source
+ * attachments for releases.
+ * </p>
+ * <p>
+ * The compliance tool exporter delegates to an instance of this class for all
+ * download-related functionality.
+ * </p>
+ */
+class SourcesExporter {
+    private static final Logger LOG = LoggerFactory.getLogger(SourcesExporter.class);
```
neubs-bsi, line 56: Convention in antenna is to call a logger a `LOGGER`

PR #531, opened 2020-06-08
Oheger bosch/compliance/#529 updater workflow

oheger-bosch:
Issue 529:

This PR adds another property to the configuration of the Compliance Tool Updater. If set to *true*, the exporter removes the source attachments from the local sources directory when a release has been cleared. So the directory gets cleaned up automatically when releases are no longer relevant for the compliance tool workflow.

@neubs-bsi 

*Type of change*:  
new feature

### How Has This Been Tested?
The existing test cases for the updater have been extended to cover the new functionality.

### Checklist
Must:

### DIFF
--- a/assembly/compliance-tool/src/site/markdown/index.md.vm
+++ b/assembly/compliance-tool/src/site/markdown/index.md.vm
@@ -72,6 +72,7 @@ It has the ability to update release information of already existing releases.
 - `csvFilePath`: Path and name to the csv file with the release information
 - `delimiter`: Delimiter used in the csv file to separate columns (by default it is `,`)
 - `encoding`: Encoding of the csv file, normally `UTF-8` 
+- `removeClearedSources`: A boolean property that controls whether the exporter should remove the source attachments of a release from the local sources directory once the release has been cleared. Cleared releases are no longer relevant for the workflow of the compliance tool; so by setting this property to *true*, an automatic cleanup of the sources directory can be enabled. The default value is *false*.
 - `proxyHost`: If a proxy is in use, supply the host name
 - `proxyPort`: If a proxy is in use, supply the port
 - `proxyUse`: If a proxy is in use, this should be set to true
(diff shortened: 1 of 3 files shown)


assembly/compliance-tool/src/site/markdown/index.md.vm
```diff
@@ -72,6 +72,7 @@ It has the ability to update release information of already existing releases.
 - `csvFilePath`: Path and name to the csv file with the release information
 - `delimiter`: Delimiter used in the csv file to separate columns (by default it is `,`)
 - `encoding`: Encoding of the csv file, normally `UTF-8` 
+- `removeClearedSources`: A boolean property that controls whether the exporter should remove the source attachments of a release from the local sources directory once the release has been cleared. Cleared releases are no longer relevant for the workflow of the compliance tool; so by setting this property to *true*, an automatic cleanup of the sources directory can be enabled. The default value is *false*.
```
neubs-bsi, line 75: Do we want to also add this in our test resource?

oheger-bosch: The unit tests run the updater with the flag enabled and the flag not present (which defaults to *false*). This can be achieved more easily if the flag is not contained in the test resource.

neubs-bsi: I'm just thinking that currently the test resource properties files serve as kind of example files... would the property, but commented out be an option for you? or do you dislike to add unused comments?

oheger-bosch: As long as the files are used in the tests and the tests make certain assumptions about their content, I do not think that they are actually example files; they can only define a specific scenario that the test is testing.

We could have multiple test resource files for the different test scenarios, but then we need to adapt the test setup.

I would rather prefer to describe the properties supported in the documentation and maybe give some examples there. This is probably also more visible.

Issue #529, opened 2020-06-04
Add better workflow support to Compliance tool


oheger-bosch:
### Summary of the Feature
The compliance tool currently supports exports from (data about components and source attachments) and updates to SW360. When used in an OSS curation workflow typically some more functionality is needed. For instance, it should be possible to somehow sync a target directory with the current state in SW360, so that source attachments that are no longer in the focus are automatically deleted. As such operations are not needed for all use cases, they should be enabled/disabled using new command line switches.

### Acceptance Criteria
- [ ] The exporter should provide an option to synchronize the sources download folder. If enabled, all source attachments that do not belong to the current list of components should be removed.
- [ ] The updater should provide a similar option. If enabled, the attachments related to components that are marked as cleared should be removed.

### Definition of Done
- Acceptance criteria fulfilled
- A PR is created, the CI infrastructure reports green
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Test cases are created to prove the functionality of the feature
- Documentation is updated

oheger-bosch:
The functionality requested has been implemented and merged to master.

PR #535, opened 2020-06-10
Neubs bsi/ct/clearing doc gen exception

neubs-bsi:
Issue: #529 
This is changing the exception thrown when a clearing document could not be generated and catches it were all exceptions are caught when something goes wrong when an artifact is updated. 

@oheger-bosch 

*Type of change*:  bug fix

Tests were adapted

### Checklist
Must:


neubs-bsi:
sonarCloud is not passing, due to missing coverage. Missing coverage is due to unthrown exception, but there are tests testing if the exception is handled correctly once thrown by the calling method.

PR #536, opened 2020-06-15
Oheger bosch/compliance/#529 attachment improvements

oheger-bosch:
Issue: #529 

This PR adds a new function to SW360AttachmentUtils to calculate hash values for local files.  The exporter and the updater from the compliance tool use this function to determine whether a local attachment file is up-to-date with the version stored in SW360. If this is the case, a download or update step can be skipped.

@neubs-bsi 

*Type of change*:  
improvements

### How Has This Been Tested?
New tests for the added/modified functionality have been added.

### Checklist
Must:


modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -108,15 +112,74 @@ public boolean isUploadSources() {
...
+     *
+     * @param path           the path to the local attachment file
+     * @param attachmentType the attachment type
+     * @param fileName       the attachment file name
+     * @param attachment     the attachment from the current release
+     * @return a flag whether the attachment is matched by these properties
+     */
+    private boolean attachmentMatches(Path path, SW360AttachmentType attachmentType, String fileName, SW360SparseAttachment attachment) {
+        try {
+            return Objects.equals(fileName, attachment.getFilename()) &&
```
neubs-bsi, line 176: Since we have checked last week, the duplicate attachment function in SW360 currently still marks anattachment as duplicate when either the filename OR the sha1 is the same, I think this should be reflected as well. 

Or is it your aim to have a logical duplicate check here and rather catch the exception thrown in SW360 until they remove the filename check in their duplicate attachment statement. In that case checking for the filename would be superfluous, because an attachment can not be updated if for example only the sha1 matches. 

Either way I think this deserves a comment for future changes that it either doesn't reflect the duplicate checking in SW360 or it should reflect the duplicate checking of SW360

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -108,15 +112,74 @@ public boolean isUploadSources() {
...
+     * @param release        the release affected
+     * @param path           the path to the local attachment file
+     * @param attachmentType the attachment type
+     * @return a flag whether this attachment file must be uploaded
+     */
+    private boolean needUpload(SW360Release release, Path path, SW360AttachmentType attachmentType) {
+        String fileName = String.valueOf(path.getFileName());
+        Optional<SW360SparseAttachment> optAttachment = release.getEmbedded().getAttachments().stream()
+                .filter(attachment -> attachmentMatches(path, attachmentType, fileName, attachment))
+                .findFirst();
```
neubs-bsi, line 159: A duplicate attachment problem can also come from an attachment not related to this release. 

An additional possibility to check for the need to upload after this check could be to make a call <sw360-url>/attachments/?sha1=<sha1> and check for the response object. but that might be overdoing it.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -108,15 +112,74 @@ public boolean isUploadSources() {
...
+     * Calculates the SHA-1 hash for the given local attachment file. This is
+     * used to determine whether a modified attachment file needs to be
+     * uploaded.
+     *
+     * @param path the path to the local file
+     * @return the SHA-1 hash of this file
+     * @throws SW360ClientException if an error occurs
+     */
+    String calculateAttachmentHash(Path path) {
+        return SW360AttachmentUtils.calculateSha1Hash(path);
```
neubs-bsi, line 142: isn't this method a bit superfluous?

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -108,15 +112,74 @@ public boolean isUploadSources() {
     public SW360Release uploadAttachments(SW360Release sw360Release, Map<Path, SW360AttachmentType> attachments) {
         AttachmentUploadRequest.Builder<SW360Release> builder = AttachmentUploadRequest.builder(sw360Release);
         for (Map.Entry<Path, SW360AttachmentType> e : attachments.entrySet()) {
-            builder = builder.addAttachment(e.getKey(), e.getValue());
+            if (needUpload(sw360Release, e.getKey(), e.getValue())) {
```
neubs-bsi, line 115: I am just wondering if this check wouldn't be better placed in the sw360 client, maybe the attachmentuploadrequest. Since this is something that should be applied to every kind of attachment uploaded, no matter which halresource it goes with and it is not an antenna specific check, but a sw360 specific one.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -108,15 +112,74 @@ public boolean isUploadSources() {
...
+     * @param path           the path to the local attachment file
+     * @param attachmentType the attachment type
+     * @param fileName       the attachment file name
+     * @param attachment     the attachment from the current release
+     * @return a flag whether the attachment is matched by these properties
+     */
+    private boolean attachmentMatches(Path path, SW360AttachmentType attachmentType, String fileName, SW360SparseAttachment attachment) {
+        try {
+            return Objects.equals(fileName, attachment.getFilename()) &&
+                    attachmentType == attachment.getAttachmentType() &&
```
neubs-bsi, line 177: if they have a different type, but the same filename and sha1 they shouldn't be uploaded, since they can't be of different types if it is the same file

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
oheger-bosch: In my tests it was possible to upload the same attachment file with the same name to multiple releases. So I do not think that other releases affect a specific upload. Or did I get you wrong?

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
oheger-bosch: It is used by tests to check whether exceptions are properly handled.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360AttachmentUtils.java
```diff
@@ -100,6 +117,46 @@ private SW360AttachmentUtils() {
                 .orElseGet(() -> CompletableFuture.completedFuture(Optional.empty()));
     }
 
+    /**
+     * Calculates a hash value on the content of the file specified. This
+     * functionality is useful in relation with attachments, as by comparing
+     * the hashes it can be determined whether a local file is identical to an
+     * attachment that was uploaded to SW360.
+     *
+     * @param file   the path to the file for which the hash should be computed
```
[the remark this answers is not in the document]
oheger-bosch: My code formatter aligns parameter descriptions based on the longest parameter name. I am not aware that I have changed this setting explicitly. Is there a way to import the Antenna-specific coding style into the IDE? (And is this style somewhere defined at all?)

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360AttachmentUtils.java
```diff
@@ -100,6 +117,46 @@ private SW360AttachmentUtils() {
...
+     * @param file the path to the file for which the hash should be computed
+     * @return the (Hex-encoded) hash value for this file
+     * @throws SW360ClientException if an error occurs
+     */
+    public static String calculateSha1Hash(Path file) {
+        try {
+            return calculateHash(file, MessageDigest.getInstance(ALG_SHA1));
+        } catch (NoSuchAlgorithmException e) {
+            // This cannot happen as every implementation of the Java platform must support this algorithm
+            throw new AssertionError("SHA-1 algorithm not supported");
```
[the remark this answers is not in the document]
oheger-bosch: In "Effective Java" Bloch recommends using this exception for things that should never happen.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360AttachmentUtils.java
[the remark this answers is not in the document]
neubs-bsi: Ah, I didn't see it was aligning due to the code changes shown below. 
I think that's fine, just though it wasn't on purpose

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360AttachmentUtilsTest.java
```diff
@@ -0,0 +1,62 @@
...
+        MessageDigest digest = MessageDigest.getInstance("md5");
+
+        SW360AttachmentUtils.calculateHash(file, digest);
+    }
+
+    @Test
+    public void testCalculateSha1Hash() throws URISyntaxException {
+        String sha1Hash = SW360AttachmentUtils.calculateSha1Hash(testFile());
+
+        assertThat(sha1Hash).isEqualTo("7a5daedffafd0be187c351968592fefee4f648f4");
```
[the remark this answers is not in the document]
oheger-bosch: Added comments.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
oheger-bosch: The matches check now only checks for the SHA-1 value. A comment has been added that mentions the current limitations of the solution.

The final solution will probably depend on the functionality SW360 will offer in this area in future. I hope that it will be possible to delete all wrong or out-dated source attachments before uploading the correct one.

neubs-bsi: Hm, I might not have said that right earlier. 
The duplicate attachment checks `(isFileNameDuplicate || isSha1Duplicate)`, so there still is a need to test the filename, since the OR condition will give back true if the filenames match

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
neubs-bsi: I just checked the SW360 code again. You are correct. I did get it wrong.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
oheger-bosch: Not sure about this. Actually, the additional check is just an optimization, which we can apply because we know our workflows. It is not clear whether this is possible in general.
For instance, the SW360Release entity that is target of the upload may not be up-to-date because of other updates on the server; then the check would be done on stale data. To be on the safe side, the most recent entity would have to be queried from the server first; but then it is easier to do the upload directly and let SW360 fail based on its logic (which after all is the ultimo ratio). - It would be nice, however, if SW360 returned a meaningful error message rather than throwing a generic 500 error when it does not like a specific attachment upload.

neubs-bsi: Fair enough.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
oheger-bosch: I hope, this is covered by the comment which says that the result of the check method does not guarantee a successful upload. The compliance tool will probably have to do additional steps (like deleting old attachments) before the curated source attachment can be uploaded.

neubs-bsi: okay, could you create an Issue for the TODO and link it in this PR? Then everything is properly documented

oheger-bosch: Issue was created:
https://github.com/eclipse/antenna/issues/538

PR #543, opened 2020-06-19
AttributionDocumentGeneratorImpl can generate the docuement with files

bs-ondem:
A new functionality is added to the AttributionDocumentGeneratorImpl,
to generate the attribution document only with given template files
without the need of the ServiceLoader.

The antenna-demo_template.* PDF test files are also added to the
attribution-document-core JAR to reuse the files in attribution-
document-generator tests.

### Request Reviewer
@sschuberth @blaumeiser-at-bosch 

### Type of Change
*improvement*


sschuberth:
Commit message typo: s/docuement/document/

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/utils/TemplateLoaderUtil.java
```diff
@@ -32,9 +33,57 @@ private TemplateLoaderUtil() {
...
     public static Templates load(String key) {
-        TemplateBundle bundle = getLoader(key);
+        return load(getLoader(key));
+    }
+
+    public static Templates load(File cover, File copyright, File content, File back) {
+        return load(new TemplateBundle() {
+            @Override
+            public String key() {
+                return "";
```
sschuberth, line 43: Would it make sense to derive an "artificial" key from the files names? I'm not sure how the return value of this function is used, if at all. Maybe it's just fine to always return a empty string here.

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/utils/TemplateLoaderUtil.java
```diff
@@ -32,9 +33,57 @@ private TemplateLoaderUtil() {
...
+                } catch (IOException e) {
+                    throw new ExecutionException("Could not load back template PDF", e);
+                }
+            }
+        });
+    }
+
+    public static Templates load(TemplateBundle bundle) {
         Templates templates = new Templates();
+
```
sschuberth, line 86: Nit: Unrelated formatting change.

modules/attribution-document/attribution-document-generator/src/test/java/org/eclipse/sw360/antenna/attribution/document/workflow/generators/YagenIntegrationTest.java
```diff
@@ -31,25 +34,52 @@
...
+              new AttributionDocumentGeneratorImpl("attribution_document.pdf", tmpDir, "antenna-demo", values);
 
       List<ArtifactAndLicense> foo = createData();
       File file = attrDocGen.generate(foo);
       assertThat(file).exists().hasExtension("pdf");
    }
 
+   @Test
+   void goGenerateWithFiles() throws IOException {
+      int expectedResultSize = 189490;
```
sschuberth, line 59: Maybe again inline this (as "pdf" below is also not put into a variable, and it makes more clear at a single glance what the expected file size is)?

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/utils/TemplateLoaderUtil.java
```diff
@@ -32,9 +33,57 @@ private TemplateLoaderUtil() {
...
+        return load(new TemplateBundle() {
+            @Override
+            public String key() {
+                return "";
+            }
+
+            @Override
+            public InputStream loadTitleTemplate() {
+                try {
+                    return new FileInputStream(cover);
```
sschuberth, line 49: Where are these streams closed? I couldn't find code / a caller that closes these.

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/utils/TemplateLoaderUtil.java
bs-ondem: Actually it's fine to return only an empty string in this case, because the key is only used when loading the template via the ServiceLoader.

modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/utils/TemplateLoaderUtil.java
bs-ondem: These are closed here: https://github.com/eclipse/antenna/blob/master/modules/attribution-document/attribution-document-core/src/main/java/org/eclipse/sw360/antenna/attribution/document/core/Templates.java#L115-L141

PR #547, opened 2020-06-24
Add documentation for how to create template PDF files

### LINKED ISSUE #546: Add documentation for how to create PDF files for the AttributionDocumentGenerator
Currently there is only a documentation for how to create a template bundle JAR for the _AttributeDocumentGenerator_, but there is no documentation for how to create the PDF files for the template bundle JAR. 

### Acceptance Criteria
- [ ] Documentation for how to create PDF files exists.

### Definition of Done
- Documentation is updated

bs-ondem:
Currently, there is only the documentation for the
AttributionDocumentGenerator which describes how to create a template
bundle JAR file. This adds a further documentation for how to create
those template PDF files.

Resolves #546.

### Request Reviewer
@sschuberth 

### Type of Change
*documentation*

### DIFF
--- a/antenna-documentation/src/site/markdown/template-bundle-development.md
+++ b/antenna-documentation/src/site/markdown/template-bundle-development.md
@@ -4,6 +4,23 @@ automatically when no other bundle is configured. This guide will show how to cr
 integrate it in the Antenna workflow run. 
 
 ## Custom Template Bundle
+
+### How to create the template PDF files
+
+First, the template has to be developed with a tool like Microsoft Word / Open Office / Libre Office. The corresponding 
+document will consist of the four pages cover, copyright, content and back page. The page size is A4 at 72 PPI with 
+595x842 pixels. The first page is the cover which can be designed with logos, corporate information etc. but it should 
+be free of any text at the pixel 80x547 for the project's name and version. The second page is the copyright page to 
+which a header and/or footer can be added, but it should be free of any text at the pixel 70x659 for the project's 
+copyright notes. The third page is the content page to which a header and/or footer can be added, but it should be free 
+of any text from the pixel 40x80 until the page bottom margin 40x762 for the list of the dependencies with PURL, file 
+name, license and copyright notes. The last page is the back page which can be designed according to your own needs 
+because this page will simply be attached to the attribution document without any additions.
+
+After the document is finished, each individual page is exported as the respective template PDF file. The first page of 
+the document has to be exported to a PDF file named cover.pdf, the second page to a PDF file named copyright.pdf, the
+third page to a PDF file named content.pdf and the last page to a PDF file named back.pdf. 
+
 ### Basic folder structure
 
 ```


antenna-documentation/src/site/markdown/template-bundle-development.md
```diff
@@ -4,6 +4,24 @@ automatically when no other bundle is configured. This guide will show how to cr
...
+* content.pdf 
+* back.pdf
+
+The page size is A4 at 72 PPI with 595x842 pixels. The cover.pdf file will be the cover page of the attribution document 
+which will contain the project name and version starting at 80x547. The copyright.pdf file will be the second page of 
+the attribution document which will contain the project copyright notes starting at 70x659. The content.pdf file will be
+used several times which will be list the used dependencies with PURL, file name, license and copyright notes. The 
+back.pdf file is the last page of the attribution document which is appended without any changes.  
+
+After the document is finished, each individual page is exported as the respective template PDF file.
```
sschuberth, line 23: I believe it's a bit confusing to speak about PDF files before this sentence. The workflow as I understand it is that first do the full design in Word, not knowing / caring that you would export to PDF later. You basically only care about the paper size and the number of pages at this point. And only at the very end I'd explain that you need to export the first page to a PDF file named `cover.pdf`, the second page to a PDF called `copyright.pdf` and so on.

antenna-documentation/src/site/markdown/template-bundle-development.md
```diff
@@ -4,6 +4,24 @@ automatically when no other bundle is configured. This guide will show how to cr
...
+
+First, the template has to be developed with a tool like Microsoft Word / Open Office / Libre Office. The corresponding 
+document will consist of 4 pages which will represent the 4 templates PDF files:
+* cover.pdf
+* copyright.pdf 
+* content.pdf 
+* back.pdf
+
+The page size is A4 at 72 PPI with 595x842 pixels. The cover.pdf file will be the cover page of the attribution document 
+which will contain the project name and version starting at 80x547. The copyright.pdf file will be the second page of
```
sschuberth, line 18: Ah, so the contents of the cover page are filled in at a fixed *pixel* position? I always thought the pages would need to contain special makers, like `$REPLACE_ME_WITH_THE_PROJECT_NAME_AND_VERSION`, which get replaced with the contents...

antenna-documentation/src/site/markdown/template-bundle-development.md
```diff
@@ -4,6 +4,24 @@ automatically when no other bundle is configured. This guide will show how to cr
...
+document will consist of 4 pages which will represent the 4 templates PDF files:
+* cover.pdf
+* copyright.pdf 
+* content.pdf 
+* back.pdf
+
+The page size is A4 at 72 PPI with 595x842 pixels. The cover.pdf file will be the cover page of the attribution document 
+which will contain the project name and version starting at 80x547. The copyright.pdf file will be the second page of 
+the attribution document which will contain the project copyright notes starting at 70x659. The content.pdf file will be
+used several times which will be list the used dependencies with PURL, file name, license and copyright notes. The
```
sschuberth, line 20: Is there also a pixel position for this file?

antenna-documentation/src/site/markdown/template-bundle-development.md
sschuberth: How do you control the font properties (type, size, style) of the text being inserted?

antenna-documentation/src/site/markdown/template-bundle-development.md
bs-ondem: Yes, it's written on the page at 40,80 until the page bottom margin is reached.

antenna-documentation/src/site/markdown/template-bundle-development.md
sschuberth: > The font type will be loaded from the pdf file

Ok, so that hardens my assumption that the cause of the [font issues](https://github.com/oss-review-toolkit/ort/issues/2755) we're seeing are caused by not all characters of the font being embedded into the PDF.

> but the size and style are hardcoded.

Please add this bit to the docs then.

antenna-documentation/src/site/markdown/template-bundle-development.md
sschuberth: Please add this bit to the docs then.

antenna-documentation/src/site/markdown/template-bundle-development.md
bs-ondem: I rewrite the section before this sentence without mentioning PDF files.

sschuberth (changes requested):

sschuberth (approved):

robertklaus-bosch:
Some remarks from my side.

I think the description by @bs-ondem is a quite exact specification. However I think it is tricky for users to actually create a template based on this. This is the reason, the example bundle exists which also features a basic odt file which can be used.
So I would suggest to keep the specification but not as number one description. Instead .. why not use the example (specification by example if you like)?

And if you want to polish the text a little more, I think it is worth to structure the text block into different things. Starting with a section on what is needed for the bundle (which pages .. what are they for) then a section how to create them (e.g. starting with the example and then a section on the exact specification.

PR #553, opened 2020-07-03
Add configuration to filter ort packages for exclusion

blaumeiser-at-bosch:

### Request Reviewer
@bs-ondem @sschuberth 

### Type of Change

### How Has This Been Tested?
Regression Test

### Checklist
Must:

### DIFF
--- a/modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
+++ b/modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
@@ -29,6 +29,9 @@ class OrtResultAnalyzer : ManualAnalyzer() {
         private val LOGGER = LoggerFactory.getLogger(OrtResultAnalyzer::class.java)
     }
 
+    private val keepExcludedArtifactsProperty = "keep.excluded.artifacts"
+    private var omitExcludedArtifacts = true
+
     init {
         workflowStepOrder = 700
     }
@@ -36,6 +39,11 @@ class OrtResultAnalyzer : ManualAnalyzer() {
     override fun yield() = WorkflowStepResult(createArtifactList(componentInfoFile))
 
     override fun getName() = "OrtResult"
+
+    override fun configure(configMap: Map<String, String>) {
+        super.configure(configMap)
+        omitExcludedArtifacts = !getBooleanConfigValue(keepExcludedArtifactsProperty, configMap)
+    }
 
     @Throws(IOException::class)
     fun createArtifactList(ortResultFile: File): Collection<Artifact> {
@@ -45,6 +53,6 @@ class OrtResultAnalyzer : ManualAnalyzer() {
         if (result.analyzer == null) throw IOException("No analyzer run found in ORT result file.")
 
         val resolver = OrtResultArtifactResolver(result)
-        return result.getPackages().map { (pkg, _) -> resolver.apply(pkg) }.toSet()
+        return result.getPackages(omitExcludedArtifacts).map { (pkg, _) -> resolver.apply(pkg) }.toSet()
     }
 }
--- a/antenna-documentation/src/site/markdown/analyzers/ort-result-analyzer-step.md
+++ b/antenna-documentation/src/site/markdown/analyzers/ort-result-analyzer-step.md
@@ -15,10 +15,14 @@ Add the following step into the `<analyzers>` section of your workflow.xml
                     <configuration>
                         <entry key="base.dir" value="${project.basedir}" />
                         <entry key="file.path" value="ort/scan-result.yml" />
+                        <entry key="keep.excluded.artifacts" value="false" />
                     </configuration>
         </step>
 ```
 
 #### Explanation of parameters
 * `base.dir`: the base dir
-* `file.path`: Relative path to the ort result file. 
+* `file.path`: Relative path to the ort result file.
+* `keep.excluded.artifacts`: Optional parameter, as default the result analyzer enters only packages
+  into the Antenna model which belong to package manager scopes that were not excluded in the ort run.
+  If set to 'true', all packages, even those excluded will be added.


modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -29,6 +29,9 @@ class OrtResultAnalyzer : ManualAnalyzer() {
         private val LOGGER = LoggerFactory.getLogger(OrtResultAnalyzer::class.java)
     }
 
+    private val filterExcludedArtifactsProperty = "filter.excluded"
```
sschuberth, line 32: I'm always a bit confused about whether "to filter" means "to keep something" or "to get rid of something". If we follow the logic of [Kotlin's filter](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter.html), it "returns a list containing only elements matching the given predicate", so here "to filter" means "to keep the filtrate". Further following that logic means that setting `filter.excluded = true` should *keep* (only) the excludes, which is not what we want here.

TL;DR, I propose to rename this property to "omit.excludes" to 1) avoid this confusion and 2) align with ORT's naming of the parameter. Further, I'd argue we could / should set the default to `true`, because in the context of creating reports, I'd say you always want to omit excluded packages.

blaumeiser-at-bosch: Good point, followed your proposal in the new commit

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -29,6 +29,9 @@ class OrtResultAnalyzer : ManualAnalyzer() {
         private val LOGGER = LoggerFactory.getLogger(OrtResultAnalyzer::class.java)
     }
 
+    private val omitExcludedArtifactsProperty = "omit.excluded"
+    private var omitExcludedArtifacts = false
```
sschuberth, line 33: Didn't we also agree to set this default to `true`, or did I misunderstand something?

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -37,6 +40,11 @@ class OrtResultAnalyzer : ManualAnalyzer() {
 
     override fun getName() = "OrtResult"
 
+    override fun configure(configMap: MutableMap<String, String>?) {
+        super.configure(configMap)
+        omitExcludedArtifacts = getBooleanConfigValue(omitExcludedArtifactsProperty, configMap)
```
sschuberth, line 45: Hmm, it just occurred to me that `getBooleanConfigValue()` seem to hard-code the default value to `false`, so it seems we *cannot* have a default of `true` for the property, as this function always gets called (?) and therefore if does not matter which default value we assign to `omitExcludedArtifacts` above...

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
sschuberth: Also, if `configure()` is guranteed to *always* be called, and always *before* `createArtifactList()`, then we should make this a `private lateinit val omitExcludedArtifacts: Boolean`.

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -37,6 +40,11 @@ class OrtResultAnalyzer : ManualAnalyzer() {
 
     override fun getName() = "OrtResult"
 
+    override fun configure(configMap: MutableMap<String, String>?) {
```
sschuberth, line 43: The signature should ideally be `configMap: Map<String, String>` if `configMap` is never written to and is guaranteed to never be `null`.

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
sschuberth: We can probably fix this by doing:

    omitExcludedArtifacts = getConfigValue(omitExcludedArtifactsProperty, configMap, "true").toBoolean()

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
blaumeiser-at-bosch: I reverse the logic, so that you have to configure this, if you do not want the scope reduced

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
blaumeiser-at-bosch: See above

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -29,6 +29,9 @@ class OrtResultAnalyzer : ManualAnalyzer() {
         private val LOGGER = LoggerFactory.getLogger(OrtResultAnalyzer::class.java)
     }
 
+    private val disableScopeReductionProperty = "disable.scope.reduction"
```
sschuberth, line 32: So, if you want to invert the logic here, how about simply calling this "keep excluded artifacts"?

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -37,6 +40,11 @@ class OrtResultAnalyzer : ManualAnalyzer() {
 
     override fun getName() = "OrtResult"
 
+    override fun configure(configMap: Map<String, String>?) {
```
sschuberth, line 43: I still believe we can safely make this a non-nullable type (as I already did [here](https://github.com/eclipse/antenna/blob/df5d05ad192e19415dc6e00a99c940439ae1b784/modules/ort/src/main/kotlin/workflow/processors/enricher/OrtDownloaderProcessor.kt#L30); the fix to also make the map immutable is [here](https://github.com/eclipse/antenna/pull/554)) as in all places I've seen so far an empty map is passed instead of `null`.

modules/ort/src/main/kotlin/workflow/analyzers/OrtResultAnalyzer.kt
```diff
@@ -37,6 +40,11 @@ class OrtResultAnalyzer : ManualAnalyzer() {
...
+        super.configure(configMap)
+        omitExcludedArtifacts = ! getBooleanConfigValue(keepExcludedArtifactsProperty, configMap)
```
sschuberth, line 45: Nit: I'm not sure about the style for Antenna, but I'd expect to be no space after `!`.

PR #557, opened 2020-07-09
Oheger bosch/sw360/#426 licenses

oheger-bosch:
Issue 426

This PR is about fine-tuning the license part of the SW360 client library.

The API was slightly reworked, new functionality to create licenses was added, Javadoc has been improved.

As a side-effect, access to licenses in SW360MetaDataUpdater has been reworked: the updater no longer asks the license adapter for each license (which was an expensive operation); rather, the licenses IDs available are loaded once and then cached locally.

@neubs-bsi 

*Type of change*:  
improvements

Existing test cases have been adapted. For new functionality, new tests have been added.

### Checklist
Must:

- [X] I have provided tests for the changes (if there are changes that need additional tests)

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
@@ -42,20 +42,17 @@ class SW360LicenseClientAdapterAsyncImpl implements SW360LicenseClientAdapterAsy
     }
 
     @Override
-    public CompletableFuture<Boolean> isLicenseOfArtifactAvailable(String license) {
-        return getLicenseClient().getLicenses()
-                .thenApply(sw360Licenses -> sw360Licenses.stream()
-                        .map(SW360SparseLicense::getShortName)
-                        .anyMatch(n -> n.equals(license)));
-    }
-
-    @Override
-    public CompletableFuture<Optional<SW360License>> getSW360LicenseByAntennaLicense(String license) {
+    public CompletableFuture<Optional<SW360License>> getLicenseByName(String license) {
         return optionalFuture(getLicenseClient().getLicenseByName(license));
     }
 
     @Override
-    public CompletableFuture<Optional<SW360License>> getLicenseDetails(SW360SparseLicense sparseLicense) {
-        return optionalFuture(getLicenseClient().getLicenseByName(sparseLicense.getShortName()));
+    public CompletableFuture<SW360License> enrichSparseLicense(SW360SparseLicense sparseLicense) {
+        return getLicenseClient().getLicenseByName(sparseLicense.getShortName());
+    }
+
+    @Override
+    public CompletableFuture<SW360License> createLicense(SW360License license) {
+        return getLicenseClient().createLicense(license);
     }
 }
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
@@ -15,6 +15,7 @@ import org.eclipse.sw360.antenna.sw360.client.rest.SW360LicenseClient;
 import org.eclipse.sw360.antenna.sw360.client.rest.resource.licenses.SW360License;
 import org.eclipse.sw360.antenna.sw360.client.rest.resource.licenses.SW360SparseLicense;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
@@ -33,6 +34,11 @@ class SW360LicenseClientAdapterAsyncImpl implements SW360LicenseClientAdapterAsy
     @Override
     public SW360LicenseClient getLicenseClient() {
         return licenseClient;
+    }
+
+    @Override
+    public CompletableFuture<List<SW360SparseLicense>> getLicenses() {
+        return getLicenseClient().getLicenses();
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdaterTest.java
```diff
@@ -148,7 +177,6 @@ public void testGetLicensesWithExistingLicense() {
 
         assertThat(licenses).hasSize(1);
 
-        verify(licenseClientAdapter, times(1)).isLicenseOfArtifactAvailable(licenseName);
```
neubs-bsi, line 151: You could add an assertion here, that licenseclientadpater getLicenses is only called once.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdaterTest.java
```diff
@@ -159,17 +187,15 @@ public void testGetLicensesWithNonExistingLicense() {
...
+                .thenReturn(createTestLicenses());
         when(licenseClientAdapter.getSW360LicenseByAntennaLicense(licenseName))
                 .thenReturn(Optional.of(license));
         setUp(true, false);
 
         final Set<SW360License> licenses = metaDataUpdater.getLicenses(Collections.singletonList(licenseAntenna));
 
         assertThat(licenses).hasSize(0);
-
-        verify(licenseClientAdapter, times(1)).isLicenseOfArtifactAvailable(licenseName);
```
neubs-bsi, line 172: Same: Assertion that getLicenses of LicenseClientAdapter is only called once

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
```diff
@@ -36,20 +37,22 @@ public SW360LicenseClient getLicenseClient() {
...
     @Override
-    public CompletableFuture<Optional<SW360License>> getLicenseDetails(SW360SparseLicense sparseLicense) {
-        return optionalFuture(getLicenseClient().getLicenseByName(sparseLicense.getShortName()));
+    public CompletableFuture<SW360License> enrichSparseLicense(SW360SparseLicense sparseLicense) {
+        return getLicenseClient().getLicenseByName(sparseLicense.getShortName());
+    }
+
+    @Override
+    public CompletableFuture<SW360License> createLicense(SW360License license) {
+        return getLicenseClient().createLicense(license);
```
neubs-bsi, line 56: Whilst this is currently only called in tests, when using it we must be aware that there must be a check whether the license contains a fullname (the only required attribute that needs to be set in a SW360 License object). Currently we most of the time use the short name attribute in Antenna functionalities. The full name is currently only ever set in tests.

neubs-bsi: Additionally, whilst I agree that this method should be included into the sw360-client library, should we ever use it in an antenna context, we need to be careful not to write bogus licenses into our instance, as has happened before.

oheger-bosch: Regarding the fullname attribute: At least SW360 rejects requests to add a license without a full name. We could add a check already in the client library to prevent server calls that will fail anyway. I am, however, not sure how far we should go with duplicating validation logic. There is always the risk that the client validation gets out of sync with the server validation. Letting SW360 reject the call has the advantage that the server sends an official error message, which currently looks like this:
```
{
    "status": 500,
    "error": "Internal Server Error",
    "message": "Required field 'fullname' was not present! Struct: License(id:TST-3, type:license, shortname:TST-3, fullname:null, GPLv2Compat:UNDEFINED, GPLv3Compat:UNDEFINED, text:License to test)",
    "timestamp": "2020-07-10T07:17:41.045Z"
}
```

I am, however, not sure whether this message is actually made available by the client library.

Regarding the creation of licenses from Antenna, this is a good point. I assume that Antenna will use the licenses in SW360 as a read-only source; otherwise, it has to be defined exactly under which conditions a new license would have to be created.

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdaterTest.java
oheger-bosch: Done

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdaterTest.java
oheger-bosch: Done

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/adapter/SW360LicenseClientAdapterAsyncImpl.java
neubs-bsi: In my opinion the message given by SW360 looks good. I see two options:
- if the message isn't given to the user (because of error handling or sth) we should change it so that it is displayed
- make our own check and throw an error message

However to the 2nd option: I agree that we should avoid duplicated validations

oheger-bosch: I have created an issue to make the error message from the server available: https://github.com/eclipse/antenna/issues/558

Issue #365, opened 2019-11-29
Container Deployment for Antenna


blaumeiser-at-bosch:
### Summary of the Improvement
Support a simple deployment as a docker container that can be easily integrated into a CI toolchain.

### Acceptance Criteria
- [ ] A Docker Container can be created that contains an configurable Antenna deployment (e.g. also proprietary additions of Antenna)
- [x] The container can be executed and gets access to a locally checkout project and runs the configured antenna execution in this environment
- [x] The antenna configuration is taken from the local project that is mapped into the container

### Definition of Done
- Acceptance criteria fulfilled
- The ticket describes what has been done and describes how the results can be verified (e.g. new Jenkins step is inserted, the report can be viewed at X)
- All known limitations are listed and it is described how they will be addressed (e.g., a ticket for the remaining stuff)
- Documentation is updated
- The result has been reviewed: Everything works as described and can be verified reading only the documentation

neubs-bsi:
A container can be created too which an Antenna configuration can be mounted. 
Not supported: Proprietary additions of Antenna.

PR #560, opened 2020-07-15
Dockerfile #365

neubs-bsi:

This PR introduces a Dockerfile with which an image can be build that executes the antenna cli. 

Issue: #365

Current limitation:
- cache is not fully used. For maven projects in docker it is possible to first download all dependencies from the pom before copying the rest of the source code and building the maven project. This makes the build very slow every time the build is executed. 
    - problems with the above descripted approach is the multi-module approach antenna has, makes this more difficult. Additonally, a plugin dependency is not resolved against its pluginRepository correctly, which needs to be investigated. A mounting of a local maven repository could be considered but was not tested. This could be a new issue. A first functioning Dockerfile that meets the issue criteria is introduced here. 

@oheger-bosch 

*Type of change*:  CI

No tests as of yet. I could add a CI step that builds the image, however that would be VERY time consuming. 

### Checklist
Must:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,12 +10,13 @@
 #
 # Build stage
 #
-FROM maven:3.3.9-jdk-8-alpine AS ANTENNA_MAVEN_BUILD
+FROM maven:3.5.0-jdk-8-alpine AS ANTENNA_MAVEN_BUILD
 
 COPY . /home/antenna/
 WORKDIR /home/antenna/
 RUN mkdir /maven
-RUN mvn -Dmaven.repo.local=/maven install -DskipTests -Dskip.pmd=true -pl '!core/frontend-stubs-testing'
+
+RUN mvn -B -e -Dmaven.repo.local=/maven install -DskipTests=true -Dskip.pmd=true
 
 #
 # Package stage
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+#
+# Package stage
+#
+FROM openjdk:8-jre-slim
+WORKDIR ../..
+COPY --from=ANTENNA_MAVEN_BUILD /home/antenna/assembly/cli/target/antenna.jar /usr/local/lib/antenna.jar
+
+COPY /example-projects/example-project/pom.xml /antenna/antennaConfiguration.xml
+COPY /example-projects/example-project/src /antenna/src
```
oheger-bosch, line 34: Are these two COPY statements needed (maybe to set defaults)? I would expect that the project using the image would provide the content via a mount.

antenna-documentation/src/site/markdown/docker.md
```diff
@@ -0,0 +1,43 @@
...
+This mounts the given configuration folder in the docker container where the antenna configuration file 
+is searched for by the docker cmd. In this folder all information the configuration file and workflow need
+from the user should be supplied. 
+
+Hence, if you want to have an Antenna run over your whole project, you should mount that project. 
+In this configuration the antenna configuration file should always be at the root of this folder.  
+
+The resulting files will be saved in the output directory you specified. 
+This output directory needs to be the folder or a subfolder of the mounted folder. 
+The docker container does not have access to anything outside of the mounted folder.
```
oheger-bosch, line 43: The description of running the image is already nice. But I think it would be even more helpful to show the usage based on a more concrete example. For instance, describe that there is a project in a folder /data/my-project, which contains a src folder and the Antenna config file (maybe in a sub folder). Then show the concrete docker command how to execute Antenna on this project.

Dockerfile
neubs-bsi: I thought to place it there as a default. 
So when someone gets the image, they can test the docker run with a default test to see if it works. 

But you are right, it is not necessary. 
I don't believe it increases the image size by a lot, though.

oheger-bosch: Thanks for clarifying. That's fine then.

Issue #544, opened 2020-06-23
SW360-client: Component Homepage URL is never set


neubs-bsi:
### Summary of the Bug
thanks to the ort analyzer it is possible to have home page url information for an artifact. 
However when translating an artifact to release that information is lost. 
This leads to the issue, that when a new component is created, which information is derived from a release, the homepage url is not set. 

### Steps to reproduce
- Check component creation in ComponentAdapterUtils and see that it is not set. 

### Acceptance Criteria
- [ ] All information from the ort result analyzer is reflected in the component that would be available

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

neubs-bsi:
Closed with #566

PR #566, opened 2020-07-31
Neubs bsi/sw360/component/homepageurl

neubs-bsi:

Issue: #544 
Homepage Url is now set when component is created. 

@oheger-bosch 

*Type of change*:  bug fix

Adapted Tests

### Checklist
Must:

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
@@ -38,6 +38,7 @@ public final class SW360Release extends SW360HalResource<SW360ReleaseLinkObjects
     private static final String CHANGESTATUS_KEY = "change_status";
     private static final String COPYRIGHTS_KEY = "copyrights";
     private static final String CLEARINGSTATE_KEY = "clearingState";
+    private static final String HOMEPAGE_KEY = "homepage";
 
     @JsonIgnore
     private boolean isProprietary;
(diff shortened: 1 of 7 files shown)


modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
```diff
@@ -291,6 +293,18 @@ public SW360Release setProprietary(boolean proprietary) {
...
+        return homepageurl;
+    }
+
+    @JsonIgnore
+    public SW360Release setHomepageurl(String homepageurl) {
+        this.homepageurl = homepageurl;
+        return this;
+    }
+
+
```
oheger-bosch, line 307: As SW360Release is a representation of a domain object in SW360, I find it a bit strange that it has a property not supported by SW360. Wouldn't it be better to use the _additional data_ mechanism provided by this class for this purpose? That way the homepage URL for the release would even be persisted in SW360.

neubs-bsi: That is a good idea, thanks.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
```diff
@@ -38,9 +38,12 @@
     private static final String CHANGESTATUS_KEY = "change_status";
     private static final String COPYRIGHTS_KEY = "copyrights";
     private static final String CLEARINGSTATE_KEY = "clearingState";
+    private static final String HOMEPAGE_KEY = "homepage";
 
     @JsonIgnore
     private boolean isProprietary;
+    @JsonIgnore
+    private String homepageurl;
     private String name;
```
oheger-bosch, line 47: The field should now be removed; it is not assigned any more. Also the references from equals() and hashCode() are no longer needed.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/rest/resource/releases/SW360Release.java
```diff
@@ -291,6 +294,17 @@ public SW360Release setProprietary(boolean proprietary) {
         return this;
     }
 
+    @JsonIgnore
+    public String getHomepageurl() {
```
oheger-bosch, line 298: One minor thing: Should the name of the property be homepageUrl rather than homepageurl? Camel case is probably more readable.

PR #568, opened 2020-08-07
added sw360 token integration for antenna

ag4ums:
Signed-off-by: Anupam Ghosh <anupam.ghosh@siemens.com>

This PR adds antenna connection with sw360 over user token

### Type of Change
 > improvements

*Type of change*:  

### Checklist
Must:

### DIFF
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProvider.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProvider.java
@@ -70,8 +70,10 @@ public class AccessTokenProvider {
      *
      * @param authClient the underlying authentication client
      */
-    public AccessTokenProvider(SW360AuthenticationClient authClient) {
+    public AccessTokenProvider(SW360AuthenticationClient authClient, String token) {
         this.authClient = authClient;
+        this.accessToken = new AccessToken(token);
+        this.tokenFuture = CompletableFuture.completedFuture(this.accessToken);
     }
 
     /**
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProvider.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProvider.java
@@ -70,10 +70,8 @@ public class AccessTokenProvider {
      *
      * @param authClient the underlying authentication client
      */
-    public AccessTokenProvider(SW360AuthenticationClient authClient, String token) {
+    public AccessTokenProvider(SW360AuthenticationClient authClient) {
         this.authClient = authClient;
-        this.accessToken = new AccessToken(token);
-        this.tokenFuture = CompletableFuture.completedFuture(this.accessToken);
     }
 
     /**
--- a/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/config/SW360ClientConfig.java
+++ b/modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/config/SW360ClientConfig.java
@@ -74,6 +74,11 @@ public final class SW360ClientConfig {
     private final String clientPassword;
 
     /**
+     * The SW360 user token credentials.
+     */
+    private final String token;
+
+    /**
      * The HTTP client to interact with the SW360 server.
      */
     private final HttpClient httpClient;
@@ -84,13 +89,14 @@ public final class SW360ClientConfig {
     private final ObjectMapper objectMapper;
 
     private SW360ClientConfig(URI baseURI, String authURL, String user, String password, String clientId,
-                              String clientPassword, HttpClient httpClient, ObjectMapper objectMapper) {
+                              String clientPassword, String token, HttpClient httpClient, ObjectMapper objectMapper) {
         this.baseURI = baseURI;
         this.authURL = authURL;
         this.user = user;
         this.password = password;
         this.clientId = clientId;
         this.clientPassword = clientPassword;
+        this.token = token;
         this.httpClient = httpClient;
         this.objectMapper = objectMapper;
     }
@@ -115,7 +121,7 @@ public final class SW360ClientConfig {
      */
     public static SW360ClientConfig createConfig(String restURL, String authURL, String user,
                                                  String password, String clientId, String clientPassword,
-                                                 HttpClient httpClient, ObjectMapper mapper) {
+                                                 String token, HttpClient httpClient, ObjectMapper mapper) {
         return new SW360ClientConfig(
                 URI.create(stripTrailingSeparator(Validate.notEmpty(restURL, "Undefined REST URL"))),
                 stripTrailingSeparator(Validate.notEmpty(authURL, "Undefined authentication URL")),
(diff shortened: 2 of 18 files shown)


oheger-bosch:
Hello @ag4ums, many thanks for your contribution. Adding support for user tokens is a nice feature. If I understand correctly, such a user token is constant; it is once configured in the Antenna configuration and then has to be used for all requests, correct?

Unfortunately, this PR causes a number of tests for _AccessTokenProvider_ to fail. This is probably because the token provider is now initialized with the configured token; so it will return this token on first access rather than sending a request to retrieve a new one as expected by the tests.

May I propose an alternative implementation? Do not modify _AccessTokenProvider_, but change _SW360AuthenticationClient_ instead. The latter has access to an _SW360ClientConfig_ and thus can obtain the user token from there. In the getOAuth2AccessToken() method the client should check whether a user token has been configured. If so, it returns a future with this token; otherwise, it falls back to the current implementation, i.e. retrieving the token from the OAuth server.

I think, this approach should not impact the existing test cases. It also has the advantage that it does not cause the _AccessTokenProvider_ to return an invalid token initially if no user token is available.

I will add some more comments to the changed files.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/auth/AccessTokenProvider.java
```diff
@@ -70,8 +70,10 @@
      *
      * @param authClient the underlying authentication client
      */
-    public AccessTokenProvider(SW360AuthenticationClient authClient) {
+    public AccessTokenProvider(SW360AuthenticationClient authClient, String token) {
```
oheger-bosch, line 73: See my other comment; do not change this class.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/config/SW360ClientConfig.java
```diff
@@ -115,14 +121,15 @@ private SW360ClientConfig(URI baseURI, String authURL, String user, String passw
...
+                                                 String token, HttpClient httpClient, ObjectMapper mapper) {
         return new SW360ClientConfig(
                 URI.create(stripTrailingSeparator(Validate.notEmpty(restURL, "Undefined REST URL"))),
                 stripTrailingSeparator(Validate.notEmpty(authURL, "Undefined authentication URL")),
                 Validate.notEmpty(user, "Undefined user"),
                 Validate.notEmpty(password, "Undefined password"),
                 Validate.notEmpty(clientId, "Undefined client ID"),
                 Validate.notEmpty(clientPassword, "Undefined client password"),
+                Validate.notEmpty(token, "Undefined token"),
                 Validate.notNull(httpClient),
```
oheger-bosch, line 133: We now support multiple ways of authentication that make use of different properties; therefore, we should force that all these properties are defined. The validation logic should be adapted to check that either user name and password are defined or the access token (and probably not both).

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/config/SW360ClientConfigTest.java
```diff
@@ -46,85 +47,85 @@ public void setUp() {
 
     @Test(expected = NullPointerException.class)
     public void testNullRestUrlThrows() {
-        SW360ClientConfig.createConfig(null, AUTH_URL, USER, PASSWORD, CLIENT_ID, CLIENT_PASS, httpClient, mapper);
+        SW360ClientConfig.createConfig(null, AUTH_URL, USER, PASSWORD, CLIENT_ID, CLIENT_PASS, USER_TOKEN, httpClient, mapper);
```
oheger-bosch, line 50: These tests will have to be adapted to comply to the new validation logic.

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360ClientTokenManagementIT.java
```diff
@@ -79,6 +86,7 @@ private static String accessTokenBody(String token) {
      * token is made.
      */
     @Test
+    @Ignore
```
oheger-bosch, line 89: Tests should not be ignored.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/SW360ConnectionConfigurationFactory.java
```diff
@@ -72,6 +72,13 @@
      */
     public static final String CLIENT_PASSWORD_KEY = "client.password";
 
+    /**
+     * The name of the property from the Antenna tool configuration defining
+     * the token of the OAuth 2 client to be passed to the authorization
+     * server when requesting an access token.
+     */
+    public static final String TOKEN_KEY = "user.token";
```
oheger-bosch, line 80: It would be great if this new property could be documented together with the other SW360-related properties supported by Antenna. This would have to be done in the files sw360-update-generator-step.md and sw360-enricher.md.

modules/sw360/sw360-client/src/main/java/org/eclipse/sw360/antenna/sw360/client/config/SW360ClientConfig.java
ag4ums: change the validation based on token

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/workflow/SW360ConnectionConfigurationFactory.java
ag4ums: updated the .md files with user.token attribute

ag4ums:
@oheger-bosch update the PR as suggested by you, please take a look.

neubs-bsi (changes requested):
Hello @ag4ums 
many thanks for including the requested review changes. This is already looking very good. 
I have noticed that both your PR description and your commit messages are quite short. 
As can be seen in the [contribution guideline](https://github.com/bosch-io/antenna/blob/master/CONTRIBUTING.md) of antenna, a commit message detailing the changes and the reason for the changes is required. 

There you can detail what you are adding and why certain changes are necessary.

modules/sw360/sw360-client/src/test/java/org/eclipse/sw360/antenna/sw360/client/rest/SW360ClientTokenManagementIT.java
```diff
@@ -43,6 +43,12 @@
  * that sends requests against the mock server.
  */
 public class SW360ClientTokenManagementIT extends AbstractMockServerTest {
+
+    /**
+     * test token for sw360 client.
+     */
+    private static final String USER_TOKEN = "";
```
oheger-bosch, line 50: Is this field needed? I think, it is never referenced.

oheger-bosch (changes requested):
This looks pretty good now. The only thing I am missing is a test case that verifies whether SW360AuthenticationClient now returns the configured token if it is available in the configuration. Would it be possible to add such a test (maybe in SW360AuthenticationClientIT)?

oheger-bosch (approved):

neubs-bsi (approved):

PR #578, opened 2020-08-26
Oheger bosch/build/#577 build on jdk11

oheger-bosch:
Issue 577

This PR makes Antenna build with JDK 11.

The PR adds dependencies to Jaxb (API and runtime), which is required by Antenna, but no longer available on newer JDKs.

To make sure that only Java8-compatible APIs are used even when building with a newer JDK, the Animal-sniffer plugin has been added; it causes the build to fail when it detects usage of classes introduced with Java 9 or later.

@neubs-bsi 

*Type of change*:  
CI

I ran the build with both JDK 8 and 11.
To check Java 8 compatibility, I built with JDK 11 and then run the Antenna test project on JDK 8.

### Checklist
Must:


neubs-bsi (approved):
Looks good. Can be merged if CI is green :)

neubs-bsi (approved):

Issue #561, opened 2020-07-15
AntennaCLIFrontend does not take properties file


neubs-bsi:
### Summary of the Bug
In the commit https://github.com/eclipse/antenna/commit/4a4aefe a possibility was introduced to give a properties file (similar to the compliance tool) containing properties one wishes to set in the Antenna workflow but wishes not to include in the project / repository antenna is running. 

The introduction of the AbstractAntennaCLIOptions in https://github.com/eclipse/antenna/commit/7f96702 seemed to have missed that there could be two paths specified in the command line arguments, which would be valid. 

When a properties file is given to the CLI tool, it does not run

### Steps to reproduce
- create example.properties in example project
- Run Antenna example project with `java -jar <path-to-antenna-jar> pom.xml example.properties`
- tool will not execute
- ...

### Acceptance Criteria
- [ ] The properties file function should still be supported in the Antenna CLI tool. 

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

neubs-bsi:
Assignee: @KorSin

neubs-bsi:
Already done

PR #592, opened 2020-09-28
AntennaCLIFrontend now takes properties file

KorSin:
Issue #561.

Giving a second path to the CLI does not result in a wrong usage message
anymore.

*Type of change*:  bug fix

### Checklist
Must:

### DIFF
--- a/assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIOptions.java
+++ b/assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIOptions.java
@@ -41,6 +41,10 @@ public final class AntennaCLIOptions extends AbstractAntennaCLIOptions {
      */
     private final String configFilePath;
 
+    /**
+     * The path to the file with environment properties.
+     */
+    private final String propertiesFilePath;
 
     /**
      * Creates a new instance of {@code AntennaCLIOptions} with the properties
@@ -52,8 +56,27 @@ public final class AntennaCLIOptions extends AbstractAntennaCLIOptions {
      * @param valid          flag whether the command line is valid
      */
     AntennaCLIOptions(String configFilePath, boolean debugLog, boolean showHelp, boolean valid) {
+        this(configFilePath, null, debugLog, showHelp, valid);
+    }
+
+    /**
+     * Creates a new instance of {@code AntennaCLIOptions} with the properties
+     * provided.
+     *
+     * @param configFilePath     the path to the Antenna config file
+     * @param propertiesFilePath the path to the properties file
+     * @param debugLog           flag whether debug log should be active
+     * @param showHelp           flag whether the help message should be printed
+     * @param valid              flag whether the command line is valid
+     */
+    AntennaCLIOptions(String configFilePath,
+                      String propertiesFilePath,
+                      boolean debugLog,
+                      boolean showHelp,
+                      boolean valid) {
         super(debugLog, showHelp, valid);
         this.configFilePath = configFilePath;
+        this.propertiesFilePath = propertiesFilePath;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


.ci-scripts/test-for-licenseHeaders.sh
```diff
@@ -69,6 +69,7 @@ getFiles() {
 }
 
 while read file ; do
+  echo "$file"
```
neubs-bsi, line 72: if the file fails it is written in the testFile() method, together with a reason. 
This addition is not really bringing any other functionality except for a list of all files tested.
Additonally, this is not relevant to the other changes and should therefore be a separate commit- if kept - I am open to a discussion

assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIOptions.java
```diff
@@ -122,18 +162,20 @@ public boolean equals(Object o) {
...
     public int hashCode() {
-        return Objects.hash(super.hashCode(), getConfigFilePath());
+        return Objects.hash(super.hashCode(), getConfigFilePath(), getPropertiesFilePath());
     }
 
     @Override
     public String toString() {
         return "AntennaCLIOptions{" +
                 "configFilePath='" + configFilePath + '\'' +
+                "propertiesFilePath='" + propertiesFilePath + '\'' +
```
neubs-bsi, line 178: very good, the propertiesFile should also be reflected in the helpMessage: https://github.com/eclipse/antenna/blob/80df3ad50475177ed57537286dc574cca3b95cd6/assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIOptions.java#L144

neubs-bsi (changes requested):
Hey, this is already looking very good.

I have some remarks. Additionally, I would want to point out, that the commit message of the one commit is a bit short. Maybe you can elaborate on the changes and their reasons a bit more.

Additionally the propertiesFile should be mentioned in the documentation, which could be done [here](https://github.com/eclipse/antenna/blob/80df3ad50475177ed57537286dc574cca3b95cd6/assembly/cli/src/site/markdown/index.md.vm)

.ci-scripts/test-for-licenseHeaders.sh
KorSin: I am sorry, this change was unintended. I will remove it.

assembly/cli/src/site/markdown/index.md.vm
```diff
@@ -110,6 +110,12 @@ SECRET=password12345 java -jar path\to\ ${docName}.jar path\to\toolConfiguration
 Should you execute it within the command, the variable will not be saved in your environment but only exist for the run of the command.
 ${docNameCap} adheres to standards and only renders environment variables that are written in upper case letters.
 
+Alternatively, it is possible to specify environment variables inside a properties file which can be given to the analyzer command:
```
neubs-bsi, line 113: Maybe you could mention that in the properties file also non-uppercase variables are taken, if I am not mistaken.

neubs-bsi (approved):

assembly/cli/src/main/java/org/eclipse/sw360/antenna/frontend/cli/AntennaCLIOptions.java
```diff
@@ -54,6 +58,27 @@
     AntennaCLIOptions(String configFilePath, boolean debugLog, boolean showHelp, boolean valid) {
         super(debugLog, showHelp, valid);
         this.configFilePath = configFilePath;
+        this.propertiesFilePath = null;
```
oheger-bosch, line 61: This constructor should simply call the other one with a *null* parameter for the properties path.

assembly/cli/src/site/markdown/index.md.vm
KorSin: Yes you are right, I checked this.

oheger-bosch (approved):

PR #594, opened 2020-09-30
Fix a problem with a platform-specific path conversion.

oheger-bosch:
Issue: eclipse#591.

CSVArtifactMapper tries to convert paths to source files to relative
paths, so that they are not specific to a local configuration.
However, for absolute paths that are not below the base directory,
this obviously does not work reliably on all platforms - and does not
make any sense either. So such paths are now stored as absolute paths.

*Type of change*:  
improvements

A new test case has been added to check the concrete problem.

### Checklist
Must:


blaumeiser-at-bosch (approved):

blaumeiser-at-bosch:
It even builds on my machine ;-)

PR #607, opened 2020-10-13
Korsin/clearing document name

KorSin:

@neubs-bsi 
@oheger-bosch 

*Type of change*:  bug fix

### Checklist
Must:

### DIFF
--- a/assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/updater/SW360UpdaterTest.java
+++ b/assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/updater/SW360UpdaterTest.java
@@ -33,6 +33,7 @@ import java.util.Map;
 import java.util.Objects;
 
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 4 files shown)


assembly/compliance-tool/src/test/java/org/eclipse/sw360/antenna/frontend/compliancetool/sw360/updater/SW360UpdaterTest.java
```diff
@@ -61,38 +58,21 @@ public void testClearingReportGeneratorMustNotBeNull() {
...
+                .thenThrow(new SW360ClientException("Clearing doc generation error"));
         SW360Updater updater = new SW360Updater(updaterImpl, configurationMock, clearingReportGenerator);
 
-        // and an exception is nevertheless thrown by SW360ReleaseClientAdapterAsyncImpl
         Assertions.assertThatThrownBy(updater::execute)
                 .isInstanceOf(SW360ClientException.class);
 
-        // but we expect that attachment upload worked one time exactly
-        verify(releaseClientAdapter, times(1)).uploadAttachments(any());
+        verify(updaterImpl, never()).artifactToReleaseWithUploads(any(), any(), anyMap());
```
oheger-bosch, line 75: One major goal of this test was to verify whether exceptions thrown during the processing of an artifact are caught and do not cause the process to abort. Therefore, the clearing report generator was configured to throw an error once and then succeed. In the second iteration, a successful upload should be performed.

With the reworked test, this is no longer the case. Was there a reason to strip down the test in this way?

KorSin: From the updater, I removed the, in my opinion, redundant second upload of attachments. But with that the method `uploadAttachments` is never called by the `SW360Updater` anymore. Consequently, I removed this from the test and changed it to verify that the method `artifactToReleaseWithUploads` is never called. Furthermore, I removed other mocking calls that are not called anymore.

KorSin: After a discussion with @nnobelis I finally understood the intention of the test. Please have a look at the changes.

oheger-bosch (approved):
Looks good now. Thank you!
I will merge when CI is green.

PR #611, opened 2020-10-19
Replacing standard java zip package with apache compress. 

### LINKED ISSUE #601: SourceZipWriter does not take the most current file from central directory of zip
### Summary of the Bug
If an archive contains several entries of the same name, the registry should be able to decide which entry is actually to be used. Currently the sourcesZipWriter simply logs a [warning](https://github.com/eclipse/antenna/blob/master/core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImpl.java#L134-L139), which comes from catching a `ZipException`, stating that there is a duplicate entry in the currently source file. 

This is due to the `java.utils.zip` library not having the functionality to decide which entry should be used if there are several of the same name. The library [commons-compress supports this functionality](https://commons.apache.org/proper/commons-compress/zip.html#ZipArchiveInputStream_vs_ZipFile). 

This issue was discovered when a user had a  curated source file that contained a java class twice (https://mvnrepository.com/artifact/com.mchange/c3p0/0.9.5.3). The warning `[WARNING] duplicate entry: c3p0-0.9.5.3/com/mchange/v2/c3p0/subst/C3P0Substitutions.java: caused by File: Optional[/home/stephanie/Projects/antenna_projects/osm-tina-extensions/example-projects/example-extensions-project/target/antenna/dependencies/c3p0-0.9.5.3-sources.jar]` was logged.  

### Steps to reproduce
- Use curated version of above mentioned dependency 
- Run antenna with an analyzer that takes sources and the `SourceZipWriter` 
- Above mentioned warning log should be printed

### Acceptance Criteria
- [ ] Investigate if error does still occurs with [commons-compressed](https://mvnrepository.com/artifact/org.apache.commons/commons-compress)
- [ ] If it can handle duplicate entries, switch to commons-compressed instead of  `java.utils.zip`

### Definition of Done
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

KorSin:
Resolves #601

java.utils.zip throws an exception when zipping a jar with duplicate
entries. This is not the case for the the compress package. Thus, the
java standard is replaced. Furthermore, ZipExceptions are not caught
anymore, since they are not pressent in the compress package and are
also derived from IOExcetption but are already taken care of.

Signed-off-by: Korbinian Singhammer <external.Korbinian.Singhammer2@bosch.io>

@oheger-bosch 
@neubs-bsi 

bug fix

*Type of change*:  

### Checklist
Must:


core/core-workflow-steps/src/test/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+        final Artifact artifact = new Artifact();
+        final ArtifactSourceFile sourceFile = new ArtifactSourceFile(Paths.get(pathToJarFile));
+        artifact.addFact(sourceFile);
+        final List<Artifact> artifacts = Arrays.asList(artifact);
+        sourceZipWriter.createZip(artifacts);
+
+        verify(processingReporterMock, never()).add(eq(artifact),
+                eq(MessageType.PROCESSING_FAILURE),
+                contains("duplicate entry"));
+    }
```
oheger-bosch, line 70: Not sure whether it is possible with reasonable effort, but would it be possible to check whether the duplicate entry actually exists in the target folder in the correct version?

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImpl.java
```diff
@@ -131,12 +135,6 @@ private void addContentToZip(Artifact artifact, ZipOutputStream zipOut) {
             try {
                 LOGGER.debug("Writing File: {}", entryName);
                 writeContentToZipEntry(zipOut, sourceFile.get().toFile(), entryName);
-            } catch (ZipException e) {
-                if (!e.getMessage().equals("zip file is empty")) {
-                    reporter.add(artifact, MessageType.PROCESSING_FAILURE,
-                            e.getMessage() + ": caused by File: " + sourceFile);
-                    LOGGER.warn("{}: caused by File: {}", e.getMessage(), sourceFile);
-                }
```
oheger-bosch, line 139: There was some special behavior for empty Zip files. Has it been tested that Commons Compress behaves in the same way here? Just asking because I am a bit sceptic regarding the existing test coverage.

oheger-bosch (commented):
Two minor questions or proposals for improvements, but I am fine with the changes.

core/core-workflow-steps/src/test/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImplTest.java
KorSin: You are right, this should be tested. It turned out that it is not difficult to implement an will therefore be added.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImpl.java
oheger-bosch: I was referring to the code fragment in the catch clause that has been deleted. Here exceptions with the message "zip file is empty" are obviously ignored; so I would assume that such files have caused problems in the past. It is pretty sure that Commons Compress will use different exception messages.

core/core-workflow-steps/src/test/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImplTest.java
oheger-bosch: Hm, what I meant here was to check that the deflating of the archive worked correctly. So that afterwards on the local disk the file corresponding to the duplicate entry is now present with the expected content. Sorry if this was not clear before.

KorSin: What I now did, is to scan through the zip after its creation and check if the duplicate name appears (at least) twice. If I do this with java standard zip, this fails. Do you think this is not enough? I could also compare the content.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImpl.java
KorSin: I see what you mean. I wrote a test with an empty file. Apache compress indeed throws a ZipException. However, the message is different. Instead of "zip file is empty" it says, "Archive is not a ZIP archive". Do you think we should just change this message? I am not quite sure, since the latter message could also mean other things than an empty file.

oheger-bosch: Checking for an error message is a pretty fragile approach. I am also not sure about the background of this check. So, some content is added to an archive, and afterwards an exception is ignored that there is no content? Maybe it is possible to change the workflow: Check if there is content to add to the ZIP, log a message if not, otherwise write the archive and treat all errors as critical.

core/core-workflow-steps/src/test/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImplTest.java
oheger-bosch: Ah, sorry. This was a problem of my understanding. I thought that the archive was deflated and not created. So yes, the tests are fine.

core/core-workflow-steps/src/main/java/org/eclipse/sw360/antenna/workflow/generators/SourceZipWriterImpl.java
KorSin: I also would want to know, if during a run a zip file was empty. We could leave it as it is now, since every IOException (a ZipException is derive from it to) is caught, logged and added to the reporter.

oheger-bosch: Ok, agreed.

oheger-bosch (approved):

Issue #588, opened 2020-09-03
Compliance Tool: Updating the same release twice with different licenses does not update the licenses


neubs-bsi:
### Summary of the Bug
When running the compliance tool updater twice with the same release set to `OSM_APPROVED` but changing the declared, observed or overwritten license does not update it the second time. 

### Steps to reproduce
- Run the same release with different licenses twice with the `OSM_APPROVED` flag in the updater

### Acceptance Criteria
- [ ] understand where this behaviour comes from
- [ ] Release gets updated with all data each time the updater run. 
- [ ] New run with `OSM_APPROVED` takes precedence

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

neubs-bsi:
Exact desired behaviour needs to be discussed:
Do we want the release to be updateable after first approval or not? Are there conditions?

neubs-bsi:
Depends on https://github.com/bosch-io/osm-antenna/issues/942

neubs-bsi:
Currently, it is true, that once a release is set to approved with the compliance tool, no  more changes can be done via the compliance tool. currently this is because of https://github.com/eclipse/antenna/issues/587. until it is fixed, this issue can't continue

KorSin:
During working on #619, I experienced that some fields, including the license, are not updated. The reason seems to be [here](https://github.com/eclipse/antenna/blob/a66aa2e60b17c51d63b97a87cea2e310ace1eb9f/modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java#L134-L135). After the updater has read a release from a csv file, a corresponding release is searched for in SW360. If found, it is merged together with the local one in which case the SW360 release takes the precedence. The actual update is performed with the merged one, making changes in the local release redundant. Maybe changing the priority during the merge just solves the problem.

PR #620, opened 2020-11-03
Neubs bsi/update release appropriately/#588

neubs-bsi:

It is solving the issue by introducing the option to merge the release gotten from the artifact with precedence when having the clearing state WIP and the release is not yet approved in any kind or form with precedence.

See commit messages for more detailed information. 

Issue: #588

@KorSin

*Type of change*:  bug fix

Tests still pass, were adapted and some added. 

### Checklist
Must:


core/model/src/main/java/org/eclipse/sw360/antenna/model/artifact/facts/ArtifactClearingState.java
```diff
@@ -31,11 +31,21 @@ public String getFactContentName() {
...
-        AUTO_EXTRACT,
-        PROJECT_APPROVED,
-        OSM_APPROVED,
-        WORK_IN_PROGRESS;
+        INITIAL(1),
+        WORK_IN_PROGRESS(2),
+        EXTERNAL_SOURCE(3),
+        AUTO_EXTRACT(4),
+        PROJECT_APPROVED(5),
+        OSM_APPROVED(6);
```
oheger-bosch, line 39: Is the order really the criterion to decide whether a release can be updated? I would rather think, there are clearing states in which updates are possible, and those in which this is not the case. So, should there just be a boolean field like _updatesAllowed_? Also, is the order meaningful for all states, e.g. does it make sense to compare _EXTERNAL_SOURCE_ with _AUTO_EXTRACT_?

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -125,14 +126,18 @@ private boolean isLicenseInSW360(License license) {
      *                                 updated if it already exists
      * @return the updated or newly created release entity
      */
-    public SW360Release getOrCreateRelease(SW360Release sw360ReleaseFromArtifact, boolean updateExisting) {
+    public SW360Release getOrCreateRelease(SW360Release sw360ReleaseFromArtifact, boolean updateExisting, boolean artifactHasPrecedence) {
```
oheger-bosch, line 129: The parameter name _artifactHasPrecedence_ (which is used in other places as well) refers to a rather technical and implementation-specific mechanism. Its purpose is to decide whether the data in SW360 should be updated or not - so better name it accordingly, e.g. _updateSW360_ or something like that.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -125,14 +126,18 @@ private boolean isLicenseInSW360(License license) {
...
         Optional<SW360SparseRelease> optSparseReleaseByIds =
                 releaseClientAdapter.getSparseReleaseByExternalIds(sw360ReleaseFromArtifact.getExternalIds());
         Optional<SW360SparseRelease> optSparseRelease = optSparseReleaseByIds.isPresent() ? optSparseReleaseByIds :
                 releaseClientAdapter.getSparseReleaseByNameAndVersion(sw360ReleaseFromArtifact.getName(),
                         sw360ReleaseFromArtifact.getVersion());
-        Optional<SW360Release> optRelease = optSparseRelease.flatMap(releaseClientAdapter::enrichSparseRelease)
-                .map(sw360ReleaseFromArtifact::mergeWith);
+        Optional<SW360Release> optRelease = optSparseRelease.flatMap(releaseClientAdapter::enrichSparseRelease);
+        Optional<String> clearingState = optRelease.map(SW360Release::getClearingState);
```
oheger-bosch, line 136: This optRelease.map() and the following two can probably be combined to a single one. I think, this would be easier to read.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
```diff
@@ -142,6 +147,24 @@ public SW360Release getOrCreateRelease(SW360Release sw360ReleaseFromArtifact, bo
         return releaseClientAdapter.createRelease(sw360ReleaseFromArtifact);
     }
 
+    /**
+     * Checks whether a given release has a higher or equal ordered clearing release state than a given clearing state string.
+     * @param sw360ReleaseFromArtifact given release that might have a clearing state information
+     * @param clearingState Optional clearing state that gets checked against
+     * @return true if the release as a higher or equal clearing state to the optional string or if the optional clearing state or both are null or empty
+     */
+    private boolean artifactClearingStateIsHigherOrEqualToSW360Release(SW360Release sw360ReleaseFromArtifact, Optional<String> clearingState) {
```
oheger-bosch, line 156: Again a rather technical method name. Something like _canUpdate()_ or _shouldUpdate()_ would reveal the intention clearer.

oheger-bosch (changes requested):

modules/sw360/sw360-workflow/src/test/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdaterTest.java
```diff
@@ -275,7 +275,60 @@ public void testGetOrCreateReleaseFoundByExternalIDs() {
...
+        when(releaseClientAdapter.enrichSparseRelease(sparseRelease)).thenReturn(Optional.of(foundRelease));
+        when(releaseClientAdapter.updateRelease(any()))
+                .thenAnswer((Answer<SW360Release>) invocationOnMock -> {
+                    SW360Release rel = invocationOnMock.getArgument(0);
+                    assertThat(rel.getExternalIds()).isEqualTo(extIDs);
+                    assertThat(rel.getCopyrights()).isEqualTo(foundCopyright);
+                    return patchedRelease;
+                });
+
+        assertThat(metaDataUpdater.getOrCreateRelease(queryRelease, true, true)).isEqualTo(patchedRelease);
```
KorSin, line 331: A lot of code in those tests is duplicated. Maybe you can refactor the duplicates into methods.

modules/sw360/sw360-workflow/src/main/java/org/eclipse/sw360/antenna/sw360/SW360MetaDataUpdater.java
neubs-bsi: I went with a mixture and chose `overwriteSW360Data`.

oheger-bosch (approved):
Nice!

neubs-bsi:
I'd actually not add the test, as I see no value in that specific test. The clearing state is tested in the metadataupdater, so at least part of the class should be covered.

KorSin:
Looks  good. Since it is the first time I review something from you, do you intend to squash some of the commits?

neubs-bsi:
No, I think each commit stands for itself and builds. A squash should not be necessary.

Issue #634, opened 2021-01-14
FailCausingResult in Analyzer if no artifacts does not break build


neubs-bsi:
### Summary of the Bug
When an analyser has a fail causing result, and no artifacts are found in any other analyser this does **not** break the build. 

### Steps to reproduce
- Have an analyser fail with a fail causing result
- No other analyser included
- build does not break, but succeed with `[WARNING] No analyser yielded artifacts, skip all other workflow steps`

### Acceptance Criteria
- [ ] Build breaks when their are fail causing results and no artifacts in analyser

### Definition of Done
- Acceptance criteria fulfilled
- A test case is created to reproduce the bug
- A PR is created, the CI infrastructure reports green, the bug test case proves that bug is fixed
- The PR is reviewed and approved
- No TODOs left in the code unless explained in the ticket, if something else is still open, this is summarized in a comment in the issue
- Documentation is updated

PR #636, opened 2021-01-15
Checks for failCausingResults when no Artifacts after Analyzer run

neubs-bsi:
This enures that the build breaks even when there are no artifacts
found in the Analyzer run. This is important because the reason
that there are no artifacts might be a fail causing result.

Issue #634
Closes #34

Signed-off-by: Stephanie Neubauer <Stephanie.Neubauer@bosch.io>

*Type of change*:  bug

New unit test was created. 

### Checklist
Must:


bs-ondem (approved):
Looks good to me.

